PR #8, opened 2018-04-30
Adds instructions to setup in README

germsvel:
What changed?
=============

Adds some basic instructions on how to setup project. A couple of things worth highlighting:

- Ran into issues because `automake` wasn't installed. Thus, I have included a `Requirements` section. But since this section does not include all requirements, maybe we should remove it or augment it. I include it here, but I'm happy to remove it if it doesn't make sense to include it.

- Added a command to clone repo with submodules. This is helpful to know in order to be able to run the shared ethereum tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,23 @@
 # Mana
 
 [![CircleCI](https://circleci.com/gh/poanetwork/mana/tree/master.svg?style=svg)](https://circleci.com/gh/poanetwork/mana/tree/master) [![Waffle.io - Columns and their card count](https://badge.waffle.io/poanetwork/mana.svg?columns=all)](https://waffle.io/poanetwork/mana)
+
+# Requirements
+
+Make sure you have `automake` && `autoconf` installed. If you don't, you can get
+them from homebrew,
+
+```
+brew install automake
+brew install autoconf
+```
+
+# Installation
+
+* Clone repo with submodules (so you can get the shared tests),
+
+```
+git clone --recurse-submodules https://gihub.com/poanetwork/mana.git
+```
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,3 +1,31 @@
...
+git clone --recurse-submodules https://gihub.com/poanetwork/mana.git
+```
+
+* Run `mix deps.get`
+
+If you have issues compiling `libsecp256k1`, try compiling it directly:
+
+```
+export "CFLAGS=-I/usr/local/include -L/usr/local/lib"
+cd deps/libsecp256k1 && rebar compile
```
ayrat555: I guess it's a bug in https://github.com/exthereum/libsecp256k1. Because you always have to compile it manually

germsvel: Yeah, I got this bit of code from https://github.com/exthereum/blockchain#installation. 

Is it okay to include like I have done so here (i.e. "If you have problems with..."), or do you prefer it to just be a clear step in the installation process?

ayrat555: I think it's better to fix problem with `libsecp256k1` compilation. I will try to do it today if it's ok

germsvel: Oh yeah, if we can fix the problem, then that'd be way better.

ayrat555: @germsvel possible temporary solution is adding to deps `{:libsecp256k1, [github: "exthereum/libsecp256k1", manager: :rebar]}`

It fetches `libsecp256k1` from github and compiles it with rebar

germsvel: @ayrat555 I'm going to remove the section talking about `libsecp256k1` from this PR since we're planning on fixing that.

PR #9, opened 2018-05-01
Fix remaining EVM tests

ayrat555:
- fix remaining EVM tests
- remove ex_rlp from umbrella app
- update ex_rlp to the latest version (0.3.0)
- fix test submodule


apps/evm/lib/evm/log_entry.ex
```diff
@@ -0,0 +1,100 @@
...
+          255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255,
+          255, 255>>
+      ]
+
+  """
+  @spec to_list(t()) :: [binary()]
+  def to_list(log) do
+    topics =
+      log.topics
+      |> Enum.map(fn topic ->
```
masonforest: would this work?

`Enum.map(&Helpers.left_pad_bytes/1)`

apps/evm/test/evm_test.exs
```diff
@@ -301,4 +246,10 @@ defmodule EvmTest do
       {address, storage}
     end)
   end
+
+  def logs_hash(logs) do
+    logs
+    |> ExRLP.encode()
+    |> :keccakf1600.sha3_256()
```
masonforest: do we have access to `exeth_crypto` here?

apps/evm/lib/evm/log_entry.ex
ayrat555: yes, thanks

apps/evm/test/evm_test.exs
ayrat555: @masonforest no, in this project we do not  have access to `exeth_crypto`. I'll try to replace all direct usages of `:keccakf1600` to `exeth_crypto`  now

PR #14, opened 2018-05-02
Add CircleCI config

cernivec:
Adds CircleCI config file to the root of the umbrella project

The tests are not passing, but CircleCI is running them.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,51 @@
+version: 2
+jobs:
+  build:
+    parallelism: 1
+    docker:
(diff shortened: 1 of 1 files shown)


ayrat555:
You can include dialyzer check to CI
` mix dialyzer  --halt-exit-status` but exleveldb should be excluded from it

ayrat555:
I suggest enabling dialyzer for some projects in this PR. Dialyzer has option for skipping dependencies.

cernivec:
@ayrat555 @germsvel thanks for the feedback. I will include dialyzer in a different PR.

Issue #20, opened 2018-05-04
implement k-buckets


ayrat555:
currently node does not save other nodes to k-buckets

https://pdos.csail.mit.edu/~petar/papers/maymounkov-kademlia-lncs.pdf

PR #22, opened 2018-05-07
Kademlia p2p

ayrat555:
Resolves https://github.com/poanetwork/mana/issues/20

Kademlia basic data structures are implemented.


vyorkin:
Almost finished reading the paper, will review soon

apps/ex_wire/lib/ex_wire/struct/routing_table.ex
```diff
@@ -0,0 +1,151 @@
...
+  Adds node to routing table.
+  """
+  @spec add_node(t(), Peer.t()) :: t()
+  def add_node(
+        table = %__MODULE__{current_node: %Peer{remote_id: current_node_id}},
+        %Peer{remote_id: current_node_id}
+      ),
+      do: table
+
+  def add_node(table = %__MODULE__{}, node = %Peer{}) do
```
vyorkin: tbh I'm not sure if its actually a good idea to call this method `add_node`.
As we know it doesn't always "add" node, according to the "Kademlia" paper it
> updates the appropriate _k_-bucket for the senders node ID

 so maybe it would be better to call it `update_sender`.
What do you think? Same thing regarding the `add_node` function in `Bucket`:
maybe `update_node` better reveals the intention there too.

apps/ex_wire/lib/ex_wire/struct/bucket.ex
```diff
@@ -0,0 +1,273 @@
...
+               134, 62, 206, 18, 196, 245, 250, 104, 227, 186, 155, 11, 81, 64, 124,
+               204, 0, 46, 238, 211, 179, 16, 45, 32, 168, 143, 28, 29, 60, 49, 84,
+               226, 68, 147, 23, 184, 239, 149, 9, 14, 119, 179, 18, 213, 204, 57, 53,
+               79, 134, 213, 214, 6>>
+           }
+         ],
+         updated_at: 1525704921
+       }}
+  """
+  @spec add_node(t(), Peer.t()) :: {atom, t()}
```
vyorkin: Personally, I think that `update_node` would be a better name than `add_node`

apps/ex_wire/lib/ex_wire/struct/bucket.ex
```diff
@@ -0,0 +1,273 @@
...
+         nodes: [],
+         updated_at: 1525704921
+       }
+  """
+  @spec remove_node(t(), Peer.t()) :: t()
+  def remove_node(bucket = %Bucket{nodes: nodes}, node) do
+    new_nodes =
+      nodes
+      |> Enum.drop_while(fn bucket_node ->
+        Peer.equal?(node, bucket_node)
```
vyorkin: I know I've already asked before but I don't remember why we're removing all `Peer`' s here?
As I understand we have a set of unique peers. We can't have two peers with the same `remote_id` here.

apps/ex_wire/lib/ex_wire/struct/bucket.ex
```diff
@@ -0,0 +1,273 @@
...
+        nodes: [],
+        updated_at: 1525704921
+      }
+
+  """
+  @spec new(Peer.t()) :: t()
+  def new(id, options \\ [time: :actual]) do
+    %__MODULE__{
+      id: id,
+      nodes: [],
```
vyorkin: I suggest using a map here instead since we'll need to lookup peers by their `remote_id`'s very often.

vyorkin: Maybe list is ok for the initial implementation, but we need order set (no such thing built in elixir yet), I'll try to checkout the branch and replace the list with [rbtree](https://hexdocs.pm/rbtree/Tree.html) implementation

apps/ex_wire/test/ex_wire/struct/bucket_test.exs
```diff
@@ -0,0 +1,69 @@
...
+        ExWire.Struct.Peer.new(
+          "13.84.180.140",
+          30303,
+          "30b7ab30a01c124a6cceca36863ece12c4f5fa68e3ba9b0b51407ccc002eeed3b3102d20a88f1c1d3c3154e2449317b8ef95090e77b312d5cc39354f86d5d606"
+        )
+
+      {:reinsert_node, ^node, bucket} =
+        bucket
+        |> Bucket.insert_node(node)
+        |> Bucket.add_node(node)
```
vyorkin: yep, naming matters :)

`insert_node` followed by the `add_node` looks slightly confusing to me, imagine instead:

```elixir
      {:reinsert_node, ^node, bucket} =
        bucket
        |> Bucket.insert_node(node)
        |> Bucket.refresh_node(node)
```

apps/ex_wire/lib/ex_wire/struct/bucket.ex
vyorkin: Ok, just tried to update your code to used it instead of list, basically it works, but there are too many changes needed to use rbtree which out of scope of this PR, also I don't like [how rbtree is written](https://github.com/vyorkin-forks/rbtree/blob/master/lib/rbtree.ex#L519). Lets just use lists for now and create a separate issue to address this later

apps/ex_wire/lib/ex_wire/util/xor_distance.ex
```diff
@@ -0,0 +1,75 @@
...
+  defp binary_to_bits(binary, acc \\ [])
+
+  defp binary_to_bits("", acc), do: acc
+
+  defp binary_to_bits(
+         <<b1::size(1), b2::size(1), b3::size(1), b4::size(1), b5::size(1), b6::size(1),
+           b7::size(1), b8::size(1), tail::binary>>,
+         acc
+       ) do
+    acc = acc ++ [b1, b2, b3, b4, b5, b6, b7, b8]
```
vyorkin:
```elixir
  defp binary_to_bits(binary) do
    for(<<bit::1 <- binary>>, do: bit)
  end
```

vyorkin:
Ok, its too late now, I'll continue tomorrow

apps/ex_wire/lib/ex_wire/struct/bucket.ex
vyorkin: I think its ok to leave it as is for now

apps/ex_wire/lib/ex_wire/struct/bucket.ex
```diff
@@ -0,0 +1,273 @@
...
+        port: 30303,
+        remote_id: <<4, 48, 183, 171, 48, 160, 28, 18, 74, 108, 206, 202, 54, 134,
+          62, 206, 18, 196, 245, 250, 104, 227, 186, 155, 11, 81, 64, 124, 204, 0,
+          46, 238, 211, 179, 16, 45, 32, 168, 143, 28, 29, 60, 49, 84, 226, 68,
+          147, 23, 184, 239, 149, 9, 14, 119, 179, 18, 213, 204, 57, 53, 79, 134,
+          213, 214, 6>>
+      }
+  """
+  @spec tail(t()) :: Peer.t()
+  def tail(%Bucket{nodes: nodes}), do: nodes |> List.last()
```
vyorkin: **tail** usually means **the rest of the elements**, not the **last element**

apps/ex_wire/lib/ex_wire/struct/routing_table.ex
ayrat555: I think `add_node` is ok. Because users of this method do not if a node is already in the routing table or not.

apps/ex_wire/test/ex_wire/struct/bucket_test.exs
ayrat555: Actually `insert_node` should be private.
More correct version:
```
 bucket
 |> Bucket.add_node(node)
 |> Bucket.add_node(node)
```

apps/ex_wire/lib/ex_wire/struct/routing_table.ex
vyorkin: yeah, I understand that, but I still don’t see how it could be a reason to name it add_node while its purpose is to refresh the node in k bucket :) maybe I didn’t understand the point

apps/ex_wire/lib/ex_wire/struct/bucket.ex
ayrat555: I knew about this problem. But there are only 16 items in every bucket.

apps/ex_wire/lib/ex_wire/kademlia_config.ex
```diff
@@ -0,0 +1,62 @@
+defmodule ExWire.KademliaConfig do
```
masonforest: What do you think about moving this to `ExWire.Kademlia::Config` so we can push as much under `Kademlia` as possible?

apps/ex_wire/lib/ex_wire/struct/routing_table.ex
```diff
@@ -0,0 +1,151 @@
...
+  defp bucket_at(%__MODULE__{buckets: buckets}, id) do
+    buckets |> Enum.at(id)
+  end
+
+  defp handle_full_bucket(_table, _bucket, _candidate_for_removal, _candidate_for_insertion) do
+    # TODO
+  end
+
+  @spec traverse(t(), integer(), [Peer.t()], integer()) :: [Peer.t()]
+  defp traverse(table, bucket_id, acc \\ [], step \\ 1) do
```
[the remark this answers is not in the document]
masonforest: Yeah this was definitely the hardest part of the PR for me to understand. Maybe name a couple more variables and pull out some private methods? Something like:
 
    left_boundry = bucket_id - step
    right_boundry = bucket_id + step
    bucket_size = Enum.count(acc)
    max_bucket_size = Config.id_size() # (See other comment on pulling out Config)


      if bucket_id in left_boundry..right_boundry && bucket_size < max_bucket_size do
        is_out_of_left_boundary = left_boundry < 0
        is_out_of_right_boundary = right_boundry >= max_bucket_size

        acc
          |> prepend_if_true(is_out_of_left_boundary, nodes_at(left_boundry))
          |> prepend_if_true(is_out_of_right_boundary, nodes_at(right_boundry))
      else
        acc
      end
    end

    # https://elixirforum.com/t/creating-list-adding-elements-on-specific-conditions/6295/4
    defp prepend_if_true(list, cond, extra) do
      if cond, do: extra ++ list, else: list
    end

Untested and probably still need some work but hopefully will be helpful ^

apps/ex_wire/lib/ex_wire/util/xor_distance.ex
```diff
@@ -0,0 +1,65 @@
+defmodule ExWire.Util.XorDistance do
```
masonforest: Would it make sense to push this module under `Kademlia`? I think we'll only be using them when using `Kademlia` related stuff and I like to keep utility modules as clean as possible.

apps/ex_wire/lib/ex_wire/struct/routing_table.ex
```diff
@@ -0,0 +1,156 @@
...
+        replace_bucket(table, bucket_idx, bucket)
+    end
+  end
+
+  @doc """
+  Returns neighbours of a specified node.
+  """
+  @spec neighbours(t(), Peer.t()) :: [Peer.t()]
+  def neighbours(table = %__MODULE__{}, node = %Peer{}) do
+    bucket_idx = bucket_id(table, node)
```
masonforest: nitpick: Is there another name we could use here that wouldn't conflict with the function and be more descriptive? `node_bucket_id`? :/

apps/ex_wire/lib/ex_wire/struct/routing_table.ex
```diff
@@ -0,0 +1,156 @@
...
+    similar_to_current_node = nodes_at(table, bucket_idx)
```
masonforest: is `similar` the same as `near`? I want to avoid using two different terms to describe the same concept if possible. `nearest_neighbors` maybe?

Issue #23, opened 2018-05-07
add String10MbData


vyorkin:
The problem is that String10MbData test eats memory heap so it takes more than allowed 4GB on CircleCI. To prevent this I've set max_heap_size to 4095 mb (see circle ci config for details).

vyorkin:
We're moving to paid plan so we can request 8 GB from CircleCI support :P

PR #24, opened 2018-05-07
Gv fix blockchain test

vyorkin:
The problem was [hitting 4G memory limit](https://circleci.com/docs/1.0/oom/) on CircleCI.
`String10MbData` test eats memory heap so it takes more than allowed 4GB.
To prevent this I've set max_heap_size to 4095 mb (see `.circleci/config.yml` config for details).

My solution is to:

1. Set the default maximum heap size to 4095 for test env (to prevent process killing).
2. Ignore the `String10MbData` test for now.
3. Create a ticket to find out whats wrong later (see the [#23](https://waffle.io/poanetwork/mana/cards/5af074b1dcbe45001be58757) in Waffle)

More info:

* http://erlang.org/doc/man/erl.html (for `hmax` CLI option).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -47,5 +47,4 @@ jobs:
           key: v1-build-cache
           paths: "_build"
 
-      - run: mix test --exclude network --trace
-
+      - run: elixir --erl "+hmax 33554432" -S mix test --exclude network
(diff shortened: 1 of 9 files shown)


vyorkin:
@germsvel I hope you don't mind if I merge this branch

PR #37, opened 2018-05-08
Add chain_id for blockchain tests that need it

germsvel:
This work is part of #10 -- making all ethereum common tests pass in the blockchain project.

What changed?
===============

According to the [EIP155 specification](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md), a chain_id is used when signing the transaction. So any chains after the EIP155 fork need to include that chain id when signing the transaction.

The implementation code already handles that. But the tests were ignoring any common tests that had a `v` value greater than `30`. We fix that here by passing the `mainnet`'s `chain_id` of `1` for any chains that came after EIP155.

Because we now handle that, the logic for all common tests generated from the `TransactionTests` directory can be combined into a single use of `define_common_tests/3` (note we continue to ignore the tests we were already excluding)

### DIFF
--- a/apps/blockchain/test/blockchain/transaction_test.exs
+++ b/apps/blockchain/test/blockchain/transaction_test.exs
@@ -3,12 +3,10 @@ defmodule Blockchain.TransactionTest do
   use EthCommonTest.Harness
   doctest Blockchain.Transaction
 
-  require Logger
-
   alias Blockchain.Transaction
   alias Blockchain.Transaction.Signature
 
-  @chains ~w(
+  @forks ~w(
     Byzantium
     Constantinople
     EIP150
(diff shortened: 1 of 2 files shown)


apps/blockchain/test/blockchain/transaction_test.exs
```diff
@@ -3,12 +3,10 @@ defmodule Blockchain.TransactionTest do
   use EthCommonTest.Harness
   doctest Blockchain.Transaction
 
-  require Logger
-
   alias Blockchain.Transaction
   alias Blockchain.Transaction.Signature
 
-  @chains ~w(
+  @forks ~w(
```
germsvel: I changed this to `@forks` but I'm not sure if it's better or not. It seems like there's no clear way to express this. They're technically the name of the chain after a fork, so maybe `@chains` was better? I also thought about `networks` but that too seems to mean something else (like a test network).

vyorkin:
Just read [the code](https://github.com/poanetwork/mana/blob/master/apps/blockchain/lib/blockchain/transaction/signature.ex#L209) & tests.
I think I understand whats going on here :)
LGTM!

apps/blockchain/test/blockchain/transaction_test.exs
masonforest: I like `@forks` better too. 👍

Issue #31, opened 2018-05-07
Fix blockchain code formatting


vyorkin:
Extract & give names to long anonymous functions

Issue #33, opened 2018-05-08
Add code formatting check to CI


vyorkin:
Like in elixir-lang repo, see: 

* https://github.com/elixir-lang/elixir/blob/master/Makefile#L203
* https://github.com/elixir-lang/elixir/issues/6643#issuecomment-336063332

PR #39, opened 2018-05-08
Format blockchain app

germsvel:
As part of #33, this PR formats the blockchain app with elixir formatter. We include the `.formatter.exs` file in the `apps/blockchain/` for now. Once we are done formatting all the apps, we should have a top level `.formatter.exs` and we can remove the app specific ones.


germsvel:
Just noticed that we an issue for some of the blockchain stuff -> https://github.com/poanetwork/mana/issues/31. I'd be either happy to extract things into functions so that we can address that issue, or I can also close this if we prefer to take a different approach (e.g. a PR per file for formatting blockchain).

vyorkin:
@germsvel I think there could be too many such functions, so maybe it's better to merge this PR and just close #31 for now (we'll refactor gradually), what do you think?

germsvel:
@vyorkin I'm definitely in favor of getting this in there, having fixed any glaring issues (like the comments that were fixed in 902f4b2fdd84061c3adfd926390211e5ce42fda4), and then refactor functions gradually. I think fixing all the functions that need to be extracted or going file by file like the elixir core team did sounds nice, but it would add a lot of overhead for the team right now. 

I'll wait for @ayrat555's thoughts on it before merging since he had tried this as well. But if he's okay with that, I say we go ahead and do what you said.

apps/blockchain/lib/blockchain/block.ex
```diff
@@ -746,11 +801,29 @@ defmodule Blockchain.Block do
       |> add_rewards_to_block(db, chain.params[:block_reward])
 
     # The following checks Holistic Validity, as defined in Eq.(29)
-    errors = []
-      ++ if child_block.header.state_root == block.header.state_root, do: [], else: [:state_root_mismatch]
-      ++ if child_block.header.ommers_hash == block.header.ommers_hash, do: [], else: [:ommers_hash_mismatch]
-      ++ if child_block.header.transactions_root == block.header.transactions_root, do: [], else: [:transactions_root_mismatch]
-      ++ if child_block.header.receipts_root == block.header.receipts_root, do: [], else: [:receipts_root_mismatch]
+    errors =
```
ayrat555: @germsvel that's definitely should be fixed before merging

apps/blockchain/lib/blockchain/block.ex
```diff
@@ -13,21 +13,25 @@ defmodule Blockchain.Block do
...
   # Defined in Eq.(18)
-  defstruct [
-    block_hash: nil,   # Hash for this block, acts simply as a cache.
-    header: %Header{}, # B_H
-    transactions: [],  # B_T
-    ommers: [],        # B_U
-  ]
+  defstruct block_hash: nil,
+
+            # Hash for this block, acts simply as a cache.
```
ayrat555: @germsvel wrong comment

apps/blockchain/lib/blockchain/block.ex
germsvel: Very good catch. This is a disaster in the new format.

germsvel: Updated in 4b64ba226a8451bba26d94181176f4267c0383b0

apps/blockchain/lib/blockchain/block.ex
germsvel: Updated in 62a5ef4ebb7a6a5c6ea3c458b7f2ce51a170199d

apps/blockchain/lib/blockchain/block.ex
ayrat555: @germsvel it looks much better. good work. You also can add typespecs to new methods.

germsvel: Oh good catch. I'll add those and merge after that.

germsvel: Added typespecs in 5f22b330e6658f940df60d8a5f607cc88c2a5af4

Issue #42, opened 2018-05-10
Dialyzer warnings


cernivec:
Working though dialyzer warnings, here is the list of the ones I need more context to fix.
(This is updated as I go though the warnings)

**Blockchain app**

- `EVM.Debugger:break_on` expects an integer but gets binary.
```
apps/blockchain/lib/blockchain/application.ex:16: The call 'Elixir.EVM.Debugger':break_on([{'address',binary()},...]) will never return since the success typing is
(['address']) -> integer() 
and the contract is (elixir:keyword('Elixir.EVM.Debugger.Breakpoint':conditions())) -> 'Elixir.EVM.Debugger.Breakpoint':id()
```
- All these three warnings come form`gen_child_block()` function. I can't see why `set_block_number()` is breaking the contract.
```
apps/blockchain/lib/blockchain/block.ex:558: Function gen_child_block/2 has no local return
apps/blockchain/lib/blockchain/block.ex:558: Function gen_child_block/3 has no local return
[...]

PR #43, opened 2018-05-10
Ignore dialyzer warnings

cernivec:
## What
Add all the warnings to the `.dialyzer.ignore-warnings` instead of ignoring whole applications

## Why
This is so the dialyzer warns about issues on newly deployed code.

Issue: https://github.com/poanetwork/mana/issues/42


ayrat555:
We shouldn't do that. Because when we will be changing code, we will have to update this ignore file

cernivec:
Ok, I see your point. Maybe we can do this if there is a way to ignore the warnings without including the line number, and when the code gets changed the warning should be removed anyways?

ayrat555:
Also considering that currently there are not many updates I think it's possible to fix dialyzer warnings going from project to project. It should take a couple of days. Of course initially dialyzer warnings would seem cryptic

PR #51, opened 2018-05-12
Add rocksdb (database backend)

### LINKED ISSUE #32: Add rocksdb backend, perform simple benchmarking
Should be faster, maybe.
see:
https://github.com/urbint/rox

vyorkin:
closes #32


vyorkin:
fixing the build now

apps/merkle_patricia_tree/test/db/rocksdb_test.exs
```diff
@@ -0,0 +1,51 @@
...
+  alias MerklePatriciaTree.DB.RocksDB
+
+  test "init creates a rocksdb file" do
+    db_name = "/tmp/db#{MerklePatriciaTree.Test.random_string(20)}"
+
+    RocksDB.init(db_name)
+    # There is no `close` function in RocksDB API (for now),
+    # This means that we can't just close and re-open the db again.
+    # Also we can not open it twice at the same time (it'll be in the locked state).
+    # So we just check if the db file exists. This should be enough.
```
vyorkin: I've forked the rox repo, will try to write `close` nif in Rust.

vyorkin:
Same problem here – not enough memory.
```internal compiler error: Killed (program cc1plus)``` (see the stack trace below)
We need to upgrade to be able to increase up to 8GB (https://circleci.com/docs/2.0/configuration-reference/#resource_class).
Then I could fix the build (hopefully it’ll fix itself after upgrading).

```sh
running: "c++" "-O3" "-ffunction-sections" "-fdata-sections" "-fPIC" "-m64" "-I" "rocksdb/include/" "-I" "rocksdb/" "-I" "rocksdb/third-party/gtest-1.7.0/fused-src/" "-I" "snappy/" "-I" "." "-Wall" "-Wextra" "-std=c++11" "-Wno-unused-parameter" "-DNDEBUG=1" "-DSNAPPY=1" "-DOS_LINUX=1" "-DROCKSDB_PLATFORM_POSIX=1" "-DROCKSDB_LIB_IO_POSIX=1" "-o" "/home/circleci/app/_build/test/rustler_crates/rox_nif/release/build/librocksdb-sys-f2cef20ced9f8fd0/out/rocksdb/options/options_parser.o" "-c" "rocksdb/options/options_parser.cc"
cargo:warning=c++: internal compiler error: Killed (program cc1plus)
cargo:warning=Please submit a full bug report,
cargo:warning=with preprocessed source if appropriate.
cargo:warning=See <file:///usr/share/doc/gcc-6/README.Bugs> for instructions.
exit code: 4
```

.circleci/config.yml
```diff
@@ -14,10 +14,15 @@ jobs:
       - checkout
       - run: git submodule sync --recursive
       - run: git submodule update --recursive --init
+      - run: curl https://sh.rustup.rs -sSf | sh -s -- -y
```
vyorkin: [Rox requires that Rust be available at compile time](https://github.com/urbint/rox#dependencies)

apps/merkle_patricia_tree/mix.exs
```diff
@@ -51,6 +51,7 @@ defmodule MerklePatriciaTree.Mixfile do
       {:ex_rlp, "~> 0.3.0"},
       {:exth_crypto, in_umbrella: true},
       {:exleveldb, "~> 0.13.1"},
```
masonforest: Can we remove the dependency on `exleveldb` now? I can't think of a reason anyone would want to use `leveldb` instead of `rocksdb` now.

If we do want to keep it we could probably make it [optional](https://hexdocs.pm/mix/Mix.Tasks.Deps.html#module-dependency-definition-options).

apps/merkle_patricia_tree/lib/db/rocksdb.ex
```diff
@@ -0,0 +1,41 @@
...
+
+    {__MODULE__, db_ref}
+  end
+
+  @doc """
+  Retrieves a key from the database.
+  """
+  @spec get(DB.db_ref(), Trie.key()) :: {:ok, DB.value()} | :not_found
+  def get(db_ref, key) do
+    case Rox.get(db_ref, key) do
```
masonforest: What's this case do? Can we just return `Rox.get(db_ref, key)` directly?

apps/merkle_patricia_tree/lib/db/rocksdb.ex
```diff
@@ -0,0 +1,41 @@
...
+      :not_found -> :not_found
+    end
+  end
+
+  @doc """
+  Stores a key in the database.
+  """
+  @spec put!(DB.db_ref(), Trie.key(), DB.value()) :: :ok
+  def put!(db_ref, key, value) do
+    case Rox.put(db_ref, key, value) do
```
masonforest: same as question as above ^

apps/merkle_patricia_tree/mix.exs
vyorkin: Yea, good point

apps/merkle_patricia_tree/lib/db/rocksdb.ex
vyorkin: good spot, it is a leftover from the leveldb implementation

PR #53, opened 2018-05-12
Add ex_dash

vyorkin:
This PR adds [ex_dash](https://github.com/urbint/ex_dash), which

> provides a mix task that rebuilds a Dash Docset.

Idk if anyone else uses [Dash](https://kapeli.com/dash) or [Zeal](https://zealdocs.org/). Pretty convenient.

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 /_build
 /cover
 /deps
+/doc
 erl_crash.dump
 *.ez
 *.beam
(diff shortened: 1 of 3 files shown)


ayrat555:
I think this kind of dependencies shouldn't be in the project's mix.exs.

vyorkin:
yep, sure, it's not that hard to do it locally if somebody needs it

PR #64, opened 2018-05-14
Dialyzer warning fixes PR#1

cernivec:
## What
Fixes some of the dialyzer warnings for the  Blockchain app.
Warning fixed is included in each commit message.

This is the 1st PR, not all the warnings are gone yet.

An extra commit included which removes the unused `identity()` function.

### DIFF
--- a/apps/blockchain/lib/blockchain/account.ex
+++ b/apps/blockchain/lib/blockchain/account.ex
@@ -514,7 +514,7 @@ defmodule Blockchain.Account do
       iex> Blockchain.Account.get_storage(updated_state, <<01::160>>, 5)
       {:ok, 9}
   """
-  @spec put_storage(EVM.state(), EVM.address(), integer(), integer()) :: t
+  @spec put_storage(EVM.state(), EVM.address(), integer(), integer()) :: EVM.state | { EVM.state, t, t }
   def put_storage(state, address, key, value) do
     update_account(state, address, fn acct ->
       updated_storage_trie = storage_put(state.db, acct.storage_root, key, value)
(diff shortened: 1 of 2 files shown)


apps/blockchain/lib/blockchain/account.ex
```diff
@@ -514,7 +514,7 @@ defmodule Blockchain.Account do
       iex> Blockchain.Account.get_storage(updated_state, <<01::160>>, 5)
       {:ok, 9}
   """
-  @spec put_storage(EVM.state(), EVM.address(), integer(), integer()) :: t
+  @spec put_storage(EVM.state(), EVM.address(), integer(), integer()) :: EVM.state | { EVM.state, t, t }
```
ayrat555: Something is wrong with this typespec. Why should the method return a tuple with duplicate items? ```{ EVM.state, t, t }```
I think it returns just `EVM.state`

apps/blockchain/lib/blockchain/account.ex
```diff
@@ -514,7 +514,7 @@ defmodule Blockchain.Account do
       iex> Blockchain.Account.get_storage(updated_state, <<01::160>>, 5)
       {:ok, 9}
   """
-  @spec put_storage(EVM.state(), EVM.address(), integer(), integer()) :: EVM.state | { EVM.state, t, t }
+  @spec put_storage(EVM.state(), EVM.address(), integer(), integer()) :: EVM.state
```
ayrat555: I'm sorry but tuple is actually needed
```
if return_accounts do
      {updated_state, account, updated_account}
    else
      updated_state
end
```

cernivec: return_accounts will always be false in put_storage case won't it? So it won't return a tuple.

ayrat555: yes, you're right

PR #86, opened 2018-05-20
Bug fixes, initial refactoring

### LINKED ISSUE #48: Add `Blockchain.Math` module
Move functions like `calculate_total_refund` to it (intrinsic cost calc, refund calc, etc)

### LINKED ISSUE #69: Reclaim old tries from the DB
When an account is deleted, the corresponding `storage_trie_root` is deleted also.
The problem is that the underlying DB may still keep all stored items.
We need to remove the `storage_trie_root` from the DB.

image

related in Parity: https://github.com/paritytech/parity/blob/master/ethcore/src/executive.rs#L644

See https://github.com/poanetwork/mana/blob/master/apps/blockchain/lib/blockchain/interface/account_interface.ex#L238
and  `del_account` fn in `Account` module

### LINKED ISSUE #81: Merkle-Patricia Tree nil value in "branch" node
According to
image

In Eq. (194) we should "interpret" `()` as to be equivalent to `<<>>`, but not `nil`
`nil` values aren't RLP-serializable.

The relevant place: https://github.com/poanetwork/mana/blob/master/apps/merkle_patricia_tree/lib/trie/destroyer.ex#L76

Proposed fix: decode `<<>>` values as `nil`'s, but store as `<<>>`.

### LINKED ISSUE #82: Delete account storage after suicide
When an account is deleted, the corresponding `storage_trie_root` is deleted also.
The problem is that the underlying DB may still keep all stored items.
We need to remove the `storage_trie_root` from the DB.

### LINKED ISSUE #83: Extract genesis functionality to Blockchain.Genesis module

### LINKED ISSUE #84: Extract holistic validity check into separate module

vyorkin:
* Add test that reveals not cleared up storage bug
* Resolve some TODOS
* Extract MathHelper module
*  Add test to make sure the root of the tree depends only on the data, not on the order in which updates are made.
* Point Yellowpaper link to the version on Github
* Remove unused fn from the `ExecEnv` module
* Fix MMPT storage, refactor blockchain and evm apps
* Fix nil values in Merkle-Patricia Tree branch node
* Delete old tries from the DB on `Trie.remove`
* Extract holistic validity check into separate module
* Delete account storage when account itself is deleted
* Extract genesis block related functionality to `Blockchain.Genesis`
* Refactor `intrinsic_gas_cost` fn in `Transaction` module
* Refactor `TransactionTest`
* Refactor blockchain Account module, extract Account.Storage
* Extract `is_genesis_block?` fn to Genesis module
* Apply some other refactorings to blockchain and evm apps
* Set ExUnit timeout to infinity in blockchain app
* Fix some YP refs

This resolves #81, resolves #82, resolves #83, resolves #84, closes #69, resolves #48

### DIFF
--- a/apps/blockchain/lib/blockchain/block.ex
+++ b/apps/blockchain/lib/blockchain/block.ex
@@ -1,6 +1,6 @@
 defmodule Blockchain.Block do
   @moduledoc """
-  This module effective encodes a Block, the heart of the blockchain.
+  This module effectively encodes a Block, the heart of the blockchain.
   A chain is formed when blocks point to previous blocks,
   either as a parent or an ommer (uncle).
   For more information, see Section 4.3 of the Yellow Paper.
@@ -8,8 +8,8 @@ defmodule Blockchain.Block do
 
   alias ExthCrypto.Hash.Keccak
   alias Block.Header
-  alias Blockchain.Chain
-  alias Blockchain.{Account, Transaction}
+  alias Blockchain.{Account, Transaction, Chain, Genesis}
+  alias Blockchain.Block.HolisticValidity
   alias Blockchain.Transaction.Receipt
   alias MerklePatriciaTree.{Trie, DB}
 
@@ -159,8 +159,8 @@ defmodule Blockchain.Block do
   @spec put_block(t, DB.db()) :: {:ok, EVM.hash()}
   def put_block(block, db) do
     hash = block |> hash
-
-    :ok = MerklePatriciaTree.DB.put!(db, hash, block |> serialize |> ExRLP.encode())
+    block_rlp = block |> serialize |> ExRLP.encode()
+    :ok = MerklePatriciaTree.DB.put!(db, hash, block_rlp)
 
     {:ok, hash}
   end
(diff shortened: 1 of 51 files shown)


apps/blockchain/lib/blockchain/block.ex
```diff
@@ -304,7 +306,7 @@ defmodule Blockchain.Block do
   def get_parent_block(block, db) do
     case block.header.number do
       0 -> :genesis
-      _number -> get_block(block.header.parent_hash, db)
+      __ -> get_block(block.header.parent_hash, db)
```
ayrat555: two underscore

apps/blockchain/lib/blockchain/block.ex
```diff
@@ -1068,30 +868,26 @@ defmodule Blockchain.Block do
...
       [%Blockchain.Account{balance: 400100}]
   """
-  @spec add_rewards_to_block(t, DB.db(), EVM.Wei.t()) :: t
-  def add_rewards_to_block(block, db, reward_wei \\ @reward_wei) do
+  @spec add_rewards(t, DB.db(), EVM.Wei.t()) :: t
+  def add_rewards(block, db, reward_wei \\ @reward_wei) do
     # TODO: Add ommer rewards
 
-    if is_genesis?(block) do
+    if Genesis.is_genesis_block?(block) do
```
ayrat555: wouldn't it look better like:
```elixir
cond  
  Genesis.is_genesis_block?(block) -> ...,
  is_nil(block.header.beneficiary) -> ...,
  true -> ...
end

File: apps/evm/lib/evm/gas.ex [hunk 3]
```diff
@@ -217,7 +213,7 @@ defmodule EVM.Gas do
     max(out_memory_cost, in_memory_cost)
   end
 
-  # From Eq 220: Cmem(μ′i)−Cmem(μi)
+  # From Eq 220: C_mem(μ′i) − C_mem(μi)
```
ayrat555: wrong equation number?

Issue #76, opened 2018-05-15
Handle builtin/precompiled contract accounts


vyorkin:
See Appendix E. in the YP:

![image](https://user-images.githubusercontent.com/988849/40082086-9ba558f4-5898-11e8-9c25-3e691f7ab8fd.png)

https://github.com/paritytech/parity/blob/master/ethcore/res/ethereum/foundation.json#L2900
https://github.com/poanetwork/mana/blob/master/apps/blockchain/lib/blockchain/chain.ex#L65

vyorkin:
see `EVM.Builtin` also

vyorkin:
Related: https://vitalik.ca/general/2017/09/14/prehistory.html
How its implemented in Parity: https://github.com/paritytech/parity/blob/master/ethcore/src/builtin.rs

cernivec:
@masonforest what is the `s0` in this case `ExthCrypto.Hash.hash(s0, ExthCrypto.Hash.kec)`?

masonforest:
Conversation in slack: https://poanetwork.slack.com/archives/GA89BNBT7/p1526561179000117

cernivec:
This needs refactoring and adding more tests.

ayrat555:
done

PR #100, opened 2018-05-23
Precompiled Contract: Identity

cernivec:
### Issue: part of #76

Just passes though the data object and returns it as the output.


apps/evm/lib/evm/builtin.ex
```diff
@@ -7,6 +7,8 @@ defmodule EVM.Builtin do
   TODO: Implement and add doc tests.
   """
 
+  @g_identity 15 + 3
```
masonforest: Could we split this into two module variables? I think that would make it more clear that 15 is the base price and it costs 3 gas per byte

PR #101, opened 2018-05-24
Precompiled Contract: ecrec

cernivec:
### Issue: part of #76
Implements a precompiled contract that recovers a public key from a signed hash
(Elliptic curve digital signature algorithm public key recovery function)
Uses fixed gas amount.

### DIFF
--- a/apps/evm/lib/evm/builtin.ex
+++ b/apps/evm/lib/evm/builtin.ex
@@ -46,14 +46,17 @@ defmodule EVM.Builtin do
   def run_ecrec(gas, exec_env = %EVM.ExecEnv{data: data}) do
     used_gas = 3000
 
-    if(used_gas < gas and byte_size(data) == 128) do
+    if(used_gas < gas) do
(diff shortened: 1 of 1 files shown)


apps/evm/lib/evm/builtin.ex
```diff
@@ -7,9 +7,58 @@ defmodule EVM.Builtin do
...
+      iex> {_, _, _, result} = ecrec
+      iex> {:ok, result} == public_key
+      true
+  """
+
   @spec run_ecrec(EVM.Gas.t(), EVM.ExecEnv.t()) ::
           {EVM.Gas.t(), EVM.SubState.t(), EVM.ExecEnv.t(), EVM.VM.output()}
-  def run_ecrec(gas, exec_env), do: {gas, %EVM.SubState{}, exec_env, <<>>}
+  def run_ecrec(gas, exec_env = %EVM.ExecEnv{data: data}) do
+    used_gas = 3000
```
masonforest: Would it make sense to move this to a module variable?

Issue #106, opened 2018-05-28
Start up node discovery on app start


ayrat555:
Basic implementation of node discovery is finished.

The only thing left to do is start it

PR #110, opened 2018-05-28
Start up node discovery

ayrat555:
https://github.com/poanetwork/mana/issues/106

### DIFF
--- a/apps/ex_wire/lib/ex_wire/kademlia/routing_table.ex
+++ b/apps/ex_wire/lib/ex_wire/kademlia/routing_table.ex
@@ -11,14 +11,15 @@ defmodule ExWire.Kademlia.RoutingTable do
   alias ExWire.Handler.Params
   alias ExWire.Struct.Endpoint
 
-  defstruct [:current_node, :buckets, :network_client_name, :expected_pongs]
+  defstruct [:current_node, :buckets, :network_client_name, :expected_pongs, :discovery_nodes]
 
   @type expected_pongs :: %{required(binary()) => {Node.t(), Node.t()}}
   @type t :: %__MODULE__{
           current_node: Node.t(),
           buckets: [Bucket.t()],
           network_client_name: pid() | atom(),
-          expected_pongs: expected_pongs()
+          expected_pongs: expected_pongs(),
+          discovery_nodes: [Node.t()]
         }
 
   @doc """
@@ -53,7 +54,8 @@ defmodule ExWire.Kademlia.RoutingTable do
       current_node: node,
       buckets: initial_buckets,
       network_client_name: client_pid,
-      expected_pongs: %{}
+      expected_pongs: %{},
+      discovery_nodes: []
     }
   end
 
@@ -129,6 +131,19 @@ defmodule ExWire.Kademlia.RoutingTable do
   end
 
   @doc """
+  Returns current node's discovery nodes. Basically it just finds the closest to current node
+  nodes and filters nodes that were already used for node discovery.
+  """
+  @spec discovery_nodes(t()) :: [Node.t()]
+  def discovery_nodes(table) do
+    table
+    |> traverse(buckets_count() - 1)
+    |> Enum.reject(fn node ->
+      Enum.member?(table.discovery_nodes, node)
+    end)
+  end
+
+  @doc """
   Checks if node exists in routing table.
   """
   @spec member?(t(), Node.t()) :: boolean()
@@ -186,9 +201,6 @@ defmodule ExWire.Kademlia.RoutingTable do
       case node do
         {removal_candidate, _insertion_candidate} ->
           refresh_node(table, removal_candidate)
-
-        pinged_node = %Node{} ->
-          refresh_node(table, pinged_node)
 
         _ ->
           table
--- a/apps/ex_wire/lib/ex_wire/kademlia/routing_table.ex
+++ b/apps/ex_wire/lib/ex_wire/kademlia/routing_table.ex
@@ -11,7 +11,14 @@ defmodule ExWire.Kademlia.RoutingTable do
   alias ExWire.Handler.Params
   alias ExWire.Struct.Endpoint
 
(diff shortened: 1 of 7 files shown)


apps/ex_wire/lib/ex_wire/crypto.ex
```diff
@@ -132,4 +132,14 @@ defmodule ExWire.Crypto do
   def node_id_from_public_key(public_key) do
     Key.der_to_raw(public_key)
   end
+
+  @spec hex_to_bin(binary()) :: binary()
+  def hex_to_bin(hex_bin) do
```
masonforest: Want to `import ExthCrypto.Math, only: [hex_to_bin: 1]` like @vyorkin did [here](https://github.com/poanetwork/mana/blob/master/apps/evm/test/evm_test.exs#L2)? I'm good with either as long as we're consistent.

apps/ex_wire/lib/ex_wire/kademlia/routing_table.ex
```diff
@@ -240,23 +289,39 @@ defmodule ExWire.Kademlia.RoutingTable do
     Enum.at(buckets, id)
   end
 
-  @spec traverse(t(), integer(), [Node.t()], integer()) :: [Node.t()]
-  defp traverse(table, bucket_id, acc \\ [], step \\ 1) do
+  @spec traverse(t(), Keyword.t()) :: [Node.t()]
+  defp traverse(table, opts) do
```
masonforest: hmmm this function is getting pretty complicated. Maybe pull out `found_nodes(bucket_id, step)` and `is_within_boundry(bucket_id, step)` into private methods?

apps/ex_wire/test/integration/node_discovery_test.exs
```diff
@@ -26,31 +32,59 @@ defmodule ExWire.NodeDiscoveryTest do
...
+    pings_count = routing_table.expected_pongs |> Map.values() |> Enum.count()
+
+    assert pings_count > 0
+  end
+
+  test "successfully finished discovery and adds nodes to routing_table", %{
+    kademlia_name: kademlia_name
+  } do
+    # let's wait for all discovery rounds
+    Process.sleep(30_000)
```
germsvel: Should we tag these tests so that we don't run them all the time since they're sleeping? We should run them on CI in a full run, but maybe we can exclude them by default. What do you think?

apps/ex_wire/lib/ex_wire/kademlia/server.ex
```diff
@@ -65,6 +74,34 @@ defmodule ExWire.Kademlia.Server do
     {:reply, neighbours, state}
   end
 
+  def handle_info({:discovery_round, nodes}, %{routing_table: routing_table}) do
+    updated_table = Discovery.start(routing_table, nodes)
+
+    schedule_discovery_round(updated_table.discovery_round)
```
germsvel: I'm not sure if I'm not getting this right, but should this be passing a modified version of `nodes` as its second argument? 

This is how I see it: 
- In round `0`, `schedule_discovery_round` gets called from the `init` function so it gets the `nodes` and it passes them into the `Process.send_after` for round `1`. 
- In round `1`,  we hit this `handle_info({:discovery_round, nodes}` function and we update the table in line 78 and then schedule another round on his line (line 80). 
- In round `2`, when we hit this `handle_info` we'll get `{:discovery_round, []}`. 
- All subsequent rounds will have the same. 

Is that intentional? Do we only need the nodes for the first time we run `updated_table = Discovery.start(routing_table, nodes)`?

apps/ex_wire/test/ex_wire/kademlia/routing_table_test.exs
```diff
@@ -60,7 +54,7 @@ defmodule ExWire.Kademlia.RoutingTableTest do
       assert Enum.count(pongs) == 1
 
       {_mdc, pair} = Enum.at(pongs, 0)
-      assert pair == {filler_node, node}
+      {^filler_node, ^node, _} = pair
```
germsvel: Since this is a test, should we be `assert`ing something?

apps/ex_wire/lib/ex_wire.ex
```diff
@@ -25,10 +27,14 @@ defmodule ExWire do
         []
       end
 
-    children =
-      [
-        worker(ExWire.NodeDiscoverySupervisor, [])
-      ] ++ sync_children
+    node_discovery =
+      if Config.discovery() do
```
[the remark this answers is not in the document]
ayrat555: @germsvel yep, Parity have an option to disable discovery

apps/ex_wire/test/ex_wire/kademlia/routing_table_test.exs
ayrat555: @germsvel  we pattern match here. if the right part does not have `filler_node`, `node` as first tuple elements, it will fail. but I can change it to assert

apps/ex_wire/test/integration/node_discovery_test.exs
ayrat555: @vyorkin yes, it is tagged

apps/ex_wire/test/ex_wire/kademlia/routing_table_test.exs
germsvel: Yeah, I know the pattern match will fail the test if it fails. But I tend to like the assertion because it makes it more clear what we are testing. Without the `assert`, on first glance, it seems like we forgot to finish the test. What do you think?

apps/ex_wire/lib/ex_wire/kademlia/server.ex
ayrat555: @germsvel we always get discovery nodes from routing table. 

In the first round we do not have any nodes in table so we pass bootnodes and add them ourselves.  

in consecutive rounds we use nodes from routing table that came from previous discovery rounds that we did not use yet

PR #115, opened 2018-05-29
Extract DEVp2p Wire Protocol from ExWire.Adapter.TCP

### LINKED ISSUE #72: Refactor TCP (Devp2p Wire Protocol implementation)
Currently, we have part of the [Devp2p Wire Protocol](https://github.com/ethereum/wiki/wiki/%C3%90%CE%9EVp2p-Wire-Protocol) implemented in `lib/ex_wire/adapter/tcp.ex` in the `ExWire` app. 

This GenServer is managing the TCP connection, the aggregation of TCP data into wire protocol "packets" (e.g. GetBlockHeaders packet), as well as initializing the [session](https://github.com/ethereum/wiki/wiki/%C3%90%CE%9EVp2p-Wire-Protocol#session-management) of the Devp2p Wire Protocol (which includes crypto and protocol handshakes, and the sending and receiving an initial Hello message).

It would be nice if we could separate those concerns into modules that handle the required logic. Whether or not they have to live in the same GenServer, I'm not sure.

We should also consider what to do when a peer disconnects from our `TCP` process. Currently, we set the session state to "active: false", but we keep the elixir process alive. We should try to either (a) reconnect, or (b) kill the elixir process so that it isn't just wasting resources when it will never reconnect.

germsvel:
Closes #72

Why extract logic from the TCP module?
======================================

The `TCP` module handles too many concerns at the moment: it is a
GenServer managing the TCP connection, it handles the aggregation of TCP
data into wire protocol "packets" (e.g. GetBlockHeaders packet), it
performs the crypto handshake, as well as initializing the session of
the Devp2p Wire Protocol (which includes the protocol handshake by
sending and receiving an initial Hello message).

It would be better if the GenServer could simply delegate most
responsibilities to other modules. In this way, we can separate a
runtime concern (i.e. what code should run in which process) from the
logic of our application.

To that end, we extract the logic related to DEVp2p Wire Protocol. We do
not yet extract the logic related to RLPx, which includes the crypto
handshake, as well as the framing of all messages after the handshakes
have been successful. But this may be a good next step.

What was done?
=============

We create a `ExWire.DEVp2p` module to handle all logic associated with
the DEVp2p Wire Procotol. The handshake (`Hello` message) and the notion
of an active session previously lived in the TCP module.

We do not yet create a GenServer to manage the actual handshake with the
peer. For now, all we need is a module that can handle all of the logic.
By passing the state in a `DEVp2p.Session` struct, we can test and
ensure that we can handle a complete handshake without having to keep
the state in a GenServer. In this way, we separate the runtime concerns
(e.g. which process to keep the state in) from the logic of the wire
protocol.

While extracing the `DEVp2p` responsibilities we refactored some
of the `TCP` code:

- Extracting logic into smaller functions, and creating convenience
`GenServer` client functions.

- Cleaning up framing and unframing.

- Removing responsibility for handling packets out of TCP. Once the
GenServer has handled the crypto handshake (sending auth and receiving
ack), it should simply forward any packets to subscribers. It should not
handle the packets themselves.


apps/ex_wire/lib/ex_wire/dev_p2p.ex
```diff
@@ -0,0 +1,156 @@
...
+  or `ExWire.Packet.Pong`.
+
+  An `ExWire.DEVp2p.Session` is required as the first argument in order to
+  properly update the session based on the message received.
+  """
+  @spec handle_message(Session.t(), struct()) ::
+          {:error, :handshake_incomplete, Session.t()}
+          | {:ok, ExWire.Pakcet.handle_response(), Session.t()}
+  def handle_message(session, message) do
+    case session_active?(session) do
```
masonforest: why not `if session_active?(session) do`?

germsvel: No particular reason. It could totally work as an `if, else`. I think I'm just used to thinking in `case` statements and pattern matching in elixir. 

I'll update this to use `if else`. I think it makes more sense.

apps/ex_wire/lib/ex_wire/adapter/tcp.ex
```diff
@@ -23,45 +23,69 @@ defmodule ExWire.Adapter.TCP do
...
-      active: false,
       subscribers: subscribers
     })
   end
 
   @doc """
-  Initialize by opening up a `gen_tcp` connection to given host and port.
+  Sends auth message to a peer
+  """
+  def send_auth_message(pid, auth_data) do
```
ayrat555: I'm thinking about public methods and genserver callbacks. Maybe we should separate them (public methods before callbacks)? There is also post by Dave Thomas - https://pragdave.me/blog/2017/07/13/decoupling-interface-and-implementation-in-elixir.html. He suggests to separate  public apis and genserver calbackes into different modules. I tried to do it in `ExWire.Kademlia.Server` and `ExWire.Kademlia` modules.

germsvel: I totally agree. Glad you mentioned that. That's what I was trying to do here. From this function all the way to `disconnect/1` on line 71 are public interface for the TCP module. The rest (from `init` down) are callbacks.

I know this function `send_auth_message/2` is only used internally currently, but I think it could be used externally, which is why I put it among the public interface of TCP. If you feel strongly about it, I can move it to the server functions.

I am also not opposed to separate them into two different modules. I tend to do that only if the GenServers are complex (which I do think this one is). I didn't do it because I was hoping that we would keep simplifying this GenServer since most of the logic on this file need not live here. What do you think?

germsvel: After some more thought. I like the idea of extracting this into a `Server` module. It cleans up things a bit. If it turns out that we don't need it later, we can always change it then. 

Updated in cb19982ed69ad0bab9c26774f452d6e872bd36a4

PR #116, opened 2018-05-29
Update exth_crypto, refactor evm & mpt

vyorkin:
* Extract `hex_to_int` to `ExthCrypto.Math`, update `hex_to_bin`
* Refactor MMPT & EVM tests

### DIFF
--- a/apps/evm/test/evm_test.exs
+++ b/apps/evm/test/evm_test.exs
@@ -1,6 +1,10 @@
 defmodule EvmTest do
+  import ExthCrypto.Math, only: [hex_to_bin: 1, hex_to_int: 1]
+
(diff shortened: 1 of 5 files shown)


apps/evm/test/evm_test.exs
```diff
@@ -200,18 +212,6 @@ defmodule EvmTest do
     "#{@ethereum_common_tests_path}/VMTests/vm#{Macro.camelize(Atom.to_string(group))}/#{name}.json"
   end
 
-  def hex_to_binary(string) do
```
vyorkin: This is the reason I've extracted it – it's just same fn copy-pasted

Issue #103, opened 2018-05-25
Handle incoming connections for TCP


germsvel:
Ou current TCP implementation (along with the RLPx crypto handshake and the DEVp2p protocol handshake) only handles outgoing connections. That is, we can connect to an external node in order to sync our information. But in order for us to have a full node, we need to support the ability for other nodes to connect to ours so they can ask for information to sync their data. 

Even though this could be done separately from #72, and thus it is technically not blocked, we should probably finish that before starting on this.

PR #118, opened 2018-05-29
Do not pass unneeded peer's uri host as argument

germsvel:
While doing work for https://github.com/poanetwork/mana/issues/103, I discovered the following:

There were several functions that were taking the peer's host as an argument solely so they could print it in a `Logger` call. This seems unnecessary since we should only pass the things that are needed to the functions.

It was also confusing since (a) it made it seem like functions needed to know about the peer when they didn't, and (b) it adds a lot of misleading noise when running tests -- logging that something is being sent over the network when in fact only a struct is being built.

### DIFF
--- a/apps/ex_wire/lib/ex_wire/adapter/tcp/server.ex
+++ b/apps/ex_wire/lib/ex_wire/adapter/tcp/server.ex
@@ -153,7 +153,7 @@ defmodule ExWire.Adapter.TCP.Server do
       my_nonce: my_nonce
     } = state
 
-    case Handshake.try_handle_ack(data, auth_data, my_ephemeral_private_key, my_nonce, peer.host) do
+    case Handshake.try_handle_ack(data, auth_data, my_ephemeral_private_key, my_nonce) do
       {:ok, secrets, frame_rest} ->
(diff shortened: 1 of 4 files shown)


apps/ex_wire/lib/ex_wire/adapter/tcp/server.ex
```diff
@@ -267,6 +266,8 @@ defmodule ExWire.Adapter.TCP.Server do
         peer.remote_id
       )
 
+    Logger.debug("[Network] Generating EIP8 Handshake for #{peer.host}")
```
germsvel: I added this in here in place of the one that read, 

```elixir
 Logger.debug("[Network] Sending EIP8 Handshake to #{remote_addr}")
```

in `EIP8.wrap_eip_8` function. 

I'm not sure how much value these are really providing. I'm in favor of removing this if others agree that there isn't much value in this.

germsvel: Makes sense. I think we'll have to think about how to do that holistically, so that we can turn all logging on or off based on flags provided.

PR #122, opened 2018-05-31
Fix Secrets.derive_secrets/7 function

germsvel:
Summary
=======

The `Secrets.derive_secrets/7` function was throwing an `ArgumentError` when trying to generate a shared secret. It was untested and for all intents an purposes, it was unused. So the bug was never caught. As part of my work on #103, I realized that this function was causing issues when trying to derive secrets for the encrypted auth exchange. 

The function failed when trying to generate an ephemeral shared secret. In order to generate that shared secret, we need our node's ephemeral private key and the remote node's ephemeral public key.

The function was trying to pass a "raw" ephemeral public key instead. And when looked at closely, it wasn't even passing the correct "raw" key. Instead it was using a function that transformed the `remote_ephemeral_public_key` from raw into `der` (i.e. `ExthCrypto.Key.raw_to_der()`).

We also add a few tests to ensure that this code is working. Though the tests are of some value (they caught the error), they are not perfect since they are simply testing that the values returned are binaries or references. Better tests would involve using the secrets to do something that ensures the secrets are indeed correct. Unfortunately, that's more than I can decipher right now. So for now, I think having some tests is better than no tests.

### DIFF
--- a/apps/ex_wire/lib/ex_wire/handshake/handshake.ex
+++ b/apps/ex_wire/lib/ex_wire/handshake/handshake.ex
@@ -194,7 +194,7 @@ defmodule ExWire.Handshake do
     shared_secret = ECDH.generate_shared_secret(my_static_private_key, her_static_public_key)
 
     # Build a nonce unless given
-    nonce = if nonce, do: nonce, else: ExthCrypto.Math.nonce(@nonce_len)
+    nonce = if nonce, do: nonce, else: new_nonce()
 
     # XOR shared-secret and nonce
     shared_secret_xor_nonce = ExthCrypto.Math.xor(shared_secret, nonce)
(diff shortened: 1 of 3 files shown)


apps/ex_wire/lib/ex_wire/handshake/handshake.ex
```diff
@@ -327,4 +327,9 @@ defmodule ExWire.Handshake do
         {:invalid, reason}
     end
   end
+
+  @spec new_nonce() :: binary()
+  def new_nonce do
```
germsvel: Extracted this into a function (a public one) because I'm having to use this from other places -- mostly tests. I could simply use `ExthCrypto.Math.nonce/1` from the tests, but since the value of `@nonce_len` lives in this module, having a random `32` in the tests can lead to confusion. So I think it best to have a function in this module that generates a nonce of the appropriate length for a handshake.

Issue #124, opened 2018-05-31
Pass refund general state tests


masonforest:
We can add these tests by updating `@passing_tests_by_group` [here](https://github.com/poanetwork/mana/blob/master/apps/blockchain/test/blockchain/state_test.exs#L9):

    @passing_tests_by_group %{
      stRefundTest: :all,
      stExample: [:add11],
      stCallCodes: [
        :callcall_00,
        :callcode_checkPC
      ]

PR #127, opened 2018-06-01
Fix EVM tests

### LINKED ISSUE #109: Fix EVM tests
We need to test the whole "post" state, not just the "storage" part of it: https://github.com/poanetwork/mana/blob/master/apps/evm/test/evm_test.exs#L232

EDIT1: Forgot to mention that we need to check the `callcreates` and `out` too, according to http://ethereum-tests.readthedocs.io/en/latest/test_types/vm_tests.html#the-callcreates-section and http://ethereum-tests.readthedocs.io/en/latest/test_types/vm_tests.html#the-gas-and-output-keys

### LINKED ISSUE #126: Fix refund on selfdestruct
image

Consider creating a new "refund account" if it doesn't exist (it should be shadow-created if it receives eth, this is already implemented).

image

We need to transfer the balance of the "suicided" account to it.

May be related:
https://github.com/paritytech/parity/pull/4757
https://github.com/ethereum/EIPs/issues/168
https://github.com/ethereum/EIPs/issues/161
https://github.com/ethereum/EIPs/issues/169

Resolving this issue should fix the `push32AndSucide` test (and maybe other failing "suicide" test too).

vyorkin:
* Rename `suicide` op to `selfdestruct` according to [EIP-6](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-6.md).

* Transfer the balance of the self-destructed account to the "refund account" (its address is the top 20-bytes on VM's stack). Refund account is "shadow-created" if it doesn't exist.

* Refactor and fix the `evm_test.exs`:
  * Remove the `code` field from the caller's account (see http://ethereum-tests.readthedocs.io/en/latest/test_types/vm_tests.html#the-exec-section).
  * Exclude caller's account from the actual state if it isn't in the "pre" or "post" addresses.
  * Don't ignore empty state values in EVM tests.

* Update `EVM.Address.new` to use only the last `@size` of bytes for address construction:
this is important because in VM operations we're getting addresses from the VM stack as integers that can be of any size.

* Replace the doctest for `SELFDESTRUCT` operation in `EVM.System` module with a regular ExUnit test because it's getting quite complex.

* Added `:sender` condition to the `EVM.Debugger.Breakpoint` for convenience.

* Slightly refactor and fix the `new_account` fn (add empty `code` field) in the `EVM.Interface.Mock.MockAccountInterface` module.
* Fix `vm_test.exs` (add empty `code` field).

resolves #126, resolves #109

### DIFF
--- a/apps/evm/test/evm_test.exs
+++ b/apps/evm/test/evm_test.exs
@@ -1,9 +1,7 @@
 defmodule EvmTest do
   import ExthCrypto.Math, only: [hex_to_bin: 1, hex_to_int: 1]
 
-  alias MerklePatriciaTree.Trie
   alias ExthCrypto.Hash.Keccak
-  alias EVM.Interface.AccountInterface
   alias EVM.Interface.Mock.{MockAccountInterface, MockBlockInterface}
 
   use ExUnit.Case, async: true
@@ -47,9 +45,19 @@ defmodule EvmTest do
 
   test "Ethereum Common Tests" do
     for {test_group_name, _test_group} <- @passing_tests_by_group do
-      for {_test_name, test} <- passing_tests(test_group_name) do
+      for {test_name, test} <- passing_tests(test_group_name) do
         {gas, sub_state, exec_env, _} = run_test(test)
-        assert_state(test, exec_env.account_interface)
+
+        context = %{
+          test_name: test_name,
+          account_interface: exec_env.account_interface,
+          addresses: %{
+            pre: get_addresses(test, "pre"),
+            post: get_addresses(test, "post")
+          }
+        }
+
+        assert_state(test, context)
 
         if test["gas"] do
           assert hex_to_int(test["gas"]) == gas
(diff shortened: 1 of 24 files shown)


apps/evm/test/evm_test.exs
```diff
@@ -47,9 +45,19 @@ defmodule EvmTest do
 
   test "Ethereum Common Tests" do
     for {test_group_name, _test_group} <- @passing_tests_by_group do
-      for {_test_name, test} <- passing_tests(test_group_name) do
+      for {test_name, test} <- passing_tests(test_group_name) do
         {gas, sub_state, exec_env, _} = run_test(test)
-        assert_state(test, exec_env.account_interface)
+
+        context = %{
```
vyorkin: I think this whole test needs to be refactored, just maybe not as part of this PR.

vyorkin:
Thanks! I'll double check if I'm running all the EVM tests from the _ethereum common tests_ & fix that.

vyorkin:
Ah, @masonforest have created a separate issue to track this #124 so we're good.

vyorkin:
ok, rebased

vyorkin:
@masonforest yes! its by mistake, thanks for pointing this out

vyorkin:
Ok, I'll merge this and refactor the `AccountInterface` to add the account to the `selfdestruct_list` instead of deleting it right away

Issue #130, opened 2018-06-02
Implement INVALID


ayrat555:
https://eips.ethereum.org/EIPS/eip-141

![https://i.imgur.com/omqEdEP.png](https://i.imgur.com/omqEdEP.png)

PR #132, opened 2018-06-04
add invalid operation

ayrat555:
resolves https://github.com/poanetwork/mana/issues/130


vyorkin:
yap, from what I've just read I think this is correct

apps/evm/lib/evm/operation/metadata/system.ex
```diff
@@ -42,6 +42,15 @@ defmodule EVM.Operation.Metadata.System do
                       output_count: 1,
                       group: :system
                     },
+                    %{
+                      id: 0xFD,
+                      description:
+                        "Designated invalid instruction.  It can be used to abort the execution.",
```
vyorkin: Two spaces here :)

Issue #131, opened 2018-06-02
Implement RETURNDATASIZE, RETURNDATACOPY


ayrat555:
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-211.md

![https://i.imgur.com/TAkkWFJ.png](https://i.imgur.com/TAkkWFJ.png)

PR #135, opened 2018-06-04
 implement RETURNDATASIZE env op 

ayrat555:
IMG

- add `last_return_data` field to `machineState` struct to track last op output data
- implement `RETURNDATASIZE`

Part of https://github.com/poanetwork/mana/issues/131


apps/evm/lib/evm/operation.ex
```diff
@@ -360,6 +363,11 @@ defmodule EVM.Operation do
         do: %{base_machine_state | stack: op_result[:stack]},
         else: base_machine_state
 
+    next_machine_state =
+      if op_result[:last_return_data],
+        do: %{next_machine_state | last_return_data: op_result[:last_return_data]},
+        else: next_machine_state
```
vyorkin:
```
do: %{next_machine_state | last_return_data: nil}
```

PR #137, opened 2018-06-04
add RETURNDATACOPY op

ayrat555:
with https://github.com/poanetwork/mana/pull/135 resolves https://github.com/poanetwork/mana/issues/131


apps/evm/lib/evm/operation/environmental_information.ex
```diff
@@ -286,12 +282,66 @@ defmodule EVM.Operation.EnvironmentalInformation do
...
+    normalized_data =
+      cond do
+        is_nil(data) -> []
+        is_list(data) -> data
+        true -> [data]
+      end
+
+    normalized_data
+    |> Enum.take(length)
+    |> Enum.reduce(<<>>, fn el, acc ->
```
vyorkin: an alternative way:
```
1..10 |> Enum.take(5) |> Enum.into(<<>>, fn x -> <<x>> end)
```
(but the current one is good also)

apps/evm/lib/evm/machine_state.ex
```diff
@@ -30,7 +30,7 @@ defmodule EVM.MachineState do
           memory: memory,
           active_words: integer(),
           stack: Stack.t(),
-          last_return_data: [EVM.val()]
+          last_return_data: [EVM.val()] | nil | EVM.val()
```
germsvel: Is there a way to normalize this data in this module to always return a list (could be empty), so that you don't have to normalize it in `read_return_data` in the other file?

ayrat555: done

Issue #129, opened 2018-06-02
Implement REVERT


ayrat555:
EIP https://github.com/ethereum/EIPs/blob/master/EIPS/eip-140.md

![revert](https://i.imgur.com/mvDhq0p.png)

PR #144, opened 2018-06-06
revert op

ayrat555:
Changes:
- Implemented `revert` op code (resolves https://github.com/poanetwork/mana/issues/129)
- Fix bug with returning modified state after exceptional exit. we should return original state.


apps/evm/test/evm/operation/system_test.exs
```diff
@@ -21,4 +21,32 @@ defmodule EVM.Operation.SystemTest do
       assert Map.get(accounts, selfdestruct_address) == nil
     end
   end
+
+  describe "revert" do
```
germsvel: Looking at these two tests (the one for `revert` on line 26 and the one for `return` on line 40), I can see that they should be have differently, but their assertions are the same. Is there a way we could write these tests so that we can see that difference clearly in the assertions? I don't know enough of the VM code to suggest a possible way to do this, but I'm curious if it can be done.

apps/evm/test/evm/operation/system_test.exs
```diff
@@ -21,4 +21,32 @@ defmodule EVM.Operation.SystemTest do
...
+    test "halts execution reverting state changes but returning data and remaining gas" do
+      machine_code = MachineCode.compile([:push1, 3, :push1, 5, :revert, :push1, 10, :pop])
+      exec_env = %ExecEnv{machine_code: machine_code}
+      machine_state = %MachineState{program_counter: 0, gas: 24, stack: []}
+      substate = %SubState{}
+
+      {updated_machine_state, _, updated_exec_env, _} = VM.exec(machine_state, substate, exec_env)
```
vyorkin: maybe it makes sense the check the output data too

apps/evm/test/evm/operation/system_test.exs
vyorkin: yup, so as I understand the difference is that `:return` doesn't not **revert** the state changes

ayrat555: I will find machine code  that changes exection env so assertions will be different

apps/evm/test/evm/operation/system_test.exs
ayrat555: @vyorkin  done

apps/evm/test/evm/operation/system_test.exs
ayrat555: @germsvel @vyorkin Now I check state after SSTORE operation. It should change state in case of RETURN op code

PR #156, opened 2018-06-08
Fix Debugger memory contents inspection

### LINKED ISSUE #146: Fix Debugger memory contents inspection
Debugger and its REPL is helpful for debugging contract execution.
The problem is that it's `m` (memory inspection) command sometimes failed to display the memory contents because of using the wrong function `String.to_charlist()`. My first guess is that it should be the `:binary.bin_to_list()` instead.

vyorkin:
resolves #146

### DIFF
--- a/apps/evm/lib/evm/debugger.ex
+++ b/apps/evm/lib/evm/debugger.ex
@@ -423,22 +423,25 @@ defmodule EVM.Debugger do
     size_width =
       if total_size > 0, do: total_size |> :math.log10() |> :math.ceil() |> round, else: 0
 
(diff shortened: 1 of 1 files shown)


apps/evm/lib/evm/debugger.ex
```diff
@@ -518,6 +526,17 @@ defmodule EVM.Debugger do
     {machine_state, sub_state, exec_env}
   end
 
+  @spec to_ascii(String.t()) :: String.t()
+  defp to_ascii(codepoint) do
+    is_printable =
+      codepoint
+      |> :binary.bin_to_list()
```
vyorkin: The actual fix

Issue #151, opened 2018-06-07
Implement Delegatecall op code


ayrat555:
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7.md
![img](https://i.imgur.com/z2lnQiD.png)

Issue #119, opened 2018-05-31
fix Call/callcode operation in EVM


ayrat555:
- fix all todos
- fix storage

PR #157, opened 2018-06-08
Delegatecall op

ayrat555:
resolves https://github.com/poanetwork/mana/issues/151
resolves https://github.com/poanetwork/mana/issues/119

Also this PR makes call operations more understandable and closer to the Yellow Paper

Changes:

- Move message call logic to `EVM.MessageCall` module
- All call operations (`call`, `delegatecall`, `staticcall`) differ only in parameters to the `Theta` function described in section 8 of the yellow paper. So now all this operation use `EVM.MessageCall`.
- Implement `delegatecall`


apps/evm/lib/evm/message_call.ex
```diff
@@ -0,0 +1,129 @@
...
+    # first transitional state
+    message_call = transfer_gas_to_recipient(message_call)
+
+    message_call
+    |> prepare_call_execution_env()
+    |> execute_call(message_call)
+    |> update_state(message_call)
+  end
+
+  def transfer_gas_to_recipient(message_call) do
```
vyorkin: The function is named `transfer_gas_to_recepient`, but as I can see you're transferring `wei` from the current sender to the "recipient"

ayrat555: @vyorkin I guess you a mistake

PR #163, opened 2018-06-12
Fix inbound handshake & add naive TCP inbound implementation

germsvel:
Summary
=======

This commit handles two things that are interrelated. It adds a naive implementation for an inbound TCP connection and fixes the function that handles the main crypto handshake when it is initiated by the remote peer. This is all part of supporting #103.

Fix inbound handshake
=====================

The main function that was called when a remote node initiated the handshake (sending an auth message) was `Handshake.try_handle_auth/4`.

`Handshake.try_handle_auth/4`, however, was broken. It had an unrealistic function signatures (assuming we would have peer data that we can only obtain after decrypting auth message). It also called `Handshake.build_ack_resp/2` incorrectly, passing incorrect argument types and values. This function was untested and practically unused, so the errors were not apparent at first.

We fix that function, rename it to `Hanshake.handle_auth` and we update the calling code (in TCP.Server) to use the updated version.

Naive implementation of incoming TCP
===================================

We also add a naive implementation for an incoming TCP connection. We say naive because in this implementation only one client can connect with our server. The full implementation needs to separate the creation of the `listen_socket` from the socket created when we accept messages. We will likely have to span a new process for each connection we receive.

We also update the `TCP.Server` code since most functions assume the presence of a peer. At this point, for our naive implementation of the incoming TCP connection, we do not have the notion of a peer when the connection is incoming. With a full implementation, we will have the public key (node id) of the remote peer (obtained from their Auth message), but we will still not have a `peer` in the sense that is defined in `ExWire.Struct.Peer` and which is used in outgoing TCP connections.

Clean up
========

In addition to the above, we also do some clean-up. The main one worth mentioning is the movement of handshake files.

The `handshake.ex` file was under `lib/ex_wire/handshake/handshake.ex`, but the module simply defined `ExWire.Handshake`. The more common pattern is to put `ExWire.Handshake` directly on `lib/ex_wire/handshake.ex`, so we do that here.

### DIFF
--- a/apps/ex_wire/lib/ex_wire/adapter/tcp/server.ex
+++ b/apps/ex_wire/lib/ex_wire/adapter/tcp/server.ex
@@ -28,9 +28,7 @@ defmodule ExWire.Adapter.TCP.Server do
   end
 
   def init(state = %{is_outbound: false}) do
-    new_state =
-      state
-      |> listen_via_tcp()
+    new_state = listen_via_tcp(state)
 
     accept_tcp_messages()
 
--- a/apps/ex_wire/lib/ex_wire/adapter/tcp/server.ex
+++ b/apps/ex_wire/lib/ex_wire/adapter/tcp/server.ex
@@ -27,6 +27,20 @@ defmodule ExWire.Adapter.TCP.Server do
     {:ok, new_state}
   end
 
+  def init(state = %{is_outbound: false}) do
+    new_state =
+      state
+      |> listen_via_tcp()
+
+    accept_tcp_messages()
+
+    {:ok, new_state}
+  end
+
+  def accept_tcp_messages do
+    GenServer.cast(self(), :accept_tcp_messages)
+  end
+
   @doc """
   Sends auth message to a peer
   """
@@ -82,11 +96,7 @@ defmodule ExWire.Adapter.TCP.Server do
     {:noreply, handle_acknowledgement_received(data, state)}
   end
 
-  def handle_info(
-        {:tcp, _socket, data},
-        state = %{is_outbound: false, my_ephemeral_key_pair: _my_ephemeral_key_pair}
-      ) do
-    # TODO: How do we set remote id?
+  def handle_info({:tcp, _socket, data}, state = %{is_outbound: false}) do
     {:noreply, handle_auth_message_received(data, state)}
   end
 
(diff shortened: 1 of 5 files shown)


apps/ex_wire/lib/ex_wire/adapter/tcp/server.ex
```diff
@@ -27,6 +27,20 @@ defmodule ExWire.Adapter.TCP.Server do
     {:ok, new_state}
   end
 
+  def init(state = %{is_outbound: false}) do
+    new_state =
+      state
+      |> listen_via_tcp()
```
ayrat555: Can we ignore pipeline operator with single method?

apps/ex_wire/lib/ex_wire/handshake.ex
```diff
@@ -143,6 +144,44 @@ defmodule ExWire.Handshake do
...
+          ExthCrypto.Key.public_key(),
+          ExthCrypto.Key.private_key(),
+          ExthCrypto.Key.public_key(),
+          nonce() | nil,
+          ExthCrypto.Key.key_pair() | nil
+        ) :: {binary(), ExthCrypto.Key.key_pair(), nonce()}
+  def build_encoded_auth_msg(
+        my_static_public_key,
+        my_static_private_key,
+        her_static_public_key,
```
germsvel: I forgot I wanted to update all of these to initiator and recipient as needed. I think it's far more clear, so I'll update those.

apps/ex_wire/lib/ex_wire/handshake.ex
```diff
@@ -143,6 +144,44 @@ defmodule ExWire.Handshake do
...
+  This message is ready to be sent to a peer to initiate the encrypted handshake.
+  """
+  @spec build_encoded_auth_msg(
+          ExthCrypto.Key.public_key(),
+          ExthCrypto.Key.private_key(),
+          ExthCrypto.Key.public_key(),
+          nonce() | nil,
+          ExthCrypto.Key.key_pair() | nil
+        ) :: {binary(), ExthCrypto.Key.key_pair(), nonce()}
+  def build_encoded_auth_msg(
```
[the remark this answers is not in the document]
germsvel: Yeah, that is something I want to do. There is a `Handshake` struct that was created but never used. So I was planning on using that later. Though I might make sense to just start using that.

apps/ex_wire/lib/ex_wire/adapter/tcp/server.ex
germsvel: Updated in 8e5f416df3e4c59c0632fbb56d42a3babcb605d1

apps/ex_wire/lib/ex_wire/handshake.ex
germsvel: Updated function names and simplified all of these parameters in da0fa1db209bacfdc535b888c766d8e076e1f0b5

vyorkin:
Thanks for such a detailed PR description, I can't say that I understand everything (because I'm trying to concentrate mostly on the blockchain & evm parts), but I've got the main idea

PR #164, opened 2018-06-13
Fix `GeneralStateTests/Call10` test

masonforest:
When making message calls the originator pays gas fees. Before we were
checking the sender's balance to ensure the account had enough gas.
Instead we now check the originator's balance.

https://ethereum.stackexchange.com/a/146

* Add Address.new_from_private_key for debugging
* Add and EVM logging function for debugging
* Comment tests that fail due to the change made

### DIFF
--- a/apps/blockchain/test/blockchain/state_test.exs
+++ b/apps/blockchain/test/blockchain/state_test.exs
@@ -77,7 +77,7 @@ defmodule Blockchain.StateTest do
     call_codes: [
       # :call_OOG_additionalGasCosts1,
       # :call_OOG_additionalGasCosts2,
-      # :callcall_00
+      :callcall_00
       # :callcall_00_OOGE,
       # :callcall_00_OOGE_valueTransfer,
       # :callcall_00_SuicideEnd,
@@ -1468,8 +1468,8 @@ defmodule Blockchain.StateTest do
       # :recursiveCreateReturnValue
     ],
     refund_test: [
-      :refund_TxToSuicide
-      # :refund_singleSuicide
+      :refund_TxToSuicide,
+      :refund_singleSuicide
 
       # :refund50_1,
       # :refund50_2,
--- a/apps/blockchain/test/blockchain/state_test.exs
+++ b/apps/blockchain/test/blockchain/state_test.exs
@@ -1468,7 +1468,7 @@ defmodule Blockchain.StateTest do
       # :recursiveCreateReturnValue
     ],
     refund_test: [
-      :refund_TxToSuicide,
+      :refund_TxToSuicide
       # :refund_singleSuicide
 
       # :refund50_1,
@@ -1942,7 +1942,7 @@ defmodule Blockchain.StateTest do
       # :static_refund_CallToSuicideTwice
     ],
     system_operations_test: [
-      :Call10
+      :Call10,
 
       # :ABAcalls0,
       # :ABAcalls1,
--- a/apps/blockchain/test/blockchain/state_test.exs
+++ b/apps/blockchain/test/blockchain/state_test.exs
@@ -77,7 +77,7 @@ defmodule Blockchain.StateTest do
     call_codes: [
       # :call_OOG_additionalGasCosts1,
       # :call_OOG_additionalGasCosts2,
-      :callcall_00
+      # :callcall_00
       # :callcall_00_OOGE,
       # :callcall_00_OOGE_valueTransfer,
       # :callcall_00_SuicideEnd,
@@ -1469,7 +1469,7 @@ defmodule Blockchain.StateTest do
     ],
     refund_test: [
       :refund_TxToSuicide,
-      :refund_singleSuicide
+      # :refund_singleSuicide
 
       # :refund50_1,
       # :refund50_2,
@@ -1942,13 +1942,14 @@ defmodule Blockchain.StateTest do
       # :static_refund_CallToSuicideTwice
     ],
     system_operations_test: [
+      :Call10
+
       # :ABAcalls0,
       # :ABAcalls1,
       # :ABAcalls2,
       # :ABAcalls3,
       # :ABAcallsSuicide0,
       # :ABAcallsSuicide1,
-      # :Call10,
       # :CallRecursiveBomb0,
       # :CallRecursiveBomb0_OOG_atMaxCallDepth,
       # :CallRecursiveBomb1,
--- a/apps/evm/lib/evm/logger.ex
+++ b/apps/evm/lib/evm/logger.ex
@@ -10,33 +10,19 @@ defmodule EVM.Logger do
 
   @spec log_state_in_parity_format(EVM.Operation.Metadata.t(), MachineState.t()) :: nil
   def log_state_in_parity_format(operation, machine_state) do
-    operation_string =
-      Atom.to_string(operation.sym)
-      |> String.upcase()
-      |> String.pad_leading(8)
+    log_opcode_and_gas_left_in_parity_format(operation, machine_state)
+    log_inputs_in_parity_format(operation, machine_state)
+  end
 
-    opcode_string =
-      operation.id
-      |> :binary.encode_unsigned()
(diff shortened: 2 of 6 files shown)


apps/evm/lib/evm/message_call.ex
```diff
@@ -41,13 +41,13 @@ defmodule EVM.MessageCall do
...
-    sender_balance =
+    originator_balance =
       AccountInterface.get_account_balance(
         message_call.current_exec_env.account_interface,
-        message_call.sender
+        message_call.originator
       )
 
-    sender_balance >= message_call.execution_value
+    originator_balance >= message_call.execution_value
```
ayrat555: Also I guess we should compare with value, not with execution value

apps/evm/lib/evm/vm.ex
```diff
@@ -144,4 +144,41 @@ defmodule EVM.VM do
...
+      |> Stream.with_index
+      |> Enum.each(fn({value, i}) ->
+        value_string = if value == 0,
+        do: "0", else: (
+          value
+                       |> :binary.encode_unsigned()
+                       |> Base.encode16(case: :lower)
+                       |> String.trim_leading("0"))
+
+IO.puts "       | #{i}: 0x#{value_string}"
```
ayrat555: IO.puts

apps/evm/lib/evm/vm.ex
```diff
@@ -144,4 +144,41 @@ defmodule EVM.VM do
...
+      |> String.pad_trailing(2)
+
+    program_counter = machine_state.program_counter + 1
+    program_counter_string = "0x#{program_counter}"
+      |> :binary.encode_unsigned()
+      |> Base.encode16(case: :lower)
+      |> String.trim_leading("0")
+      |> String.pad_trailing(3)
+
+    IO.puts "[#{program_counter_string}][#{operation_string}(0x#{opcode_string}) Gas Left: #{machine_state.gas})"
```
ayrat555: Maybe we could use logger instead of puts

apps/blockchain/test/blockchain/state_test.exs
```diff
@@ -77,7 +77,7 @@ defmodule Blockchain.StateTest do
     call_codes: [
       # :call_OOG_additionalGasCosts1,
       # :call_OOG_additionalGasCosts2,
-      :callcall_00
+      # :callcall_00
```
ayrat555: Why this test began to fail?

apps/evm/lib/evm/vm.ex
masonforest: This is formatted exactly the same as Parity's `evm-debug` output. That makes it really easy to compare output when debugging differences between parity and our implementation. I think we can change the formatting of `Logger` but that might change it globally. We should start to think about logging at some point though. It's be nice to have log levels like the other clients do at some point...

apps/blockchain/test/blockchain/state_test.exs
masonforest: Still not sure. I thought I'd get this PR merged and then fix this test in a follow-on PR

ayrat555:
@masonforest it looks like parity also compares with sender's balance.

apps/evm/lib/evm/vm.ex
```diff
@@ -144,4 +144,41 @@ defmodule EVM.VM do
...
+      |> String.pad_trailing(2)
+
+    program_counter = machine_state.program_counter + 1
+    program_counter_string = program_counter
+      |> :binary.encode_unsigned()
+      |> Base.encode16(case: :lower)
+      |> String.trim_leading("0")
+      |> String.pad_trailing(3)
+
+    IO.puts "[0x#{program_counter_string}][#{operation_string}(0x#{opcode_string}) Gas Left: #{machine_state.gas})"
```
germsvel: Should we make this `Logger.debug` or something like that since it's for debugging?

apps/evm/lib/evm/vm.ex
```diff
@@ -144,4 +144,41 @@ defmodule EVM.VM do
...
+      |> String.pad_leading(8)
+
+    opcode_string = operation.id
+      |> :binary.encode_unsigned()
+      |> Base.encode16(case: :lower)
+      |> String.trim_leading("0")
+      |> String.pad_trailing(2)
+
+    program_counter = machine_state.program_counter + 1
+    program_counter_string = program_counter
```
germsvel: Are you using the code formatter? I believe it would format these by putting `program_counter` on the next line, 

```
program_counter_string = 
  program_counter
  |> :binary.encode_unsigned()
# rest
```

apps/evm/lib/evm/vm.ex
```diff
@@ -144,4 +144,41 @@ defmodule EVM.VM do
...
+      |> Stream.with_index
+      |> Enum.each(fn({value, i}) ->
+        value_string = if value == 0,
+        do: "0", else: (
+          value
+                       |> :binary.encode_unsigned()
+                       |> Base.encode16(case: :lower)
+                       |> String.trim_leading("0"))
+
+        IO.inspect "       | #{i}: 0x#{value_string}"
```
germsvel: Should we make this `Logger.debug`?

apps/evm/lib/evm/vm.ex
```diff
@@ -144,4 +144,41 @@ defmodule EVM.VM do
 
     {final_machine_state, n_sub_state, n_exec_env}
   end
+
+  def log(operation, machine_state) do
```
germsvel: There seems to be a lot of stuff going on in this function. Do you think it might make sense to create a debugging module that allows us to call individual functions? Not sure if it's overkill, but just a suggestion.

masonforest:
Ohhh that actually makes sense. I was thinking it was checking whether the sender had enough funds to pay for gas. This didn't make sense because contracts can't pay gas fees. This is actually checking if the sender has enough tokens to make cover a transfer-and-call. I changed `originator` back to `sender` like you suggested and the tests that started failing went back to passing as well 👍

apps/evm/lib/evm/vm.ex
masonforest: Good call!

I extracted this [out into it's own module](https://github.com/poanetwork/mana/pull/164/commits/28f4533da5a3c83fb015fc4ad2750c612fa369b3).

apps/evm/lib/evm/logger.ex
```diff
@@ -0,0 +1,60 @@
...
+  alias EVM.{Operation, MachineState}
+
+  @doc """
+  This function logs state in the same format as Parity's `evm-debug` function. This makes comparing implementations and debugging easier.
+
+  `cargo test --features "json-tests evm/evm-debug-tests" --release -- BlockchainTests_GeneralStateTest_stSystemOperationsTest --nocapture`
+  """
+
+  @spec log_state_in_parity_format(EVM.Operation.Metadata.t(), MachineState.t()) :: nil
+  def log_state_in_parity_format(operation, machine_state) do
```
ayrat555: As @germsvel mentioned above this method too big and can be divided

masonforest: ah, yep! Done: [Refactor `EVM.Logger`](https://github.com/poanetwork/mana/pull/164/commits/29a6ef8f809f4703c46f6716b8d1e636d569e9f6)

apps/evm/lib/evm/logger.ex
```diff
@@ -10,33 +10,19 @@ defmodule EVM.Logger do
...
-      program_counter
-      |> :binary.encode_unsigned()
-      |> Base.encode16(case: :lower)
-      |> String.trim_leading("0")
-      |> String.pad_trailing(3)
+    log_opcode_and_gas_left_in_parity_format(operation, machine_state)
+    log_inputs_in_parity_format(operation, machine_state)
+  end
 
+  defp log_opcode_and_gas_left_in_parity_format(operation, machine_state) do
```
vyorkin: Maybe just `log` or smth, idk, its just too long 😂

PR #165, opened 2018-06-13
Handle multiple inbound tcp connections

germsvel:
I believe this final issues closes https://github.com/poanetwork/mana/issues/103

Summary
=========

Previously we were only accepting one inbound tcp connection. This commit updates the previous code so that we can handle multiple connections.

What was done?
==============

Since inbound connections are pretty self-contained, we create a supervision tree (under the main ex wire supervisor) that is separate from other trees. There is no reason why a failure on this supervision tree should crash, say, the node discovery supervision tree.

The top-level supervisor for the tcp inbound communication is the `TCPListeningSupervisor`. This supervisor supervises an `TCP.Listener` and a `TCP.InboundConnectionsSupervisor`.

The `TCP.Listener` is a process responsible for opening the tcp connection on a given port. Once a connection is accepted, it will hand over control of the connection to another process. Those processes are started and supervised by the `TCP.InboundConnectionSupervisor`. In this way, we can have a single process for every inbound connection that is established.

### DIFF
--- a/apps/ex_wire/lib/ex_wire/tcp/listener.ex
+++ b/apps/ex_wire/lib/ex_wire/tcp/listener.ex
@@ -1,0 +1,46 @@
+defmodule ExWire.TCP.Listener do
+  @moduledoc """
+  Module responsible for opening a tcp listen socket to listen for incoming
+  connections. Once a connection is accepted, it will hand off control of the
+  accepted socket connection to a separate process.
+  """
+
+  use GenServer
+
+  alias ExWire.TCP
+
+  def start_link(args) do
+    port = Keyword.fetch!(args, :port)
+    name = Keyword.fetch!(args, :name)
+
+    GenServer.start_link(__MODULE__, port, name: name)
+  end
+
+  def init(port) do
+    {:ok, listen_socket} = TCP.listen(port)
+
+    accept_tcp_connection()
+
+    {:ok, %{listen_socket: listen_socket}}
+  end
+
+  @doc """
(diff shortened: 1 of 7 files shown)


apps/ex_wire/lib/ex_wire/tcp/listener.ex
```diff
@@ -0,0 +1,46 @@
...
+
+  @doc """
+  Accepts a connection, and gives control of the connection to a separate process
+  that will henceforth handle that tcp connection.
+  """
+  def handle_cast(:accept_tcp_connection, state = %{listen_socket: listen_socket}) do
+    {:ok, socket} = TCP.accept_connection(listen_socket)
+    {:ok, pid} = TCP.InboundConnectionsSupervisor.new_connection_handler()
+
+    TCP.hand_over_control(socket, pid)
```
ayrat555: status of these two operations are not  being checked. they can fail. is it ok?

germsvel: That's a really good question. I'll do some due diligence on that. My original thought was that we don't care if these fail because that'll just drop the connection in the other process. The `listener` process should not fail because of bad response. But... now that I'm looking at this, if the hand over fails, then I want to make sure this process would not get the messages in its mailbox, so I'll do a little digging.

germsvel: We decided to make sure the hand-over of control is successful before accepting messages. Changes in 8fde0645d358917ba545d2a6571b3a28c3abc1b1

PR #168, opened 2018-06-14
[WIP] Nested calls

ayrat555:


apps/evm/lib/evm/gas.ex
```diff
@@ -351,28 +351,44 @@ defmodule EVM.Gas do
...
+        machine_state,
         exec_env
       ) do
     to_address = Address.new(to_address)
 
-    @g_call + call_value_cost(value) + new_account_cost(exec_env, to_address) + gas_limit
+    gas_limit = if exec_env.stack_depth == 1024, do: 0, else: gas_limit
+
+    call_value =
+      if exec_env.stack_depth == 1024 && value != 0,
```
masonforest: `max_stack_depth` should probably go in a config file at some point. Maybe just put it in a module var for now?

apps/evm/lib/evm/gas.ex
```diff
@@ -351,28 +351,44 @@ defmodule EVM.Gas do
...
   """
-  def operation_cost(:sstore, [key, new_value], _machine_state, exec_env) do
-    old_value = ExecEnv.get_storage(exec_env, key)
-
-    if old_value == 0 && new_value != 0 do
-      @g_sset
-    else
-      @g_sreset
+  def operation_cost(:sstore, params = [key, new_value], _machine_state, exec_env) do
+    case ExecEnv.get_storage(exec_env, key) do
```
masonforest: I wish we could find a way to simplify this branching logic but it looks like this is as good as we are going to get 😕

apps/evm/lib/evm/gas.ex
```diff
@@ -351,28 +351,44 @@ defmodule EVM.Gas do
...
         :call,
         [gas_limit, to_address, value, _in_offset, _in_length, _out_offset, _out_length],
-        _machine_state,
+        machine_state,
         exec_env
       ) do
     to_address = Address.new(to_address)
 
-    @g_call + call_value_cost(value) + new_account_cost(exec_env, to_address) + gas_limit
+    gas_limit = if exec_env.stack_depth == 1024, do: 0, else: gas_limit
```
vyorkin: the name `gas_limit` is confusing. we're setting it to `0`... its either `gas_available` or `gas_left`, smth like that

apps/evm/lib/evm/gas.ex
ayrat555: @masonforest actually there is already variable in `Functions` module. Thanks!

apps/evm/lib/evm/gas.ex
ayrat555: done

PR #172, opened 2018-06-18
Add `generate_state_tests` mix task for generating general state tests

masonforest:
Run `mix generate_state_tests` to generate `GeneralState` tests and `mix generate_state_tests --count` to print the count of passing vs failing tests.


apps/blockchain/scripts/generate_state_tests.exs
```diff
@@ -0,0 +1,2571 @@
+defmodule GenerateStateTests do
+  alias MerklePatriciaTree.Trie
+  alias Blockchain.{Account, Transaction}
+  alias Blockchain.Interface.AccountInterface
+
+  use EthCommonTest.Harness
+  # use ExUnit.Case, async: true
```
germsvel: Should we remove this comment if it's not used?

apps/blockchain/scripts/generate_state_tests.exs
```diff
@@ -0,0 +1,2571 @@
...
+
+  @ethereum_common_tests_path "../../ethereum_common_tests"
```
germsvel: I think this could be replaced with `EthCommonTest.Helpers.ethereum_common_tests_path` (we'd have to make it a public function). That way we only have that relative url in a single place. What do you think?

apps/blockchain/scripts/generate_state_tests.exs
```diff
@@ -0,0 +1,2571 @@
...
+      "CallRecursiveBomb0_OOG_atMaxCallDepth",
+      "CallRecursiveBomb1",
+      "CallRecursiveBomb2",
+      "CallRecursiveBomb3",
+      "CallRecursiveBombLog",
+      "CallRecursiveBombLog2",
+      "CallToNameRegistrator0",
+      "CallToNameRegistratorAddressTooBigLeft",
+      "CallToNameRegistratorAddressTooBigRight",
+
```
germsvel: What do you think about extracting these `IO.puts` comments into functions that reveal the intention for them? 
For example, this one seems like it's closing a test group, so it might be better to call, 

```elixir
if !only_count, do: log_closing_group()

# later

defp log_closing_group do 
  IO.puts (" ],")
end
```

It might also help clean up some of the other logging. For example for test names, 

```elixir
def log_test(test_name) do 
  IO.puts ("     \"#{test_name}\",")
end
```

Not sure if it's overkill for the script, but I think it could help.

germsvel: What's the significance of `{2,1}` and `{"passing", 0}` here?

germsvel: Not sure if this is doing the same thing, but in case it is,  `EthCommonTest.Helpers.test_files` does something that looks similar.

germsvel: There's something similar to this in `EthCommonTest.Helpers.read_test_file`. It doesn't take the same number of arguments though, so not sure.

mix.exs
```diff
@@ -21,7 +21,7 @@ defmodule Exthereum.MixProject do
         excluded_paths: [
           Path.join(@root_path, "_build/test/lib/abi/ebin"),
           Path.join(@root_path, "_build/test/lib/exth_crypto/ebin"),
-          Path.join(@root_path, "_build/test/lib/ex_wire/ebin")
+          Path.join(@root_path, "_build/test/lib/ex_wire/ebin"),
```
germsvel: Is this change needed?

apps/blockchain/scripts/generate_state_tests.exs
```diff
@@ -0,0 +1,2571 @@
+defmodule GenerateStateTests do
+  alias MerklePatriciaTree.Trie
+  alias Blockchain.{Account, Transaction}
+  alias Blockchain.Interface.AccountInterface
+
+  use EthCommonTest.Harness
+  # use ExUnit.Case, async: true
+
+  @ethereum_common_tests_path "../../ethereum_common_tests"
+  @passing_tests_by_group %{
```
germsvel: Is the intention for us to keep this list updated when we (and by that I mean you 😄since I'm not working on this) fix tests? If so, would it be helpful to separate this list into a file and read it in here?

apps/blockchain/scripts/generate_state_tests.exs
```diff
@@ -0,0 +1,2571 @@
+defmodule GenerateStateTests do
```
ayrat555: @masonforest  I think it's better to use `.ex` extension. Maybe we can skip this file from running dialyzer

masonforest: Hmmm, I tried adding it to `excluded_paths` in `mix.exs` but it still failed on CI :/

Does anyone know how to make dialyzer ignore a single file, cc: @cernivec ?

apps/blockchain/scripts/generate_state_tests.exs
masonforest: It's a [odd API](http://erlang.org/doc/man/ets.html#update_counter-4).

`{2,1}` is an `UpdateOp` which increments a counter by 1.

`{"passing", 0}` initializes the "passing" key to `0`

mix.exs
masonforest: nope! Reverted 👍

apps/blockchain/scripts/generate_state_tests.exs
masonforest: Yep, good call.

Ideally we just read the test directories and build this list dynamically. I made [an issue](https://github.com/poanetwork/mana/issues/175) to cover this in a follow-on PR.

apps/blockchain/scripts/generate_state_tests.exs
masonforest: This should be covered in https://github.com/poanetwork/mana/issues/175

masonforest: Yeah, this is specific to `state_tests` :/ this should also be covered in https://github.com/poanetwork/mana/issues/175

apps/blockchain/scripts/generate_state_tests.exs
ayrat555: @masonforest we can add this file to ignore

apps/blockchain/scripts/generate_state_tests.exs
germsvel: Thanks for the explanation!

PR #184, opened 2018-06-25
fix failed call execution

ayrat555:
From the Yellow Paper:
```if the execution halts in an exceptional fashion  (i.e.  due to an exhausted gas supply, stack underflow, invalid jump destination or invalid instruction), then no gas is refunded to the caller and the state is reverted to the point immediately prior to balance transfer```

### CODE REVIEW

File: apps/blockchain/lib/blockchain/contract/message_call.ex [hunk 1]
```diff
@@ -81,7 +83,18 @@ defmodule Blockchain.Contract.MessageCall do
...
 
     {gas, sub_state, exec_env, output} = fun.(params.available_gas, exec_env)
-    {exec_env.account_interface.state, gas, sub_state, output}
+
+    # From the Yellow Paper:
+    # if the execution halts in an exceptional fashion
+    # (i.e.  due to an exhausted gas supply, stack underflow, in-
+    # valid jump destination or invalid instruction), then no gas
+    # is refunded to the caller and the state is reverted to the
+    # point immediately prior to balance transfe
```
masonforest: typo: transfer*

PR #187, opened 2018-06-25
increment words count for call operation

ayrat555:
Resolves https://github.com/poanetwork/mana/issues/185

The problem was caused by the wrong active words count after call operation

### DIFF
--- a/apps/evm/lib/evm/operation/system.ex
+++ b/apps/evm/lib/evm/operation/system.ex
@@ -229,7 +229,7 @@ defmodule EVM.Operation.System do
   def return([_mem_start, mem_end], %{machine_state: machine_state}) do
     # We may have to bump up number of active words
 
-    words = EVM.Memory.get_active_words(mem_end)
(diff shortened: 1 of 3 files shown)


apps/evm/lib/evm/operation/system.ex
```diff
@@ -84,15 +84,18 @@ defmodule EVM.Operation.System do
...
         machine_state: machine_state
-      }) do
+           }) do
     to = Address.new(to)
     {data, machine_state} = EVM.Memory.read(machine_state, in_offset, in_size)
 
     call_gas = if value != 0, do: call_gas + Gas.callstipend(), else: call_gas
 
+    words = Memory.get_active_words(out_offset + out_size)
+    updated_machine_state = MachineState.maybe_set_active_words(machine_state, words)
```
masonforest: Should we push this logic into `EVM.MessageCall` or is this specific to `call` and doesn't apply to `delegate_call` and `static_call`?

Issue #166, opened 2018-06-14
Handle DEVp2p handshake and other messages


germsvel:
Currently, our ExWire tcp connection is handling the encrypted handshake. 

We now need to be able to handle a DEVp2p handshake (Hello messages) as well as Disconnect, Ping, and Pong messages. All of these messages should only be send/received if the crypto handshake was successful. 

For more on DEVp2p protocol, see https://github.com/ethereum/wiki/wiki/%C3%90%CE%9EVp2p-Wire-Protocol

Note that all messages must be framed. For more on framing see https://github.com/ethereum/devp2p/blob/master/rlpx.md#framing.

hayesgm:
Was this closed by #410?

InoMurko:
Can we close this @hayesgm ?

germsvel:
Yeah, I think #410 may have handled this. I'm not sure how robust it is, but it seems like this can be closed.

PR #206, opened 2018-07-02
Handle session when devp2p handshake recieved

cernivec:
## Receive DevP2p Handshake

When Dev P2P handshake (`%ExWire.Packet.Hello`) is received from a peer, we check if it matches our Dev P2P handshake. In other words, the capabilities of our node are compared with the peer's.
If they are matched the session is set to active.

## After DevP2p Handshake verification
When we receive further framed data we always check if the session is set to active before we continue with unframing.

## Incompatible peer 
If the capabilities do not match then we disconnect the session, generate Dev P2P Disconnect (`%ExWire.Packet.Disconnect`) and send it to the peer and at the end shutdown the connection.

Part of the issue: https://github.com/poanetwork/mana/issues/166


germsvel:
I'll close this in favor of https://github.com/poanetwork/mana/pull/252

PR #218, opened 2018-07-03
Return a padded address from `BuiltIn.ecrecover` instead of a public key

masonforest:
* Add `Address.new_from_public_key`


apps/blockchain/test/blockchain/state_test.exs
```diff
@@ -833,24 +833,29 @@ defmodule Blockchain.StateTest do
       # "CALLCODESha256_3_prefix0",
       "CALLCODESha256_4",
       "CALLCODESha256_4_gas99",
+<<<<<<< HEAD
```
ayrat555: merge artifacts :)

PR #220, opened 2018-07-04
Don't write to memory when invalid input is given

masonforest:
When ecrecover is called with an invalid signature it should push a `1` (success) to the stack. It shouldn't, however, write to memory or set `gas` to `0`.

Other changes:

* Refactored `MessageCall.update_state`
* Added `MachineState.add_refund`


apps/evm/lib/evm/message_call.ex
```diff
@@ -83,25 +83,47 @@ defmodule EVM.MessageCall do
...
-        sub_state: n_sub_state
-      }
-    else
-      failed_call(message_call)
+  def update_state({gas_remaining, n_sub_state, n_exec_env, output}, message_call) do
+    case output do
+      :failed ->
+        failed_call(message_call)
+
+      :invalid_input ->
```
ayrat555: logic for `:invalid_input` and `_`  are almost exactly the same. Maybe we should move it to separate method with memory writing condition parameter?

masonforest: Yup, good call: [(d58c5) Combine two match branches for simplicity](https://github.com/poanetwork/mana/pull/220/commits/d58c50a6e91a7c3e42250265653c76cdbb226955)

PR #222, opened 2018-07-04
remove key from storage on sstore

ayrat555:
Changes:
- Clear account's storage on sstore operation if the value is zero

fixes https://github.com/poanetwork/mana/issues/221


apps/blockchain/lib/blockchain/account.ex
```diff
@@ -539,6 +539,15 @@ defmodule Blockchain.Account do
     end)
   end
 
+  @spec put_storage(EVM.state(), EVM.address(), integer(), integer()) :: EVM.state()
```
masonforest: was this meant to be for `remove_storage`?

PR #228, opened 2018-07-05
Add a script for syncing mainnet through Infura

masonforest:
Other changes:

* Add ommer rewards
* Add foundation (mainnet) chain config
* Allow gas limits that are equal to the minimum


ayrat555:
nice work! so how many blocks are synced?

PR #233, opened 2018-07-09
Blockchain tests

ayrat555:
Resolves https://github.com/poanetwork/mana/issues/231


apps/blockchain/lib/blockchain/chain.ex
```diff
@@ -181,14 +190,20 @@ defmodule Blockchain.Chain do
...
+    accounts =
+      Enum.reduce(json_accounts, [], fn json_account = {_address, info}, acc ->
+        account =
+          if is_nil(info["builtin"]) do
+            get_account(json_account)
+          else
+            get_builtin_account(json_account)
+          end
+
+        acc ++ [account]
```
germsvel: Would it be better to prepend to the list instead of adding it to the end, `[account | acc]`? Not sure how important it is, but it should be slightly more performant. The only downside is that it would invert the order of elements in the list since it's prepending them.

apps/blockchain/lib/blockchain/chain.ex
```diff
@@ -207,6 +222,31 @@ defmodule Blockchain.Chain do
...
+    address = load_raw_hex(raw_address)
+
+    balance = if info["balance"], do: load_decimal(info["balance"]), else: 0
+
+    nonce =
+      if info["nonce"],
+        do: load_hex(info["nonce"]),
+        else: 0
+
+    builtin_account = %{
```
germsvel: Would it be an improvement to turn this `builtin_account` into a struct and have a function that generates it from json (essentially this function you have here)?  What do you think?

apps/blockchain/test/blockchain_test.exs
```diff
@@ -1,4 +1,1486 @@
...
+    # "./GeneralStateTests/stTransactionTest/Opcodes_TransactionInit_d22g0v0.json",
+    # "./GeneralStateTests/stTransactionTest/Opcodes_TransactionInit_d23g0v0.json",
+    # "./GeneralStateTests/stTransactionTest/Opcodes_TransactionInit_d24g0v0.json",
+    # "./GeneralStateTests/stTransactionTest/Opcodes_TransactionInit_d25g0v0.json",
+    # "./GeneralStateTests/stTransactionTest/Opcodes_TransactionInit_d26g0v0.json",
+    # "./GeneralStateTests/stTransactionTest/Opcodes_TransactionInit_d27g0v0.json",
+    # "./GeneralStateTests/stTransactionTest/Opcodes_TransactionInit_d28g0v0.json",
+    # "./GeneralStateTests/stTransactionTest/Opcodes_TransactionInit_d29g0v0.json",
+    # "./GeneralStateTests/stTransactionTest/Opcodes_TransactionInit_d2g0v0.json",
+    # "./Gene
```
germsvel: I think you have an extra space in this test description.

germsvel: Could some of the functions in `EthCommonTest.Helpers` be helpful here?

apps/blockchain/test/blockchain_test.exs
```diff
@@ -1,4 +1,1486 @@
...
+  alias Blockchain.{Blocktree, Account, Transaction}
+  alias MerklePatriciaTree.Trie
+  alias Blockchain.Account.Storage
+  alias Block.Header
+
   doctest Blockchain
+
+  @ethereum_common_tests_path "/../../ethereum_common_tests/BlockchainTests/"
+  @tests [
+    "./GeneralStateTests/stAttackTest/ContractCreationSpam_d0g0v0.json",
```
germsvel: How are we deciding which tests to include here vs doing something like we've done in `blockchain/transaction_test.exs`?

ayrat555: We're including all tests. Do you suggest specifying only directories? I tried to make it similar to `state_tests.exs`

germsvel: Oh, I see. Yeah `blockchain/state_tests.exs` is also a good example. I don't know if I have a suggestion. Mine was just a question to try to understand which tests we're putting in which files (since I haven't been working on tests, I'm not sure which approach you're taking). 

For example, us this file (`test/blochain_test.exs`) supposed to include _only_ `GeneralStateTests`? Or will all the tests that we are currently testing in places like `state_test.exs` eventually be moved here? If this doesn't make sense, I can also ask on slack or during standup. I'm not sure I'm writing the most understandable question.

ayrat555: The idea was to include all tests from `BlockchainTests' directory from ethereum common tests. Maybe there is a confusion because blockchain tests also have general state tests group which tests adding blocks. 

From Ethereum tests' site:
```
Of special notice is the /BlockchainTests/GeneralStateTests folder within the blockchain tests folder structure, which contains a copy of the General State Tests but executes them within the logic of the blockchain tests.
```

ayrat555:
@masonforest  no. but good idea. I will add it in this pr

apps/blockchain/test/blockchain_test.exs
germsvel: Oh yeah, that was confusing me. That clarifies things a bit. 

So eventually, we would remove `blockchain/state_test.exs` and `blockchain/transaction_test.exs` because those tests would be here, right?

ayrat555: im)
`trantransaction_test.exs` tests files from the `TransactionTests` directory
`state_test.exs` tests files from the `GeneralStateTests` directory
`blockchain_test.exs` tests files from the `BlockchainTests` directory

We can move all json test to separate files or separate directory.

germsvel: Ohh, I see. Sorry. I saw `Blockchain/TransitionTests/` and read "TransactionTests" instead. Yeah, what you're saying makes a lot more sense now. Thanks for explaning!

apps/blockchain/test/blockchain_test.exs
ayrat555: I think these functions will be useful when we will fix all blockchain tests. Currently, it's very helpful to see which tests are failing

ayrat555: @germsvel I automated generating passing tests. It won't be an issue from now

PR #235, opened 2018-07-09
Standardize ExWire.Handshake functions

germsvel:
ExWire.Handshake has been changing over time as we continuously refactor while working on the networking layer. This commit updates the functions to more consistently take an `ExWire.Handshake` struct as a first argument. It also creates a more even flow whether the node is initiating the handshake or is the recipient of the handshake. This mirroring of the code paths makes it easier to reason about and debug.

Why these changes?
===================

In an effort to make the `DEVp2p` session handling easier to reason about (and hence add more messages), I am trying to simplify the TCP module first.

Eventually, we should have `ExWire.Adapter.TCP` be an "adapter", as the name suggests. Currently, it contains all the logic that coordinates the encrypted handshake (and the DEVp2p session initiation). Ideally, it should not contain all that logic, and it should simply focus on the tcp connection, looking more like the `ExWire.Adapter.UDP` module does.

### DIFF
--- a/apps/ex_wire/lib/ex_wire/adapter/tcp/server.ex
+++ b/apps/ex_wire/lib/ex_wire/adapter/tcp/server.ex
@@ -104,7 +104,8 @@ defmodule ExWire.Adapter.TCP.Server do
     Logger.debug("[Network] Generating EIP8 Handshake for #{peer.host}")
 
     handshake =
-      Handshake.new(peer.remote_id)
+      peer.remote_id
+      |> Handshake.new()
       |> Handshake.generate_auth()
 
     Logger.debug("[Network] [#{peer}] Sending Handshake to #{peer.host}")
--- a/apps/ex_wire/lib/ex_wire/adapter/tcp/server.ex
+++ b/apps/ex_wire/lib/ex_wire/adapter/tcp/server.ex
@@ -32,7 +32,7 @@ defmodule ExWire.Adapter.TCP.Server do
   @doc """
   Sends auth message to peer
   """
-  def send_auth_message() do
+  def send_auth_message do
     GenServer.cast(self(), :send_auth_message)
   end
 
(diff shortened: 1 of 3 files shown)


apps/ex_wire/test/ex_wire/handshake_test.exs
```diff
@@ -36,321 +53,296 @@ defmodule HandshakeTest do
...
-      |> Handshake.Struct.AckRespV4.serialize()
-      |> Handshake.EIP8.wrap_eip_8(my_static_public_key, her_ephemeral_key_pair)
-
-    {:ok, my_decoded_ack_resp, _ack_bin, <<>>} =
-      Handshake.read_ack_resp(encoded_ack_msg, my_static_private_key)
-
-    assert my_decoded_ack_resp == her_ack_resp
-  end
+  describe "EIP-8 test vectors" do
+    test "handshake auth plain" do
```
germsvel: The changes that follow in this test are all just indentation. I put them under the `describe "EIP-8 test vectors"` to more clearly outline test groups.

apps/ex_wire/lib/ex_wire/adapter/tcp/server.ex
```diff
@@ -104,15 +104,13 @@ defmodule ExWire.Adapter.TCP.Server do
     Logger.debug("[Network] Generating EIP8 Handshake for #{peer.host}")
 
     handshake =
-      %Handshake{remote_pub: peer.remote_id}
-      |> Handshake.initiate()
-
-    new_state = Map.merge(state, %{handshake: handshake})
+      Handshake.new(peer.remote_id)
```
ayrat555: we can start pipeline operator with raw value

germsvel: Updated in 47eb9a8d5784216ef1b79409de580bfe1429e261

PR #237, opened 2018-07-10
Closes #236: Use rocksdb when syncing with Infura

### LINKED ISSUE #236: Store results of `sync_with_infura` in rocksdb
Currently every time we run `sync_with_infura` we have to sync the chain from scratch because it's storing block data in Erlang Term Storage. We should store the blockdata in rocksdb so we can continue syncing from when we left off. This will also be valuable once we hook up ex_wire to sync blocks.

masonforest:
This will allow us to save our state state so we won't need to re-sync
every time we run the script.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 erl_crash.dump
 *.ez
 *.beam
+apps/*/config/dev.exs
 
 /db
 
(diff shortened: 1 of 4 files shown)


jabson009:
It was cool and nice

Issue #153, opened 2018-06-08
Substate in call operations 


ayrat555:
Currently we're rewriting substate after message-call operations (call, callcode, delegatecall, static call)

But we should merge them

![img](https://i.imgur.com/hNf3Y3j.png)

vyorkin:
Same thing regarding `CREATE` op, should we create a separate issue to track this?

PR #240, opened 2018-07-11
Merge nested sub states

ayrat555:
Nested operations (message calls and contract creations) did not save execution sub states

Resolves https://github.com/poanetwork/mana/issues/153
Fixes https://github.com/poanetwork/mana/issues/239

### DIFF
--- a/apps/evm/lib/evm/message_call.ex
+++ b/apps/evm/lib/evm/message_call.ex
@@ -1,5 +1,5 @@
 defmodule EVM.MessageCall do
-  alias EVM.{ExecEnv, Memory, Builtin, VM, Functions, Stack, MachineState}
+  alias EVM.{ExecEnv, Memory, Builtin, VM, Functions, Stack, MachineState, SubState}
   alias EVM.Interface.AccountInterface
 
   @moduledoc """
@@ -8,6 +8,7 @@ defmodule EVM.MessageCall do
   defstruct [
     :current_exec_env,
     :current_machine_state,
+    :current_sub_state,
     :output_params,
     # s
     :sender,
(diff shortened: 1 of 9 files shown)


apps/evm/lib/evm/message_call.ex
```diff
@@ -105,10 +106,19 @@ defmodule EVM.MessageCall do
         message_call.current_exec_env
         |> Map.put(:account_interface, n_exec_env.account_interface)
 
+      refund = n_sub_state.refund + message_call.current_sub_state.refund
+
+      sub_state = %SubState{
+        refund: n_sub_state.refund + message_call.current_sub_state.refund,
+        selfdestruct_list:
+          n_sub_state.selfdestruct_list ++ message_call.current_sub_state.selfdestruct_list,
+        logs: message_call.current_sub_state.logs
```
masonforest: We don't have to append logs from the message call like we did with the selfdestruct_list?

apps/evm/lib/evm/message_call.ex
```diff
@@ -105,10 +106,19 @@ defmodule EVM.MessageCall do
         message_call.current_exec_env
         |> Map.put(:account_interface, n_exec_env.account_interface)
 
+      refund = n_sub_state.refund + message_call.current_sub_state.refund
+
+      sub_state = %SubState{
```
masonforest: Could you add `SubState.merge` or something to avoid duplication in `System.ex`. It's minor so might not be worth it

apps/evm/lib/evm/message_call.ex
```diff
@@ -105,10 +106,19 @@ defmodule EVM.MessageCall do
...
+        refund: n_sub_state.refund + message_call.current_sub_state.refund,
```
germsvel: Could we use the `refund` that is calculated in line `109` here? I think it's the same calculation, right?

PR #243, opened 2018-07-11
Include `init` when initializing transactions in tests

masonforest:
`init` is basically an alias to `data` in contract creating transactions


masonforest:
This passes some tests but I'm going to look into why it fails others now...

masonforest:
Fixed: [Only set `init` if `transaction[:to]` is empty](https://github.com/poanetwork/mana/pull/243/commits/1447c82dfee67722a4cf241648cdacf7187ef156)

Ready for review 🙂

PR #254, opened 2018-07-16
fix invalid blockchain rlp tests

ayrat555:
- Not all tests contain Genesis RLP.
- We shouldn't fail on invalid block rlp 
- We shouldn't fail on invalid hex encoding
- RLP may be valid but the list with data is invalid

fixes https://github.com/poanetwork/mana/issues/253


apps/blockchain/lib/blockchain/block.ex
```diff
@@ -119,6 +119,36 @@ defmodule Blockchain.Block do
     }
   end
 
+  @spec decode_rlp(binary()) :: {:ok, [ExRLP.t()]} | {:error, any()}
+  def decode_rlp("0x" <> hex_data) do
+    try do
+      hex_binary = Base.decode16!(hex_data, case: :mixed)
```
germsvel: I might not be following all the logic through, but is there a benefit to using `Base.decode16!` along with a `try rescue` instead of `Base.decode16` (without the `!`) and just do a `case` on the return value?

apps/blockchain/lib/blockchain/block.ex
```diff
@@ -119,6 +119,36 @@ defmodule Blockchain.Block do
...
+
+      decode_rlp(hex_binary)
+    rescue
+      e ->
```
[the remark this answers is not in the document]
ayrat555: @masonforest yes, good idea. But the problem is not only in RLP library. We should also check if hex encoding is valid and decoded from RLP block list has valid structure

apps/blockchain/lib/blockchain/block.ex
ayrat555: @germsvel good point. I will change it in my next PR with blockchain test fixes.

PR #264, opened 2018-07-18
Don't double count `data` and `init` when calculating intrinsic gas

germsvel:
We need to calculate the intrinsic gas cost of the input data based on whether we are performing a contract creation transaction or a message call transaction.

We find that we perform the same check for whether or not a transaction is a contract creation transaction or a message call transaction in several places in `Blockchain.Transaction` so we extract some helper functions to deal with that and to silo that logic into a single place.

Having both `init` and `data` as keys in the `Transaction` struct is somewhat confusing. It may be better to simply rid ourselves of `init` and to always put contract creation code or message call input data the  `data` attribute. After all, the `RLP` decoding uses the same position for both of them, our code already handles contract creation and message call transactions correctly, and other implementation seem to simply not define `init` (at least Parity doesn't).

## Other changes:

* Fix transaction setup in blockchain/state_test.exs

The setup for a transaction was being done incorrectly. When the transaction is a contract creation transaction, we should populate the `init` field with the data. Thus far we were always populating the `data` field with the data (which is only the case for message call transactions).

* Update expected gas and expected accounts

We update the expected gas and accounts in `test/blockchain/transaction_test.exs` and `test/blockchain/block_test.exs` tests that had hardcoded values. The gas was hardcoded to `21_756`. We update the value to `28_180` because we think that is the correct value. Both of those transactions are contract creation transactions which until now were not accounting for the correct amount of gas because when the `Transaction.execute` function created a contract, it passed the `data` field into `init_data`. But for contract creation, the transaction should pass the data in the `init` field.


ayrat555:
@germsvel looking good! This branch has a couple of conflicts with master:
1. Conflict in `.dialyzer.ignore-warnings`
2. ethereum_common_tests is newer in master

PR #265, opened 2018-07-18
Add bin/setup and update README

germsvel:
Add a `bin/setup` script that installs dependencies (except for Elixir itself) and copies configuration files.

Also update README to mention `bin/setup` script and to describe syncing with Infura and testing.


ayrat555:
@masonforest `ethereum common tests` in master is newer

masonforest:
Ah, good catch! I'll update them

PR #271, opened 2018-07-20
check readability with credo

ayrat555:


masonforest:
Nice! I'm always in for quick wins that improve readability 👍

PR #278, opened 2018-07-23
Set correct directory for tests in README

germsvel:
The state and transaction tests (for ethereum common tests) currently live under the blockchain app and in the blockchain directory. We update the README to correctly show that

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,8 +65,8 @@ cd apps/evm && mix test test/evm_test.exs
 cd apps/blockchain && mix test test/blockchain_test.exs
 
 # Ethereum General State tests
-cd apps/blockchain && mix test test/state_test.exs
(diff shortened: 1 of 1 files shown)


germsvel:
@masonforest dialyzer check here failed with that strange error we've seen before. I only changed the README here, so I wouldn't have expected anything to fail. I wonder if it happens intermittently.

masonforest:
A couple ideas:
1. Can we restore the mix cache [here](https://github.com/poanetwork/mana/blob/master/.circleci/config.yml#L106) so the mix dependencies can be used from the build step?

2. If not maybe we need to set [resource_class: large](https://github.com/poanetwork/mana/blob/master/.circleci/config.yml#L106)

PR #280, opened 2018-07-24
fix randomStatetest21BC

ayrat555:
fixes https://github.com/poanetwork/mana/issues/279

Changes:
- We should calculate memory cost based on code size
- We shouldn't copy anything to the memory if code is empty and offset is zero


apps/evm/lib/evm/operation/environmental_information.ex
```diff
@@ -287,14 +287,14 @@ defmodule EVM.Operation.EnvironmentalInformation do
...
-      0
-    else
-      wrapped_address = Helpers.wrap_address(address)
+    wrapped_address = Helpers.wrap_address(address)
 
-      account_code =
-        AccountInterface.get_account_code(exec_env.account_interface, wrapped_address)
+    account_code = AccountInterface.get_account_code(exec_env.account_interface, wrapped_address)
 
+    if length == 0 || length + mem_offset > EVM.max_int() ||
```
germsvel: This is a pretty complex `if` check. What do you think about moving it to its own line or to a private function? It may add clarity to what is going on there. What are we checking for there?

ayrat555: Here we're checking memory boundaries: If we overflowed memory (`length + mem_offset > EVM.max_int()`) or if we shouldn't copy anything (`length == 0`)

ayrat555: Unfortunately, it can not be moved to its own line because we're formatter :(

ayrat555:
@germsvel I added explanation

PR #282, opened 2018-07-25
[WIP] increase words count on logging operations

ayrat555:
fixes https://github.com/poanetwork/mana/issues/281


apps/evm/lib/evm/operation/logging.ex
```diff
@@ -243,6 +234,14 @@ defmodule EVM.Operation.Logging do
     address = exec_env.address
     updated_substate = sub_state |> SubState.add_log(address, topics, data)
 
-    %{sub_state: updated_substate}
+    updated_machine_state =
+      if size > 0 do
```
germsvel: I'm not super familiar with active words. What does `size > 0` mean in terms of that?

PR #285, opened 2018-07-25
Don't include accounts without balances in the genesis block

masonforest:
Other changes:

* Rescue errors without an :exit atom in `sync_with_infura`
* Include `dev.secret.exs` in `apps/blockchain/config/dev.exs`


apps/blockchain/lib/blockchain/chain.ex
```diff
@@ -225,7 +225,7 @@ defmodule Blockchain.Chain do
   defp get_builtin_account({raw_address, info}) do
     address = load_raw_hex(raw_address)
 
-    balance = if info["balance"], do: load_decimal(info["balance"]), else: 0
+    balance = if info["balance"], do: load_decimal(info["balance"]), else: nil
```
ayrat555: I think `else` is unnecessary

PR #290, opened 2018-07-26
set mix_hash on validation stage

ayrat555:
fixes https://github.com/poanetwork/mana/issues/289


apps/evm/lib/evm/gas.ex
```diff
@@ -32,8 +32,6 @@ defmodule EVM.Gas do
   @g_sset 20_000
   # Paid for an SSTORE operation when the storage value’s zeroness remains unchanged or is set to zero.
   @g_sreset 5000
-  # Amount of gas to pay for a SELFDESTRUCT operation.
-  @g_selfdestruct 5000
```
germsvel: Why are we deleting `@g_selfdestruct` and `g_txcreate`? Are they defined elsewhere?

apps/blockchain/lib/blockchain/block.ex
```diff
@@ -447,12 +447,14 @@ defmodule Blockchain.Block do
     beneficiary = opts[:beneficiary] || nil
     extra_data = opts[:extra_data] || <<>>
     state_root = opts[:state_root] || parent_block.header.state_root
+    mix_hash = opts[:mix_hash] || parent_block.header.mix_hash
```
germsvel: Interesting catch. I've been trying to track a mismatching mix_hash for another test. I wonder if this will help!

apps/evm/lib/evm/gas.ex
ayrat555: @germsvel yes, @g_selfdestruct is defined in `Refunds` module, @g_txcreate is defined is `CreateContract` module

PR #292, opened 2018-07-26
fix signed operations helper

ayrat555:
fixes https://github.com/poanetwork/mana/issues/291

### DIFF
--- a/apps/blockchain/test/blockchain/state_test.exs
+++ b/apps/blockchain/test/blockchain/state_test.exs
@@ -1927,7 +1927,7 @@ defmodule Blockchain.StateTest do
     # "stSystemOperationsTest/suicideCoinbase.json",
     "stSystemOperationsTest/suicideNotExistingAccount.json",
     "stSystemOperationsTest/suicideOrigin.json",
-    # "stSystemOperationsTest/suicideSendEtherPostDeath.json",
+    "stSystemOperationsTest/suicideSendEtherPostDeath.json",
     "stSystemOperationsTest/suicideSendEtherToMe.json",
     "stSystemOperationsTest/testRandomTest.json",
     "stTransactionTest/ContractStoreClearsOOG.json",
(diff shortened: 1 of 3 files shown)


apps/blockchain/test/blockchain/state_test.exs
```diff
@@ -1927,7 +1927,7 @@ defmodule Blockchain.StateTest do
     # "stSystemOperationsTest/suicideCoinbase.json",
     "stSystemOperationsTest/suicideNotExistingAccount.json",
     "stSystemOperationsTest/suicideOrigin.json",
-    # "stSystemOperationsTest/suicideSendEtherPostDeath.json",
+    "stSystemOperationsTest/suicideSendEtherPostDeath.json",
```
germsvel: That's strange. Shouldn't this have already been uncommented? Did we forget to do that @masonforest?

apps/evm/lib/helpers.ex
```diff
@@ -7,6 +7,9 @@ defmodule EVM.Helpers do
   use Bitwise
   alias EVM.Address
 
+  @max_uint_255 round(:math.pow(2, 255)) - 1
+  @max_256_ceiling round(:math.pow(2, 256))
```
germsvel: Just out of curiosity, does the 256 limit come from the yellow paper?

apps/evm/lib/helpers.ex
```diff
@@ -97,17 +100,15 @@ defmodule EVM.Helpers do
...
+    end
   end
 
   def decode_signed(n) when is_binary(n) do
-    <<sign::size(1), _::bitstring>> = n
-
-    if sign == 0 do
-      :binary.decode_unsigned(n)
-    else
-      :binary.decode_unsigned(n) - EVM.max_int()
```
germsvel: Do we use `EVM.max_int()` anywhere else or can that function now be removed?

ayrat555: yes, it's used

apps/evm/lib/helpers.ex
ayrat555: yep, 32 bytes

apps/blockchain/test/blockchain/state_test.exs
ayrat555: yes, it looks like so

PR #296, opened 2018-07-27
fix credo warnings

ayrat555:
resolves https://github.com/poanetwork/mana/issues/30


apps/blockchain/scripts/sync_with_infura.ex
```diff
@@ -95,6 +95,7 @@ defmodule SyncWithInfura do
     Blockchain.Block.add_ommers(block, ommers)
   rescue
     error ->
+      # credo:disable-for-next-line
```
germsvel: Should we use `Logger` instead of `IO` for these kinds of things? I'm guessing that's what credo is complaining about?

apps/evm/lib/evm/debugger.ex
```diff
@@ -362,6 +362,7 @@ defmodule EVM.Debugger do
          exec_env,
          input_sequence
        ) do
+    # credo:disable-for-next-line
     IO.inspect([machine_state, sub_state, exec_env], limit: :infinity)
```
germsvel: Should we use `Logger` instead of  `IO` for these things?

PR #297, opened 2018-07-27
Return hash of block in BLOCKHASH

germsvel:
Closes https://github.com/poanetwork/mana/issues/256

What?
====

Our `BLOCKHASH` operation was incorrect on several points.

1. It returned the `mix_hash` of the block instead of the hash of the block.

2. We were also requesting the incorrect block internally. The function `BlockInterface.get_ancestor_header/2` gets the nth ancestor from the block. So we need to pass the difference between the current block and the one that is requested in the `BLOCKHASH` operation. We were simply passing the requested block number before.

3. When generating a block via `Block.gen_child_block/3`, we were not setting the parent hash.

Other changes:
==============

* Some doc tests started failing due to the changes that generate a block hash
for a parent block. Those failures revealed that we were not setting the
`mix_hash` and `nonce` correctly when creating the genesis block.

Test changes:
=============

The fixes for BLOCKHASH operation passed several tests. But they also seem to have created regressions in two tests. The following are now failing:

* `bcRandomBlockhashTest/randomStatetest463BC.json` and
* `bcRandomBlockhashTest/randomStatetest56BC.json`


apps/blockchain/lib/blockchain/block.ex
```diff
@@ -460,6 +463,16 @@ defmodule Blockchain.Block do
     }
   end
 
+  @doc """
+  Sets block's parent's hash
+  """
+  @spec set_block_parent_hash(t, t) :: t
+  def set_block_parent_hash(block, parent_block) do
+    hash = parent_block.block_hash || hash(parent_block)
```
masonforest: `hash` -> `parent_hash` to match the name of `parent_block`?

PR #299, opened 2018-07-30
add blocks by timestamp

ayrat555:
Some blocks have the same total difficulty. So we must choose the first block sorted by timestamp as a canonical block.

Fixes https://github.com/poanetwork/mana/issues/298


apps/blockchain/scripts/generate_blockchain_tests.ex
```diff
@@ -33,9 +33,14 @@ defmodule GenerateBlockchainTests do
             end)
 
           cond do
-            failing != 0 -> log_commented_test(relative_path)
-            passing != 0 -> log_test(relative_path)
-            true -> :ok
+            failing != 0 ->
```
germsvel: Just curious about the reasoning for this formatting change. Do we need it?6

ayrat555: @germsvel it looks like new formatter version issue. I returned formatting by hand.

PR #301, opened 2018-07-30
Replace :math.floor with Float.floor

gballet:
I am getting issues compiling Mana; it seems that the Erlang `:math` module isn't available on my system (Fedora 2.6, elixir 1.6.6, Erlang 8.3.5.4).

This PR replaces that call with the native elixir `Float.floor` call.


ayrat555:
@gballet Thanks

PR #306, opened 2018-07-31
Extract P2P logic from TCP.Server

germsvel:
What?
=====

The fact that the peer to peer server uses tcp is an implementation detail, and it should not be the primary driver of the implementation. For this reason, we extract the business logic out of `TCP.Server` and into a new `P2P` module.

We then combine `ExWire.Adapter.TCP` (the client side of the GenServer) and `ExWire.Adapter.TCP.Server` `into ExWire.P2P.Server`, a GenServer that makes use of the `P2P` module.

Ideally, we could make `P2P` completely ignorant of the tcp connection. For now, we are passing the `socket` into the p2p module when it needs to send messages out. In the future, it may be better to have the `P2P` module return tuples that have instructions on what must be done, e.g.

```elixir
{:send, ack_message} = P2P.handle_message(auth)
```

In this way, the `P2P.Server` would be entirely responsible for the tcp connection and keeping state, and the `P2P` module is just a module with functions.


apps/ex_wire/lib/ex_wire/adapter/tcp.ex
```diff
@@ -1,78 +0,0 @@
-defmodule ExWire.Adapter.TCP do
```
germsvel: This module was for the most part moved to `ExWire.P2P.Server`

apps/ex_wire/lib/ex_wire/adapter/tcp/server.ex
```diff
@@ -1,282 +0,0 @@
-defmodule ExWire.Adapter.TCP.Server do
```
germsvel: Most of the logic in each of the functions of this module now lives in `P2P`. I essentially extracted the body of all the functions from this module into `P2P`.

Issue #190, opened 2018-06-26
EVM hardfork configuration


ayrat555:
EVM should know that specific commands, features etc are not implemented in current instance of EVM. So for example FRONTIER version of EVM should fail on delegatecall operation because it was introduced in HOMESTEAD.

PR #308, opened 2018-08-01
evm configuration

ayrat555:
Resolves https://github.com/poanetwork/mana/issues/190
Resolves https://github.com/poanetwork/mana/issues/309


apps/blockchain/lib/blockchain/configuration/homestead_test.ex
```diff
@@ -0,0 +1,15 @@
+defmodule Blockchain.Configuration.HomesteadTest do
+  defstruct contract_creation_cost: 53_000, has_static_call: true
```
germsvel: Does homestead have the static call opcode (I thought it was introduced later) or does this refer to something else?

apps/blockchain/lib/blockchain/configuration/homestead_test.ex
```diff
@@ -0,0 +1,15 @@
+defmodule Blockchain.Configuration.HomesteadTest do
```
germsvel: Why does the Homestead configuration live here instead of the EVM? What do you think about moving it within the evm app?

apps/blockchain/lib/blockchain/transaction.ex
```diff
@@ -330,11 +331,11 @@ defmodule Blockchain.Transaction do
...
   """
-  @spec intrinsic_gas_cost(t, Header.t()) :: Gas.t()
-  def intrinsic_gas_cost(tx, block_header) do
+  @spec intrinsic_gas_cost(t, EVM.Configuration.t()) :: Gas.t()
+  def intrinsic_gas_cost(tx, config) do
     data_cost = input_data_cost(tx)
-    cc_cost = contract_creation_cost(tx, block_header)
-    data_cost + cc_cost + Gas.g_transaction()
+
+    data_cost + contract_creation_cost(tx, config)
```
germsvel: I see that you have combined the contract creation cost and the cost for the transaction. Should we rename the function so it no longer reads `contract_creation_cost`? As it stands it's a little misleading because this could return the transaction cost and not the contract creation cost.

apps/blockchain/lib/blockchain/transaction.ex
```diff
@@ -330,11 +331,11 @@ defmodule Blockchain.Transaction do
       iex> Blockchain.Transaction.intrinsic_gas_cost(%Blockchain.Transaction{to: <<>>, init: <<1, 2, 0, 3>>, data: <<>>}, %Block.Header{number: 5_000_000})
       3 * 68 + 4  + 21000
   """
-  @spec intrinsic_gas_cost(t, Header.t()) :: Gas.t()
-  def intrinsic_gas_cost(tx, block_header) do
+  @spec intrinsic_gas_cost(t, EVM.Configuration.t()) :: Gas.t()
+  def intrinsic_gas_cost(tx, config) do
     data_cost = input_data_cost(tx)
-    cc_cost = contract_creation_cost(tx, block_header)
-    data_cost + cc_cost + Gas.g_transaction()
```
germsvel: This used to add the contract creation cost and the transaction cost. I think in practice one of those will always be `0` right? That's why you switched it from adding to an `if`/`else` statement within the function `contract_creation_cost/2`?

apps/blockchain/lib/blockchain/configuration/homestead_test.ex
ayrat555: Yes, my bad. `delegatecall` was introduced in Homestead

apps/blockchain/lib/blockchain/configuration/homestead_test.ex
ayrat555: I'm totally ok with it. But some parameters will be used only in the blockchain app. For example, contract creation cost

apps/blockchain/lib/blockchain/transaction.ex
ayrat555: yes, good point

apps/blockchain/lib/blockchain/configuration/homestead_test.ex
germsvel: Yeah, it just seems strange to me that the different implementations may live in different apps. If some parameters are only used in blockchain and some are only used in the evm, does it make sense to have something else (I dont' know what) where we keep the fork configurations?

masonforest:
Looking good!

At some point we'll need to branch on `block_number` correct? Since we have access to `block_number` [here](https://github.com/poanetwork/mana/pull/308/files#diff-2a96e8e1e18bb42a650736d45fd8f73bR169)  through block_header could we just pass block_number through to `EVM.Configuration` and have a function that takes a block number and returns the associated configuration?

```EVM.Configuration.for_block_number(block_header.number)```

File: apps/blockchain/lib/blockchain/transaction.ex (hunk 4 above)
Comment (ayrat555) [reply apps/blockchain/lib/blockchain/transaction.ex]: yes, it's related to [EIP 2](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2.md). In frontier cost for simple transaction and contract creation transactions were 21_000. Starting from Homestead contract creation transaction cost increased to 53_000

**Comment by ayrat555** [comment]
@masonforest yes, my idea is for production we can have a separate configuration based on block_numbers. Also, in tests, it is assumed that hardfork started from the beginning so it's not possible  to use block numbers in test configuration

**Comment by masonforest** [comment]
Got it. Makes sense.

LookingGTM

File: apps/evm/lib/evm/configuration/frontier_test.ex [hunk 5]
```diff
@@ -0,0 +1,15 @@
...
+    %__MODULE__{}
+  end
+end
+
+defimpl EVM.Configuration, for: EVM.Configuration.FrontierTest do
+  @spec contract_creation_cost(EVM.Configuration.t()) :: integer()
+  def contract_creation_cost(config), do: config.contract_creation_cost
+
+  @spec has_delegate_call?(EVM.Configuration.t()) :: boolean()
+  def has_delegate_call?(config), do: config.has_delegate_call
```
germsvel: No need to change this if you don't want to, but we could just change this to, 

```elixir
def has_delegate_call?(_config), do: false
```

There's no real reason why it needs to live in a `struct`, right?

apps/evm/lib/evm/configuration/frontier_test.ex
```diff
@@ -0,0 +1,15 @@
+defmodule EVM.Configuration.FrontierTest do
```
germsvel: Just curious, why is this called `FrontierTest` instead of `Frontier`?

ayrat555: @germsvel I think these configurations will be used only in tests. Btw I can rename them

apps/evm/lib/evm/configuration/frontier_test.ex
ayrat555: @germsvel yes. good idea

apps/evm/lib/evm/configuration/frontier_test.ex
germsvel: If they're only going to be used in tests, should we put them somewhere within `test/support` or something like that? Also, what do you think of renaming them `TestFrontier` instead of `FrontierTest` since ex unit uses the ending-with-Test convention?

Issue #310, opened 2018-08-02
Fix StateTests/stSystemOperationsTest/suicideCoinbase.json


germsvel:
This is one of the last 2% of common StateTests failing.

PR #311, opened 2018-08-02
Reset account on SELFDESTRUCT if it is beneficiary

germsvel:
Closes https://github.com/poanetwork/mana/issues/310

What?
=====

It turns out that Frontier and Homestead need to reset an account that has been marked for destruction (instead of deleting it). They only do this, however, if the account is the beneficiary account (ie. the coinbase account).

Resetting the account means that we set the balance to 0 and remove the code from it.

A commit in Ethereum common tests led us to understand this. I include it here for future reference: https://github.com/ethereum/tests/commit/013aa117669dfe82b4ea289a7191fa6658da1334


apps/blockchain/lib/blockchain/transaction.ex
```diff
@@ -190,8 +190,23 @@ defmodule Blockchain.Transaction do
...
+    final_state =
+      updated_state
+      |> pay_and_refund_gas(sender, tx, refund, block_header)
+      |> clean_up_accounts_marked_for_destruction(sub_state, block_header)
+
+    # { σ', Υ^g, Υ^l }, as defined in Eq.(79) and Eq.(80)
+    {final_state, expended_gas, sub_state.logs}
+  end
+
+  defp apply_transaction(state, tx, block_header, sender, config) do
```
ayrat555: typespec would be nice

apps/blockchain/lib/blockchain/account.ex
```diff
@@ -601,4 +601,21 @@ defmodule Blockchain.Account do
...
+  ## Examples
+
+      iex> state = MerklePatriciaTree.Trie.new(MerklePatriciaTree.Test.random_ets_db())
+      ...>   |> Blockchain.Account.put_account(<<0x01::160>>, %Blockchain.Account{balance: 100, code_hash: "abc"})
+      iex> state
+      ...> |> Blockchain.Account.reset_account(<<0x01::160>>)
+      ...> |> Blockchain.Account.get_account(<<0x01::160>>)
+      %Blockchain.Account{}
+  """
+  @spec clear_balance(EVM.state(), EVM.address()) :: EVM.state()
```
ayrat555: wrong typespec

germsvel: Is it? I think it is right. Which part is wrong?

apps/blockchain/lib/blockchain/transaction.ex
germsvel: Added some in https://github.com/poanetwork/mana/pull/311/commits/08f862ab9eae240e29ff45fe1f41542c572fb984

apps/blockchain/lib/blockchain/account.ex
ayrat555: the method is called `reset_account`. the typespec is for `clear_balance`

germsvel: hahahaa. I was looking at the specs themselves and didn't look at the name. Thanks!

germsvel: Updated in https://github.com/poanetwork/mana/pull/311/commits/08f862ab9eae240e29ff45fe1f41542c572fb984

PR #313, opened 2018-08-02
Enable setup for various Linux platforms

gballet:
`bin/setup` now works on Linux. Tested with Fedora 26 and Debian 9.

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -7,6 +7,71 @@ fancy_echo() {
   printf "\\n$fmt\\n" "$@"
 }
 
+install_linux_deps() {
+  local LINUX_DEPS="automake autoconf autogen libtool gcc curl"
+  local flavor=`grep "^ID=" /etc/os-release | cut -d"=" -f 2`
+  echo $flavor
+  case $flavor in
+    debian|ubuntu)
+      sudo apt-get install -y $LINUX_DEPS libgmp-dev
+      ;;
+    fedora|centos)
+      sudo yum install $LINUX_DEPS gmp-devel
+      ;;
+    *)
+      fancy_echo "Unrecognized distribution $flavor"
+      exit 1
+      ;;
+  esac
+}
+
+install_linux_rust() {
+  curl https://sh.rustup.rs -sSf | sh
+  if [ $? -ne 0 ]; then
+    fancy_echo "Rust installation failed"
+    exit 1
+  fi
+}
+
+install_rust() {
+  local sys=`uname -s`
+  case $sys in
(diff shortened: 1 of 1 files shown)


ayrat555:
@gballet good work. I'm also using Linux but I did not try the setup script. I think we can run it on circle ci to check if everything's ok. What do you think @masonforest @germsvel

masonforest:
> @gballet good work. I'm also using Linux but I did not try the setup script. I think we can run it on circle ci to check if everything's ok. What do you think @masonforest @germsvel

Yeah, that would ensure our `./bin/setup` would stay up to date as well 👍

bin/setup
```diff
@@ -7,6 +7,71 @@ fancy_echo() {
...
+
+install_deps() {
+  local sys=`uname -s`
+  echo $sys
+  case $sys in
+    Linux*)
+      install_linux_deps
+      ;;
+    Darwin*)
+      fancy_echo "Installing dependencies via brew"
```
masonforest: could we pull this into a `install_darwin_deps` function?

bin/setup
```diff
@@ -7,6 +7,71 @@ fancy_echo() {
...
+
+install_linux_rust() {
+  curl https://sh.rustup.rs -sSf | sh
+  if [ $? -ne 0 ]; then
+    fancy_echo "Rust installation failed"
+    exit 1
+  fi
+}
+
+install_rust() {
```
masonforest: `rustup` worked on my mac too. We can probably just use that everywhere and drop the `case` statement.

bin/setup
```diff
@@ -7,6 +7,71 @@ fancy_echo() {
   printf "\\n$fmt\\n" "$@"
 }
 
+install_linux_deps() {
+  local LINUX_DEPS="automake autoconf autogen libtool gcc curl"
+  local flavor=`grep "^ID=" /etc/os-release | cut -d"=" -f 2`
+  echo $flavor
+  case $flavor in
+    debian|ubuntu)
+      sudo apt-get install -y $LINUX_DEPS libgmp-dev
```
germsvel: Will `apt-get` install the package even if it's already installed in the system? If so, it would be nice to check for the dependencies before we try to install them. That way the `bin/setup` script can be re-run multiple times without negative side effects. 

The same question applies to `yum` for fedora and centos.

ayrat555:
I'll merge it to add to circle ci config @masonforest @germsvel

ayrat555:
https://github.com/poanetwork/mana/pull/371

PR #314, opened 2018-08-03
Fix order of arguments in extcodecopy opcode

masonforest:
Other changes:

* Refactored state tests


apps/blockchain/test/blockchain/state_test.exs
```diff
@@ -10,2365 +10,105 @@ defmodule Blockchain.StateTest do
...
-    "stRandom2/randomStatetest422.json",
-    "stRandom2/randomStatetest423.json",
-    "stRandom2/randomStatetest424.json",
-    "stRandom2/randomStatetest425.json",
-    "stRandom2/randomStatetest426.json",
-    "stRandom2/randomStatetest428.json",
-    "stRandom2/randomStatetest429.json",
-    "stRandom2/randomStatetest430.json",
-    "stRandom2/randomStatetest433.json",
-    "stRandom2/r
```
ayrat555: @masonforest wow :clap: . That's exactly what we need! Good work
 Can we use test name prefixed with its test group to easier find test? Also, can we add to `generate_state_tests` script  hardfork flag so it can print failing tests by hardfork

ayrat555:
I think `general state tests` are newer is master

apps/blockchain/test/blockchain/state_test.exs
```diff
@@ -10,2365 +10,103 @@ defmodule Blockchain.StateTest do
...
```
ayrat555: duplicates. I think it's related to the fact that some tests have multiple transactions

masonforest: Whoops! I merged already. I'll make a new PR...

PR #320, opened 2018-08-06
Save block_tree state in `sync_with_infura` script

masonforest:
Before we'd have to rebuild the blocktree every time we ran
`sync_with_infura` this saves the blocktree to the state database so we
don't have to do that anymore.

### DIFF
--- a/apps/blockchain/scripts/sync_with_infura.ex
+++ b/apps/blockchain/scripts/sync_with_infura.ex
@@ -13,10 +13,16 @@ defmodule SyncWithInfura do
 
   def add_block_to_tree(db, chain, tree, n) do
     next_block = get_block(n)
-    {:ok, next_tree} = Blockchain.Blocktree.verify_and_add_block(tree, chain, next_block, db)
-    Logger.info("Verified Block #{n}")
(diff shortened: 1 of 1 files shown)


apps/blockchain/scripts/sync_with_infura.ex
```diff
@@ -136,10 +142,10 @@ current_block =
 parents = SyncWithInfura.get_parents(current_block, db, [])
 
 tree =
-  Enum.reduce(parents, empty_tree, fn block, tree ->
-    {:ok, tree} = Blockchain.Blocktree.verify_and_add_block(tree, chain, block, db)
-    Logger.info("Verified Saved Block #{block.header.number}")
-    tree
-  end)
+  case MerklePatriciaTree.DB.get(db, "current_block_tree") do
```
ayrat555: What if there is a value of "current_block_tree" in the DB but you want to start from the first block?

masonforest: I've just been removing the `db` directory. Ideally we replace this script with a CLI tool that's pulling blocks from ex_wire soon.

PR #323, opened 2018-08-06
Fix memory cost for EXTCODECOPY

germsvel:
Closes https://github.com/poanetwork/mana/issues/322

What?
=====

Uses memory offset for memory cost calculation for `EXTCODECOPY` (instead of code offset).


apps/evm/lib/evm/gas.ex
```diff
@@ -145,11 +145,11 @@ defmodule EVM.Gas do
     memory_expansion_cost(machine_state, memory_offset, length)
   end
 
-  def memory_cost(:extcodecopy, [_address, _memory_offset, code_offset, length], machine_state) do
-    if code_offset + length > EVM.max_int() do
+  def memory_cost(:extcodecopy, [_address, mem_offset, _code_offset, length], machine_state) do
+    if mem_offset + length >= EVM.max_int() do
```
masonforest: After thinking about it I think this should be `=` instead of `>=` (the `max_int` itself should be allowed). Do all the same tests  pass with `=` instead?

germsvel: Oh yeah, I think what's there is wrong. But it should be `>` not `>=` or `=`, right? In other words, the memory cost should be equal to the memory expansion cost if it exceeds the max_int. Otherwise it should be zero.

masonforest: oh yeah, `>` 👍

germsvel: Updated in https://github.com/poanetwork/mana/pull/323/commits/a1e49c48656034df209f9aa546ff76ede3bc0b8c

ayrat555:
@germsvel let's merge it!

PR #324, opened 2018-08-06
abi README updates

andogro:
There are a few things to resolve but thought I would submit a pr for comments & review

- There is another abi repo at https://github.com/poanetwork/ex_abi - do we want it to live here or within the mana repo, or both?
- I'd like to move to the GNU license we use on other projects, are there any issues with this?
- I was not able to run code examples in elixir, I may just need clearer instructions to setup and run. `(UndefinedFunctionError) function ABI.encode/2 is undefined (module ABI is not available)`
- There is a JSON file in priv (dog.abi.json) meant to be used as an example and referred to in code examples. If we want to continue to use, it would be good to use as the example in the README. I removed from the description for now as it was confusing in the current context.

### DIFF
--- a/apps/abi/README.md
+++ b/apps/abi/README.md
@@ -1,12 +1,12 @@
 # ABI
 
-The [Application Binary Interface](https://solidity.readthedocs.io/en/develop/abi-spec.html) (ABI) of Solidity describes how to transform binary data to types which the Solidity programming language understands. For instance, if we want to call a function `bark(uint32,bool)` on a Solidity-created contract `contract Dog`, what `data` parameter do we pass into our Ethereum transaction? This project allows us to encode such function calls.
+The [Application Binary Interface](https://solidity.readthedocs.io/en/develop/abi-spec.html) (ABI) application encodes and decodes  data for use in Solidity-created smart contracts. Data is encoded according to its [type](#Support), and contains the parameters and contents necessary to pass into an Ethereum transaction.
 
 ## Usage
 
 ### Encoding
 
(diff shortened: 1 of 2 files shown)


masonforest:
> There is another abi repo at https://github.com/poanetwork/ex_abi - do we want it to live here or within the mana repo, or both?

Good call: [PR to remove it](https://github.com/poanetwork/mana/pull/327).

> I'd like to move to the GNU license we use on other projects, are there any issues with this?

Nope that should be fine.

> I was not able to run code examples in elixir, I may just need clearer instructions to setup and run. (UndefinedFunctionError) function ABI.encode/2 is undefined (module ABI is not available)

Hmmm.. How did you run it?

This worked for me:

````
tmp % git clone git@github.com:poanetwork/ex_abi.git
...
tmp % cd ex_abi
ex_abi master % mix deps.get
...
ex_abi master % iex -S mix
...
Generated ex_abi app
Interactive Elixir (1.6.6) - press Ctrl+C to exit (type h() ENTER for help)
iex(1)> ABI.encode("baz(uint,address)", [50, <<1::160>> |> :binary.decode_unsigned])
<<162, 145, 173, 214, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
  0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 50, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
  0, 0, ...>>
iex(2)>
````

PR #326, opened 2018-08-07
EIP2: out of gas error on contract creation if not enough gas

ayrat555:
fixes https://github.com/poanetwork/mana/issues/325

From [EIP2](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2.md):
```
If contract creation does not have enough gas to pay for the final gas fee for adding the contract code to the state, the contract creation fails (i.e. goes out-of-gas) rather than leaving an empty contract.
```
merge after https://github.com/poanetwork/mana/pull/323

### DIFF
--- a/apps/blockchain/lib/blockchain/contract/create_contract.ex
+++ b/apps/blockchain/lib/blockchain/contract/create_contract.ex
@@ -135,15 +135,10 @@ defmodule Blockchain.Contract.CreateContract do
         end
 
       resultant_state =
-        cond do
-          state_after_init == nil ->
-            params.state
-
-          insufficient_gas ->
-            state_after_init
-
-          true ->
-            Account.put_code(state_after_init, address, output)
+        if insufficient_gas do
+          state_after_init
+        else
+          Account.put_code(state_after_init, address, output)
         end
 
       {:ok, {resultant_state, resultant_gas, accrued_sub_state}}
--- a/apps/blockchain/lib/blockchain/contract/create_contract.ex
+++ b/apps/blockchain/lib/blockchain/contract/create_contract.ex
@@ -71,7 +71,7 @@ defmodule Blockchain.Contract.CreateContract do
       if output == :failed do
         {:error, {params.state, 0, SubState.empty()}}
       else
-        {:ok, finalize(result, params, contract_address)}
+        finalize(result, params, contract_address)
       end
     end
   end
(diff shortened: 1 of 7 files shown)


apps/blockchain/test/blockchain_test.exs
```diff
@@ -17,13 +17,7 @@ defmodule BlockchainTest do
...
     ],
     "Homestead" => [
-      "GeneralStateTests/stDelegatecallTestHomestead/delegatecodeDynamicCode_d0g0v0.json",
-      "GeneralStateTests/stMemoryTest/extcodecopy_dejavu_d0g0v0.json",
-      "GeneralStateTests/stRandom2/randomStatetest571_d0g0v0.json",
-      "GeneralStateTests/stRandom2/randomStatetest643_d0g0v0.json",
-      "bcRandomBlockhashTest/randomStatetest21BC.json",
-      "bcValidBlockTest/dataTx2.json",
-      "bcWalletTest/wallet2outOf3txs.json"
+      "GeneralStateTests/stDelegatecallTestHomestead/delegatecodeDynamicCode_d0g0v0.json"
```
masonforest: Down to one failing test! :D

apps/evm/lib/evm/configuration.ex
```diff
@@ -1,13 +1,19 @@
...
 
   @type t :: module()
 
+  # EIP2
   @spec contract_creation_cost(t) :: integer()
   def contract_creation_cost(t)
 
+  # EIP2
+  @spec fail_contract_creation?(t) :: boolean()
+  def fail_contract_creation?(t)
```
germsvel: What do you think about naming this something more specific? `fail_contract_creation?` seems like a thing that could happen for many reasons. What do you think about something like `fail_contract_creation_lack_of_gas?` or something like that? I know at that point it's getting really verbose, but I think in this case it might be better to be verbose over it being confusing. What do you think?

apps/evm/lib/evm/configuration.ex
```diff
@@ -1,13 +1,19 @@
 defprotocol EVM.Configuration do
   @moduledoc """
-  Prototcol for defining hardfork configurations.
+  Protocol for defining hardfork configurations.
```
germsvel: 👍

apps/blockchain/lib/blockchain/contract/create_contract.ex
```diff
@@ -122,30 +122,32 @@ defmodule Blockchain.Contract.CreateContract do
...
+    else
+      resultant_gas =
+        cond do
+          state_after_init == nil -> 0
+          insufficient_gas -> remaining_gas
+          true -> remaining_gas - contract_creation_cost
+        end
+
+      resultant_state =
+        cond do
```
germsvel: This `cond` and the one above are kind of confusing to me. Or at least, I can't quite figure out why we are checking those conditions. I know you didn't really change this and it's mostly indentation, but do you know why we check these conditions? And is it something we could encapsulate in a well-named function?

apps/evm/lib/evm/configuration.ex
ayrat555: yes, good idea

PR #331, opened 2018-08-07
Do not transfer value in DELEGATECALL

germsvel:
Fixes https://github.com/poanetwork/mana/issues/329

What?
=====

For opcode `DELEGATECALL` we should not attempt to transfer the value (funds) from the sender's account to the recipient since the sender and value are persisted. For a helpful description of the all the message call types, see https://www.badykov.com/ethereum/2018/06/17/message-calls-in-ethereum/

Other changes:

- We also add a `EVM.TestFactory` to help simplify the setup needed for tests.

### DIFF
--- a/apps/evm/lib/evm/message_call.ex
+++ b/apps/evm/lib/evm/message_call.ex
@@ -10,31 +10,59 @@ defmodule EVM.MessageCall do
     :current_machine_state,
     :current_sub_state,
     :output_params,
-    # s
     :sender,
-    # o
     :originator,
-    # r
     :recipient,
-    # c
     :code_owner,
-    # p
     :gas_price,
-    # v
     :value,
-    # v with overline
     :execution_value,
-    # d
     :data,
-    # e
     :stack_depth
   ]
+
+  @type out_size :: integer()
+  @type out_offset :: integer()
+  @type output_params :: {out_offset(), out_size()}
+
+  @typedoc """
+  Terms from the Yellow Paper:
+
+  s: sender,
+  o: originator,
+  r: recipient,
+  c: code_owner,
+  p: gas_price,
+  v: value,
+  v with overline: execution_value,
+  d: data,
+  e: stack_depth
+  """
+  @type t :: %__MODULE__{
+          current_exec_env: ExecEnv.t(),
+          current_machine_state: MachineState.t(),
+          current_sub_state: SubState.t(),
+          output_params: output_params(),
+          sender: EVM.Address.t(),
+          originator: EVM.Address.t(),
+          recipient: EVM.Address.t(),
+          code_owner: EVM.Address.t(),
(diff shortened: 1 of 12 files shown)


apps/evm/lib/evm/message_call.ex
```diff
@@ -64,21 +90,31 @@ defmodule EVM.MessageCall do
...
-
     message_call
+    |> transfer_funds_if_needed()
     |> prepare_call_execution_env()
     |> execute_call(message_call)
     |> update_state(message_call)
   end
 
+  defp transfer_funds_if_needed(message_call) do
+    if should_transfer_funds?(message_call) do
```
germsvel: This is where we are now checking whether or not we should transfer the funds. The `message_call` structs define a boolean field that informs us as to whether or not we should transfer the value. It's not the greatest implementation (I don't like switching on a boolean), so I'm totally open to other ideas.

apps/evm/lib/evm/message_call.ex
```diff
@@ -10,26 +10,52 @@ defmodule EVM.MessageCall do
...
-    # v
     :value,
-    # v with overline
     :execution_value,
-    # d
     :data,
-    # e
-    :stack_depth
+    :stack_depth,
+    :transfer_value
```
ayrat555: maybe we can pass zero for `delegatecall` instead of adding extra field?

germsvel: That's an interesting idea. So you're thinking we could just set `value` to `0` for `delegatecall`? Wouldn't that conflict with the notion that the `value` is persisted from the previous call (what lives in `exec_env.value_in_wei`)?

ayrat555: Actually, you're right. I forgot that it's passed to child's exec env

apps/evm/lib/evm/message_call.ex
```diff
@@ -64,21 +88,32 @@ defmodule EVM.MessageCall do
...
-
     message_call
+    |> transfer_funds_if_needed()
     |> prepare_call_execution_env()
     |> execute_call(message_call)
     |> update_state(message_call)
   end
 
+  defp transfer_funds_if_needed(message_call) do
+    if sender_is_recipient?(message_call) do
```
masonforest: if [value is `0`](https://github.com/poanetwork/mana/pull/331/files#diff-ca8e96a2417500c71d9313d86cffcbb3R140) don't you no longer need the conditional? `deletegatecall` will just transfer 0 which should do nothing right?

germsvel: Yeah, I was adding this for completeness since it has that condition in the Yellow Paper. But I think I'm going to split that into a separate issue if I add it at all. I'm not sure if this is breaking the build.

germsvel:
@ayrat555 Yeah, good idea. Will do

PR #335, opened 2018-08-08
blockchain app content

andogro:
Includes typo fixes, README modifications and content additions for some of the blockchain modules. I will continue with this process but wanted to update with incremental progress.

- Need to add step by step instructions for main README debugging, current instructions do not produce a tangible result.
-Blockchain.Bloom was missing documentation - I added some related information from yellow paper but unsure if it covers module processes
- Also added documentation for blockchain.ex as a brief summary of other mods. May make more sense to add   @moduledoc File.read!("README.md") for these top level files?


apps/blockchain/lib/blockchain/application.ex
```diff
@@ -1,5 +1,5 @@
 defmodule Blockchain.Application do
-  # See http://elixir-lang.org/docs/stable/elixir/Application.html
+  # See https://elixir-lang.org/getting-started/mix-otp/supervisor-and-application.html
```
germsvel: I think we can remove the comments from this module. They are auto-generated when created the `mix` project.

apps/blockchain/lib/blockchain/application.ex
```diff
@@ -32,7 +32,7 @@ defmodule Blockchain.Application do
       # worker(Blockchain.Worker, [arg1, arg2, arg3]),
     ]
 
-    # See http://elixir-lang.org/docs/stable/elixir/Supervisor.html
+    # See https://elixir-lang.org/getting-started/mix-otp/supervisor-and-application.html
```
germsvel: Same with this one. I think we should remove them since they're auto-generated.

apps/blockchain/lib/blockchain/bloom.ex
```diff
@@ -1,4 +1,9 @@
 defmodule Blockchain.Bloom do
+  @moduledoc """
+  When a block is generated or verified, the contract addresses and fields from the generated logs are added to a bloom filter. This is included in the block header.
+
+  _From Yellow Paper 4.3.1. Transaction Receipt_: The transaction receipt is a tuple of four items comprising the post-transaction state, R, the cumulative gas used in the block containing the transaction receipt as of immediately after the transaction has happened, Ru, the set of logs created through execution of the transaction, Rl and the Bloom filter composed from information in those logs, Rb:
```
germsvel: What do you think about breaking this paragraph (and the previous one) into multiple lines? It makes it a bit hard to read like this.

germsvel: I also noticed that this ends with `Rb:`. Is there supposed to be something after that or is the colon supposed to be a period?

apps/blockchain/lib/blockchain/bloom.ex
```diff
@@ -1,4 +1,17 @@
 defmodule Blockchain.Bloom do
+  @moduledoc """
```
andogro: @ayrat555 I see you wrote a post on the bloom filter - adding any of those comments or examples to this file may be useful, or at least a link to the article. http://www.badykov.com/ethereum/2017/10/29/ethereum-bloom-filter/

ayrat555: @andogro good idea

ayrat555:
@andogro @germsvel  @masonforest  currently, credo check for required module documentations is turned off. should we turn on it?

ayrat555:
@andogro we should also delete old licenses from `mix.exs` files 
https://github.com/poanetwork/mana/blob/master/apps/blockchain/mix.exs#L15

PR #336, opened 2018-08-10
one 64th of the parent gas for CREATE

ayrat555:
From eip150:
```CREATE only provides all but one 64th of the parent gas to the child call.```

### CODE REVIEW

File: apps/evm/lib/evm/operation/system.ex [hunk 1]
```diff
@@ -53,7 +72,7 @@ defmodule EVM.Operation.System do
           exec_env.config
         )
       else
-        {:error, {exec_env.account_interface, machine_state.gas, SubState.empty()}}
+        {:error, {exec_env.account_interface, available_gas, SubState.empty()}}
```
germsvel: Do you know if this should return the original `machine_state.gas` or if it is okay for it to return the modified version (in the case that `available_gas = EVM.Helpers.all_but_one_64th(machine_state.gas)`)?

ayrat555: @germsvel in both cases (failure and success) remaining gas from execution added to 1/64 gas that left in machine state. so in case of failure => 63/64 + 1/64

PR #346, opened 2018-08-14
eip160 change byte cost for exp op

ayrat555:
Changes:
- add EIP158 config
- implement EIP160 - change gas cost for exp op
- run EIP158 state tests
- run EIP158 blockchain tests


apps/evm/lib/evm/configuration/eip158.ex
```diff
@@ -0,0 +1,49 @@
...
+  def new do
+    %__MODULE__{}
+  end
+end
+
+defimpl EVM.Configuration, for: EVM.Configuration.EIP158 do
+  alias EVM.Configuration
+
+  @spec contract_creation_cost(Configuration.t()) :: integer()
+  def contract_creation_cost(config),
```
germsvel: I think we might be able to use `defdelegate` for a lot of these functions that are simply calling the `fallback_config`. Don't think it changes things much in practice, but just a thought if you feel like it'd be nicer.

ayrat555: @germsvel  Can you give a simple example? It looks like `defdelegate` is used to delegate to a different module

germsvel: Hmmm, yeah seems you're right. I thought there was another way to use `defdelegate` but doesn't seem like it makes sense here.

Issue #342, opened 2018-08-13
Run the difficulty tests from the Ethereum common tests


masonforest:
We should run the difficulty tests located here: [ethereum/tests _/_ BasicTests/difficultyFrontier.json](https://github.com/ethereum/tests/blob/develop/BasicTests/difficultyFrontier.json) for all working forks.

It can probably just call [Block.Header.get_difficulty](https://github.com/poanetwork/mana/blob/master/apps/evm/lib/block/header.ex#L422) directly.

Issue #343, opened 2018-08-13
Homestead Progress


germsvel:
In order to fully implement all of the [Homestead hardfork changes](https://github.com/ethereum/homestead-guide/blob/master/source/introduction/the-homestead-release.rst), we need to do the following: 

- [EIP 2](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2.md)
  - [ ] cost for creating contracts via a transaction is increased from 21000 to 53000. Contract creation from a contract using the CREATE opcode is unaffected.
  - [ ] transaction signatures whose s-value is greater than secp256k1n/2 are now considered invalid
  - [ ] If contract creation does not have enough gas to pay for the final gas fee for adding the contract code to the state, the contract creation fails (ie. goes out-of-gas) rather than leaving an empty contract.
  - [ ] Change the difficulty adjustment algorithm
- [EIP 7](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7.md)
  - [ ] Add a new opcode, DELEGATECALL
- [EIP 8](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-8.md)
  - [ ] Forward Compatibility compliance with the Robustness Principle

PR #347, opened 2018-08-14
Add Ethereum common difficulty tests

germsvel:
Closes https://github.com/poanetwork/mana/issues/342

What changed?
============

We add Ethereum common tests that test the difficulty calculation described under the "Block Header Validity" section of the Yellow Paper.

In the common tests, under the `BasicTests` directory, there are several files that test difficulty. Some are generated manually while others are refilled automatically. We make use of the file `difficultyFrontier.json`. It is worth noting that those tests do not contain difficulty tests only for the Frontier chain. They also test other forks (like Homestead). I am not sure why they have `Frontier` in the name since other forks are included.

In order to limit which tests run, so that we can pass tests progressively as we implement various forks, we check the block number in the test to ascertain which fork we're testing. For now, we are only testing Frontier block headers.

Other changes
=============

By adding the tests, we discovered that we our difficulty calculations for Frontier (we're only testing Frontier so far) were wrong. The errors were all an off-by-one type of errors, so we make use of Elixir's built-in integer division `div/2` function instead of the `MathHelper.floor` function when calculating a block header's difficulty.

For an example of how the test failures look, see this failed build https://circleci.com/gh/poanetwork/mana/1816:

screen shot 2018-08-14 at 9 06 20 am
screen shot 2018-08-14 at 9 06 30 am

### DIFF
--- a/apps/evm/test/block/header_test.exs
+++ b/apps/evm/test/block/header_test.exs
@@ -1,7 +1,119 @@
 defmodule Block.HeaderTest do
   use ExUnit.Case, async: true
   doctest Block.Header
+
+  import ExthCrypto.Math, only: [hex_to_bin: 1, hex_to_int: 1]
+
   alias Block.Header
+  alias EVM.EthereumCommonTestsHelper, as: Helper
+
+  @forks_with_block %{
+    frontier: 1,
+    homestead: 1_150_000,
+    dao_fork: 1_920_000,
+    eip_150: 2_463_000,
+    spurious_dragon: 2_675_000,
+    byzantium: 4_370_000
+  }
+
+  @next_fork_block_number @forks_with_block[:homestead]
+
+  describe "Difficulty Tests (Ethereum Common Tests)" do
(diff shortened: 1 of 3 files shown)


apps/evm/test/block/header_test.exs
```diff
@@ -1,7 +1,119 @@
...
+  @forks_with_block %{
+    frontier: 1,
+    homestead: 1_150_000,
+    dao_fork: 1_920_000,
+    eip_150: 2_463_000,
+    spurious_dragon: 2_675_000,
+    byzantium: 4_370_000
+  }
+
+  @next_fork_block_number @forks_with_block[:homestead]
```
ayrat555: are we going to run `Homestead` difficulty tests in next pr?

germsvel: Yeah, I figured I'd break the work apart by adding the tests and passing Frontier, and then doing the Homestead work as part of https://github.com/poanetwork/mana/issues/343

PR #348, opened 2018-08-14
do not fail credo on todo comments

ayrat555:


germsvel:
Before we turn this off, I would ask the question: should we be failing with `TODO` comments? 

Maybe instead of `TODO` comments we should just create an `issue` for every instance of them. `TODO` comments can get stale and there's no high visibility of the fact that they need to be done other than when you run into one by them. 

That's what I have been doing in practice when I run into them. I turn them into issues or simply delete them if I feel like I'm addressing them already. But I hardly ever go looking for them when I try to figure out what needs to be done next.

ayrat555:
@germsvel I totally agree with you. We should use issues instead of todo comments

MattMSumner:
It looks like we decided to keep credo failing on Todo comments and start moving them over to issues on the project. @ayrat555 how would you feel about closing this PR?

PR #350, opened 2018-08-14
limit code size EIP170

ayrat555:
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-170.md

### DIFF
--- a/apps/blockchain/lib/blockchain/contract/create_contract.ex
+++ b/apps/blockchain/lib/blockchain/contract/create_contract.ex
@@ -131,8 +131,7 @@ defmodule Blockchain.Contract.CreateContract do
       insufficient_gas && EVM.Configuration.fail_contract_creation_lack_of_gas?(params.config) ->
         {:error, {params.state, 0, SubState.empty()}}
 
-      # EIP170 https://github.com/ethereum/EIPs/blob/master/EIPS/eip-170.md
-      byte_size(output) > 24_577 ->
+      EVM.Configuration.limit_contract_code_size?(params.config, byte_size(output)) ->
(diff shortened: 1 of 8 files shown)


apps/blockchain/lib/blockchain/contract/create_contract.ex
```diff
@@ -127,24 +127,30 @@ defmodule Blockchain.Contract.CreateContract do
...
-          Account.put_code(state_after_init, address, output)
-        end
-
-      {:ok, {resultant_state, resultant_gas, accrued_sub_state}}
+    cond do
+      insufficient_gas && EVM.Configuration.fail_contract_creation_lack_of_gas?(params.config) ->
+        {:error, {params.state, 0, SubState.empty()}}
+
+      # EIP170 https://github.com/ethereum/EIPs/blob/master/EIPS/eip-170.md
+      byte_size(output) > 24_577 ->
```
germsvel: What do you think about extracting this condition to a function that is more self-explanatory? 

```elixir
# in the `cond` block 
eip_170_byte_size?(output) -> 

# then further below
def eip_170_byte_size?(output), do: byte_size(output) > 24_577

# or even better, add docs
@doc """
This is implemented because of EIP170
See more at https://github.com/ethereum/EIPs/blob/master/EIPS/eip-170.md
"""
def eip_170_byte_size?(output), do: byte_size(output) > 24_577
```

What do you think?

ayrat555: yes, good idea. btw it looks like this change affects previous tests so I'll rewrite it using `EVM.Configuration`

PR #353, opened 2018-08-15
change start nonce for EIP161

ayrat555:
```
a. Account creation transactions and the CREATE operation SHALL, prior to the execution of the 
initialisation code, increment the nonce over and above its normal starting value by one (for normal 
networks, this will be simply 1, however test-nets with non-zero default starting nonces will be different).
```


apps/evm/lib/evm/configuration/homestead.ex
```diff
@@ -47,4 +47,7 @@ defimpl EVM.Configuration, for: EVM.Configuration.Homestead do
 
   @spec limit_contract_code_size?(Configuration.t(), integer()) :: boolean()
   def limit_contract_code_size?(config, _), do: config.fallback_config.limit_contract_code_size
+
+  @spec start_nonce(Configuration.t()) :: integer()
+  def start_nonce(config), do: config.fallback_config.start_nonce
```
germsvel: What do you think about doing `Configuration.start_nonce(config.fallback_config)`? That way, you let the `fallback_config` determine whether or not it responds to `start_nonce`. It may be that the `fallback_config` also has to call another fallback configuration (I know that's not the case in this particular example). Minor change but I think it adds a little bit of robustness.

ayrat555: good point

PR #364, opened 2018-08-20
static call EIP 214

ayrat555:
Changes:
- add `static` flag to execution environment
- add static call to evm configuration
- fix gas cost for static cost
- halt execution on static call state modification


apps/evm/lib/evm/functions.ex
```diff
@@ -175,4 +181,24 @@ defmodule EVM.Functions do
   end
 
   defp is_invalid_jump_destination?(_operation, _inputs, _machine_code), do: false
+
+  defp static_state_modification?(:call, [_, _, value, _, _, _, _]), do: value > 0
```
germsvel: Do you think it makes sense for these `static_state_modification?` functions to live in their respective operation modules? I'm not sure that it _does_ make sense since I'm not quite sure which module should have knowledge about what a `static_state_modification` is, but I'm commenting here just in case that makes sense to you.

apps/evm/lib/evm/gas.ex
```diff
@@ -382,14 +384,14 @@ defmodule EVM.Gas do
...
-        [gas_limit, to_address, value, _in_offset, _in_length, _out_offset, _out_length],
+        [gas_limit, to_address, _in_offset, _in_length, _out_offset, _out_length],
         _machine_state,
         exec_env
       ) do
     to_address = Address.new(to_address)
 
-    Configuration.call_cost(exec_env.config) + call_value_cost(value) +
-      new_account_cost(exec_env, to_address, value) + gas_limit
+    Configuration.call_cost(exec_env.config) + new_account_cost(exec_env, to_address, 0) +
```
germsvel: What is that `0` being passed? Could we extract that into a variable and give it a name? Otherwise it's a little bit of a mystery number.

apps/evm/lib/evm/message_call.ex
```diff
@@ -2,10 +2,7 @@ defmodule EVM.MessageCall do
   alias EVM.{ExecEnv, Memory, Builtin, VM, Functions, Stack, MachineState, SubState}
   alias EVM.Interface.AccountInterface
 
-  @moduledoc """
-  Describes a message call function that used for all call operations (call, delegatecall, callcode, staticcall).
-  """
-  defstruct [
+  @enforce_keys [
```
germsvel: 👍

apps/evm/lib/evm/functions.ex
ayrat555: It makes sense to put some kind of flag to operation group metadata module but I don't know how to store them flag based on input params for `call` operation

masonforest: Could we define it as a function that takes the input params? ie

    %EVM.Operation.Metadata{
       id: 0xFE,
       description: "Designated invalid instruction. It can be used to abort the execution.",
       sym: :invalid,
       input_count: 0,
       output_count: 0,
       group: :system,
       is_static?: fn (_, _, value, _, _, _, _) ->
        value > 0
       end
    },

Not sure how much metaprogramming and complexity it'll take to make this work. Might be worth a shot ¯\_(ツ)_/¯

ayrat555: it looks like anonymous functions are not supported as struct fields
```
** (ArgumentError) invalid value for struct field is_static, cannot escape #Function<0.102477801 in file:lib/evm/ope
ration/metadata.ex>. The supported values are: lists, tuples, maps, atoms, numbers, bitstrings, PIDs and remote func
tions in the format &Mod.fun/arity      
```

PR #366, opened 2018-08-21
add returndatacopy and returndatasize EIP 211

ayrat555:


apps/evm/lib/evm/functions.ex
```diff
@@ -137,6 +137,7 @@ defmodule EVM.Functions do
     end
   end
 
+  # credo:disable-for-next-line
```
germsvel: What was credo complaining about here?

ayrat555: `Function is too complex (CC is 13, max is 10).` EIPs introduce too many new operations

PR #367, opened 2018-08-21
Fix test_group from path retrieval in state tests

germsvel:
The `test_group` variable, which is used to determine known failures, was incorrectly returning the wrong part of the path (likely due to us changing from a relative path to an absolute path at some point). It was usually returning `mana`. We reverse the index to grab the elements from the end of the list so that we always get the group name, regardless of where the path starts.

### DIFF
--- a/apps/blockchain/test/blockchain/state_test.exs
+++ b/apps/blockchain/test/blockchain/state_test.exs
@@ -471,7 +471,7 @@ defmodule Blockchain.StateTest do
 
   test "Blockchain state tests" do
     Enum.each(test_directories(), fn directory_path ->
-      test_group = Enum.fetch!(String.split(directory_path, "/"), 4)
+      test_group = Enum.fetch!(String.split(directory_path, "/"), -1)
 
       directory_path
       |> tests()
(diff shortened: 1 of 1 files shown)


apps/blockchain/test/blockchain/state_test.exs
```diff
@@ -471,7 +471,7 @@ defmodule Blockchain.StateTest do
 
   test "Blockchain state tests" do
     Enum.each(test_directories(), fn directory_path ->
-      test_group = Enum.fetch!(String.split(directory_path, "/"), 4)
+      test_group = Enum.fetch!(String.split(directory_path, "/"), -1)
```
ayrat555: can we use pipe  operator?

germsvel: Yeah, we definitely can. I'll update this.

germsvel: Updated in https://github.com/poanetwork/mana/pull/367/commits/ab7ed76b0478dda896fb450f5b7449178ad167f1

apps/blockchain/test/blockchain/state_test.exs
```diff
@@ -471,7 +471,7 @@ defmodule Blockchain.StateTest do
 
   test "Blockchain state tests" do
     Enum.each(test_directories(), fn directory_path ->
-      test_group = Enum.fetch!(String.split(directory_path, "/"), 4)
+      test_group = test_group_from_directory(directory_path)
```
ayrat555: btw pipe operator can be used here
```
directory_path 
|> test_group_from_directory()
|> tests()

PR #369, opened 2018-08-21
Increment nonce of touched accounts on CREATE

### LINKED ISSUE #368: Fix "GeneralStateTests/stRevertTest/RevertPrefound_d0g0v0.json" blockchain test
This also has a corresponding "stRevertTest/RevertPrefound" general state test.

germsvel:
This closes #368.

What changed?
=============

We increment the nonce of an account when it is touched on CREATE (operation or
transaction call). We change the implementation of `start_nonce` introduced in
https://github.com/poanetwork/mana/pull/353, to `increment_nonce_on_create`
since I think that may be what is described in EIP 161 part a:

> a. Account creation transactions and the CREATE operation SHALL, prior to
the execution of the initialisation code, increment the nonce over and
above its normal starting value by one (for normal networks, this will be
simply 1, however test-nets with non-zero default starting nonces will be
different).

Note that the test-nests comment above would be configured at the chain level
not at the EVM-fork level. So I think it is safe to replace the `start_nonce`
implementation with this `increment_nonce_on_create`, though I am happy to
change the name of the configuration if something else seems more appropriate
and/or descriptive.

### DIFF
--- a/apps/blockchain/lib/blockchain/contract/create_contract.ex
+++ b/apps/blockchain/lib/blockchain/contract/create_contract.ex
@@ -7,28 +7,32 @@ defmodule Blockchain.Contract.CreateContract do
   alias Blockchain.Interface.{BlockInterface, AccountInterface}
   alias Block.Header
   alias Blockchain.Contract.Address
-  alias Blockchain.{Account, Contract}
+  alias Blockchain.Account
   alias EVM.{SubState, Gas}
 
-  # σ
   defstruct state: %{},
-            # s
             sender: <<>>,
-            # o
             originator: <<>>,
-            # g
             available_gas: 0,
-            # p
             gas_price: 0,
-            # v
             endowment: 0,
-            # i
             init_code: <<>>,
-            # e
             stack_depth: 0,
             block_header: nil,
             config: EVM.Configuration.Frontier.new()
 
+  @typedoc """
+  Yellow Paper Terms:
+
+  - σ: state,
+  - s: sender,
+  - o: originator,
+  - g: available_gas,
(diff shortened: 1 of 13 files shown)


apps/blockchain/lib/blockchain/contract/create_contract.ex
```diff
@@ -7,28 +7,32 @@ defmodule Blockchain.Contract.CreateContract do
...
             endowment: 0,
-            # i
             init_code: <<>>,
-            # e
             stack_depth: 0,
             block_header: nil,
             config: EVM.Configuration.Frontier.new()
 
+  @typedoc """
+  Yello Paper Terms:
```
masonforest: Yello -> Yellow

germsvel: Good catch. Updated in bd04e6984ecbc935d77f745c4a9561bfc88d9c48

apps/blockchain/lib/blockchain/contract/create_contract.ex
```diff
@@ -7,28 +7,32 @@ defmodule Blockchain.Contract.CreateContract do
...
+
+  - σ: state,
```
masonforest: I like this format! I think looking up terms will be easier this way and it also cleans up the hash definition. We could probably do some similar refactoring in `apps/evm/lib/evm/gas.ex` but that can be in a follow on PR

germsvel: 👍  I've been trying to update any comments I find in structs to this format in the files I touch.

PR #372, opened 2018-08-21
Fix bug at block #116524

masonforest:
What Changed
============

When `size` was 0 on callcode we were returning 0. Instead we should be
returning the machine state unchanged. This didn't cause an error until
block #116524 where callcode was called with `size = 0`.

Other changes
-----------
* Added a script to run raw EVM code. I used this while debugging this
issue.
* Minor EVM README fix

### DIFF
--- a/apps/evm/lib/evm/operation/environmental_information.ex
+++ b/apps/evm/lib/evm/operation/environmental_information.ex
@@ -205,14 +205,10 @@ defmodule EVM.Operation.EnvironmentalInformation do
         exec_env: exec_env,
         machine_state: machine_state
       }) do
-    if size == 0 do
-      0
-    else
-      data = Memory.read_zeroed_memory(exec_env.machine_code, code_offset, size)
-      machine_state = Memory.write(machine_state, mem_offset, data)
+    data = Memory.read_zeroed_memory(exec_env.machine_code, code_offset, size)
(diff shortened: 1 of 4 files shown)


apps/blockchain/scripts/evm_run.ex
```diff
@@ -0,0 +1,41 @@
+defmodule EVMRun do
```
germsvel: What do you think about `EVM.Runner` or something like that? 

And is there any documentation as to how to use this? It could be just a `moduledoc` here in this file if we don't want to expose this to the world in the README. What do you think?

apps/evm/lib/evm/operation/environmental_information.ex
```diff
@@ -206,7 +206,7 @@ defmodule EVM.Operation.EnvironmentalInformation do
         machine_state: machine_state
       }) do
     if size == 0 do
-      0
+      %{machine_state: machine_state}
```
germsvel: Not sure if it's worth it or not. What do you think about adding a test to `environmental_information_test.exs`?

masonforest: Yep, good call: [
Add a test for EnvironmentalInformation.codecopy where size == 0](https://github.com/poanetwork/mana/pull/372/commits/079c24cd71602f892977af208410fc8dcca0f595)

Issue #330, opened 2018-08-07
Parallelize ci builds


ayrat555:
After we started running Homestead and Frontier tests on circle ci, builds take 10-20 minutes. We need to find a way to parallelize tests

PR #374, opened 2018-08-22
Run Blockchain Tests in different processes

germsvel:
This is a first step for https://github.com/poanetwork/mana/issues/330

What changed?
=============

We are currently running all blockchain tests (that is the ethereum common tests under `/BlockchainTests` directory) in a single ExUnit `test`. So we can't get any of the automatic parallelization that ExUnit's `async: true` provides.

Here we split each hardfork we are testing into a separate process. We then combine the results and assert whether or not the ExUnit test succeeded or failed as a whole.

In a very limited sample (two CI builds), this seems to shave between 6 and 7 minutes off the build: the [latest build](https://circleci.com/workflow-run/0bc29d11-7d5e-494d-a819-7281d37ea600) I had seen was 32:13 minutes, and the two runs for this PR were at [25:46](https://circleci.com/workflow-run/a990c3f1-981d-4420-9bc0-edb232d3622e) and [24:49](https://circleci.com/workflow-run/82f462b2-c70b-4b22-9777-d8774df960e0).

### DIFF
--- a/apps/blockchain/test/blockchain_test.exs
+++ b/apps/blockchain/test/blockchain_test.exs
@@ -27,25 +27,28 @@ defmodule BlockchainTest do
   }
 
   test "runs blockchain tests" do
-    Enum.each(tests(), fn json_test_path ->
-      json_test_path
-      |> read_test()
-      |> ignore_known_failing_tests(json_test_path)
-      |> Enum.filter(&available_chains/1)
-      |> Enum.map(&run_test/1)
-      |> Enum.filter(&failing_tests/1)
-      |> make_assertions()
-    end)
+    for fork <- forks_with_existing_implementation() do
+      tests()
+      |> Enum.reject(&known_fork_failures?(&1, fork))
+      |> Enum.each(fn json_test_path ->
+        json_test_path
+        |> read_test()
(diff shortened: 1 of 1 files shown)


apps/blockchain/test/blockchain_test.exs
```diff
@@ -27,22 +27,86 @@ defmodule BlockchainTest do
...
+    end
+  end
+
+  defp run_tests_for_fork(fork) do
+    tests()
+    |> Stream.reject(&known_fork_failures?(&1, fork))
+    |> Enum.flat_map(fn json_test_path ->
+      json_test_path
+      |> read_test()
+      |> Stream.filter(&fork_test?(&1, fork))
```
ayrat555: what is justification for using `Stream` instead of `Enum`?

germsvel: Since we're creating a lot of intermediate lists in each step, and the list of tests can be big, I thought we could benefit from lazily evaluating those things. That way we aren't constantly building lists over and over again. I'm happy to bring it back to `Enum` if it makes more sense, but it seemed like a good use case to me.

PR #375, opened 2018-08-22
fix circle ci cache

ayrat555:


.circleci/config.yml
```diff
@@ -39,38 +40,41 @@ jobs:
...
+          key: v3-mix-cache-{{ .Branch }}
           paths: "deps"
       - save_cache:
-          key: v2-mix-cache
+          key: v3-mix-cache
           paths: "deps"
 
       - save_cache:
-          key: v2-build-cache-{{ .Branch }}
+          key: v3-build-cache-{{ .Branch }}-{{ checksum "mix.lock" }}
```
ayrat555: I added one more build key

bin/setup
```diff
@@ -103,6 +103,6 @@ fancy_echo "Installing elixir dependencies and compiling."
 mix local.hex --force
 mix local.rebar --force
 mix deps.get
-mix do deps.compile, compile
+mix compile
```
ayrat555: I removed deps.compile. I should be triggered by `mix compile`

PR #380, opened 2018-08-24
fix revert operation

ayrat555:
Changes:
- Revert operation saves remaining gas but reverts states
- Parallelize state tests generation script
- Fix returndatasize, returndatacopy
  - change format of last_return_data from integer list to binary
  - use default methods for reading data with offset from MachineCode


apps/blockchain/scripts/generate_state_tests.ex
```diff
@@ -8,49 +8,49 @@ defmodule GenerateStateTests do
...
-        "Byzantium" => []
-      },
-      failing: %{
-        "Homestead" => [],
-        "Frontier" => [],
-        "EIP150" => [],
-        "EIP158" => [],
-        "Byzantium" => []
-      }
+  @twenty_minutes 60 * 20 * 1000
```
germsvel: What do you think about pulling these changes into its own PR? Since they seem like two independent pieces of work, I think it would make sense to include them separately.

apps/evm/lib/evm/message_call.ex
```diff
@@ -125,43 +125,78 @@ defmodule EVM.MessageCall do
...
+
+  defp finalize_successful_message_call(
+         {gas_remaining, n_sub_state, n_exec_env, output},
+         message_call
+       ) do
+    {out_offset, _out_size} = message_call.output_params
+
+    machine_state =
+      message_call.current_machine_state
+      |> MachineState.push(1)
```
germsvel: I'm not quite certain about the meaning of pushing a `0` vs a `1` when the operation fails vs succeeds. Do you think there would be a way to make that more clear? Maybe we could extract this `1` and the `0` in the `finalize_revert_message_call` into module attributes and name them?

apps/blockchain/test/blockchain/state_test.exs
```diff
@@ -10,454 +10,445 @@ defmodule Blockchain.StateTest do
...
-      "stPreCompiledContracts/modexp_37120_37111_0_22000",
-      "stPreCompiledContracts/modexp_37120_37111_0_20500",
-      "stPreCompiledContracts/modexp_37120_37111_0_155000",
-      "stPreCompiledContracts/modexp_37120_37111_0_1000000",
-      "stPreCompiledContracts/modexp_37120_22411_22000",
+      "stBugs/returndatacopyPythonBug_Tue_03_48_41-1432",
+      "stCreateTest/CreateOOGafterInitCodeReturndata",
       "stPreCompiledContracts/modexp",
-      "stCreateTest/CreateOOGafterInitCodeReturndata"
+      "stPreCompiledContracts/modexp_37120_22411_22000",
```
germsvel: Do you know what changed here? It's hard to see how many tests passed thanks to the fix you made.

apps/evm/lib/evm/message_call.ex
ayrat555: :+1: . Btw it's defined in operation's description in the Yellow Paper

apps/blockchain/test/blockchain/state_test.exs
ayrat555: @germsvel around 40 tests are fixed related to `revert` operation

apps/blockchain/scripts/generate_state_tests.ex
ayrat555: @germsvel https://github.com/poanetwork/mana/pull/381

Issue #382, opened 2018-08-27
EIP-198


ayrat555:
Support for big integer modular exponentiation, enabling RSA signature verification and other cryptographic applications
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-198.md

Byzantium tracker https://github.com/poanetwork/mana/issues/385

PR #383, opened 2018-08-27
implement exp mod builtin

ayrat555:
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-198.md

Resolves https://github.com/poanetwork/mana/issues/382


apps/evm/lib/evm/builtin.ex
```diff
@@ -41,17 +42,12 @@ defmodule EVM.Builtin do
   @spec run_ecrec(EVM.Gas.t(), EVM.ExecEnv.t()) ::
           {EVM.Gas.t(), EVM.SubState.t(), EVM.ExecEnv.t(), EVM.VM.output()}
   def run_ecrec(gas, exec_env = %EVM.ExecEnv{data: data}) do
-    used_gas = @g_ecrec
+    required_gas = @g_ecrec
```
MattMSumner: I like this rename. Since we haven't "used" the gas yet it makes sense that is is how much is required for the work we'd like to do. 👍

apps/evm/lib/evm/builtin.ex
```diff
@@ -138,14 +134,102 @@ defmodule EVM.Builtin do
...
+    if required_gas <= gas do
+      remaining_gas = gas - required_gas
 
       {remaining_gas, %EVM.SubState{}, exec_env, data}
     else
       {0, %EVM.SubState{}, exec_env, :failed}
     end
   end
+
+  @spec exp_mod(EVM.Gas.t(), EVM.ExecEnv.t()) ::
```
MattMSumner: Jumping around the yellow paper; it looks like this is the modifier for "Partial payment for an EXP operation". It looks like this can be 0 or 1. I'm struggling to find a high level definition of what this value represents. I'm assuming a value of 1 means this calculation requires more gas then if it's 0 but it's not clear to me yet why these factors are considered.

I'm leaving this comment here more in case someone can succinctly describe why this is part of the specification rather then for any code changes.

apps/evm/lib/evm/builtin.ex
```diff
@@ -138,14 +134,111 @@ defmodule EVM.Builtin do
...
+          {EVM.Gas.t(), EVM.SubState.t(), EVM.ExecEnv.t(), EVM.VM.output()}
+  def exp_mod(gas, exec_env) do
```
germsvel: What do you think about switching the order of words here to `mod_exp` since its referred to as "modular exponentiation"?

apps/evm/lib/evm/builtin.ex
```diff
@@ -138,14 +134,111 @@ defmodule EVM.Builtin do
...
+    length_data = Memory.read_zeroed_memory(exec_env.data, 0, 96)
```
germsvel: Why do we pass `0` and `96` here? I was looking at EIP 198 and couldn't find those numbers in particular. Do you think we could extract them into variables to give them names?

apps/evm/lib/evm/builtin.ex
```diff
@@ -138,14 +134,111 @@ defmodule EVM.Builtin do
...
+
+    <<b_length_bin::binary-size(32), e_length_bin::binary-size(32),
+      m_length_bin::binary-size(32)>> = length_data
+
+    b_length = :binary.decode_unsigned(b_length_bin)
+    e_length = :binary.decode_unsigned(e_length_bin)
+    m_length = :binary.decode_unsigned(m_length_bin)
```
germsvel: What do you think about calling these `base_length`, `exponent_length` and `modulus_length` to more closely match what EIP 198 calls them? It may also make this code more clear to someone who looks at this code in the future without as much context.

apps/evm/lib/evm/builtin.ex
```diff
@@ -138,14 +134,111 @@ defmodule EVM.Builtin do
...
+
+    if b_length == 0 && m_length == 0 do
+      {gas, %EVM.SubState{}, exec_env, <<0>>}
+    else
+      if b_length <= 24_577 && e_length <= 24_577 && m_length <= 24_577 do
+        calculate_exp_mod(b_length, e_length, m_length, exec_env, gas)
+      else
+        {0, %EVM.SubState{}, exec_env, :failed}
+      end
```
germsvel: What do you think of making these nested `if else` statements into a `cond`?

apps/evm/lib/evm/builtin.ex
```diff
@@ -138,14 +134,111 @@ defmodule EVM.Builtin do
...
+      m_length_bin::binary-size(32)>> = length_data
+
+    b_length = :binary.decode_unsigned(b_length_bin)
+    e_length = :binary.decode_unsigned(e_length_bin)
+    m_length = :binary.decode_unsigned(m_length_bin)
+
+    if b_length == 0 && m_length == 0 do
+      {gas, %EVM.SubState{}, exec_env, <<0>>}
+    else
+      if b_length <= 24_577 && e_length <= 24_577 && m_length <= 24_577 do
```
germsvel: What do you think of extracting the `24_577` into a variable (could be module attribute if it makes sense) to add clarity as to why this is the limit?

apps/evm/lib/evm/builtin.ex
```diff
@@ -138,14 +134,111 @@ defmodule EVM.Builtin do
...
+    end
+  end
+
+  @spec calculate_exp_mod(integer(), integer(), integer(), EVM.ExecEnv.t(), EVM.Gas.t()) ::
+          {EVM.Gas.t(), EVM.SubState.t(), EVM.ExecEnv.t(), EVM.VM.output()}
+  defp calculate_exp_mod(b_length, e_length, m_length, exec_env, gas) do
+    data = Memory.read_zeroed_memory(exec_env.data, 96, b_length + e_length + m_length)
+
+    <<b_bin::binary-size(b_length), e_bin::binary-size(e_length), m_bin::binary-size(m_length)>> =
+      data
```
germsvel: I love that we can do binary pattern matching in Elixir like this. It makes it so nice. 👍

apps/evm/lib/evm/builtin.ex
ayrat555: we pass 96 because the first three numbers (lengths of modulus, base and exponent) have a size of 32 bytes. Yes, I'll extract it into a variable.

PR #388, opened 2018-08-27
Delete unnecessary license files, clean up README docs

andogro:
## What Changed?

- Deleted old MIT license files
- Updated EVM README: I did not do much here as this may fold into the blockchain project as per current discussions. I was unsure how to run the example, and we may want a diagram to make it more clear.
- Blockchain README: fixed relative link. Also, I noticed we are running 0.1.7 but there is no entry in the Changelog for 0.1.7. Do we want to continue with changelogs / versioning for each umbrella app or move to a single changelog?
- ExthCrypto README: Replaced Exthereum with Mana-Ethereum
- Changed urls to common yellow paper link at http://yellowpaper.io/. Not sure if this is standard link but we should choose one.


ayrat555:
@andogro  i think we should also update licence in `mix.exs' file https://github.com/poanetwork/mana/blob/master/apps/evm/mix.exs#L15

ayrat555:
@andogro  I have a post about evm basics https://www.badykov.com/elixir/2018/04/29/evm-basics/ What do you think about adding it to README?

andogro:
I made quite a few changes to the README, incorporating items from Ayrat's post. Any additional helpful items  for someone new to the project can be added to the File Structure Notes section.

apps/evm/README.md
```diff
@@ -1,43 +1,160 @@
...
+      inputs = Operation.inputs(operation, machine_state)
+      
+      IO.puts("stack:")
+      IO.inspect(machine_state.stack)
+      IO.puts("operation: #{operation.sym}")
+      
+      ```
+   - Save the file.
+
+ 4. Start the Elixir REPL.  `cd apps/evm && iex -S mix`
```
germsvel: This is under the `setup` section, but it seems to have more do to with seeing how the EVM works. Should this be more of an example section?

apps/evm/README.md
```diff
@@ -1,43 +1,160 @@
...
+
+   - Find the `cycle/3` method and **add the IO code**.
+      ```elixir
+      def cycle(machine_state, sub_state, exec_env) do
+      operation = MachineCode.current_operation(machine_state, exec_env)
+      inputs = Operation.inputs(operation, machine_state)
+      
+      IO.puts("stack:")
+      IO.inspect(machine_state.stack)
+      IO.puts("operation: #{operation.sym}")
```
germsvel: We also have `EVM.Logger.log_state(operation, machine_state)` which prints out a fairly nice error message, but I imagine it is not as detailed as what you have here.

PR #398, opened 2018-08-30
Add a simple distillery build

masonforest:
[Distillery](https://github.com/bitwalker/distillery) will allow us to
build mana binaries which can be run by users.

This is the default configuration generated by `release.init` new
releases can be generated by running `mix release`.

Other changes:

* Enable sync by default in dev environment
* Use an application configuration value for `db_path` instead of relying on `Mix.env()` as `Mix.env()` isn't available in releases.
* Add debug logging to see when each Supervisor comes online


rel/plugins/.gitignore
```diff
@@ -0,0 +1,3 @@
+*.*
+!*.exs
+!.gitignore
```
MattMSumner: Super minor. But would you mind adding a newline at the end of this file? Mike Burns wrote a fantastic summary of why these get marked by GitHub https://robots.thoughtbot.com/no-newline-at-end-of-file

masonforest:
Closing for now. I'll reopen once we address: https://github.com/poanetwork/mana/issues/409

Issue #394, opened 2018-08-29
EIP-658


germsvel:
Implements https://github.com/ethereum/EIPs/blob/master/EIPS/eip-658.md

Issue #66, opened 2018-05-14
Bloom filter is not used in Transaction.Receipt


vyorkin:
See Eq.(26) in the Yellow Paper.

ayrat555:
yep, I implemented it, but it is not used yet.

vyorkin:
As understand we get the `R_u` (the set of logs created through execution of the transaction, According to the YP (4.3.1)) from the `A_l` (log series that we accumulate in tx sub-state `A` during EVM execution).
So the right place to initialize the bloom filter should be this one:
```
    {new_state, gas_used, logs} = Blockchain.Transaction.execute_transaction(state, trx, header)

    total_gas_used = block.header.gas_used + gas_used

    # TODO: Add bloom filter
    receipt = %Blockchain.Transaction.Receipt{
      state: new_state.root_hash,
      cumulative_gas: total_gas_used,
      logs: logs
    }
```
https://github.com/poanetwork/mana/blob/master/apps/blockchain/lib/blockchain/block.ex#L959

vyorkin:
When this one is finished we need to add the logs bloom equality check to the block holistic validation.

```elixir
    # The following checks Holistic Validity,
    # as defined in Eq.(31), section 4.3.2 of Yellow Paper
    errors =
      []
      |> check_state_root_validity(child_block, block)
      |> check_ommers_hash_validity(child_block, block)
      |> check_transactions_root_validity(child_block, block)
      |> check_logs_bloom(child_block, block) # << smth like that
      |> check_receipts_root_validity(child_block, block)
```

As mentioned in Eq. (31) of the YP.

PR #405, opened 2018-09-04
Implement EIP-658

germsvel:
Resolves https://github.com/poanetwork/mana/issues/394

Summary
========

Adds a transaction status of `0` for failure and `1` for success. This transaction status is included in the transaction receipts if the block number is past the Byzantium hardfork.

More detail
===========

* We update `Transaction.execute`, as well as `MessageCall.execute` and `ContractCreate.execute` to return the status specified as Gamma^z in section 6, Transaction Execution, of the Yellow Paper (0 for failure, 1 for success).

* After the Bynzatium fork, the transaction receipt's state root should include a `0` if the transaction failed or a `1` if it succeeded. This is EIP 658.

### DIFF
--- a/.dialyzer.ignore-warnings
+++ b/.dialyzer.ignore-warnings
@@ -132,6 +132,11 @@ ain.Block':put_state/2. The success typing is (#{'__struct__':='Elixir.Blockchai
 }) -> #{'__struct__':='Elixir.Blockchain.Block', 'header':=#{'__struct__':='Elixir.Block.Header', 's
 tate_root':=_, _=>_}, _=>_}
 
+# We ignore this becuase we cannot reproduce it locally. It may be because CI is
+# using erlang 21 while we are using erlang 20. But we cannot yet update to
+# erlang 21 because we need rox to update
+apps/blockchain/lib/blockchain/block.ex:736: Function create_receipt/6 will never be called
+
 -------------------------------
 # blockchain/chain.ex
 -------------------------------
--- a/apps/blockchain/lib/blockchain/block.ex
+++ b/apps/blockchain/lib/blockchain/block.ex
@@ -697,14 +697,14 @@ defmodule Blockchain.Block do
 
       # Create a contract
   """
-  @spec add_transactions(t, [Transaction.t()], DB.db(), EVM.Configuration.t()) :: t
-  def add_transactions(block, transactions, db, config \\ EVM.Configuration.Frontier.new()) do
+  @spec add_transactions(t, [Transaction.t()], DB.db(), Chain.t()) :: t
+  def add_transactions(block, transactions, db, chain) do
     trx_count = get_transaction_count(block)
 
-    do_add_transactions(block, transactions, db, trx_count, config)
+    do_add_transactions(block, transactions, db, trx_count, chain)
   end
 
-  @spec do_add_transactions(t, [Transaction.t()], DB.db(), integer(), EVM.Configuration.t()) :: t
+  @spec do_add_transactions(t, [Transaction.t()], DB.db(), integer(), Chain.t()) :: t
   defp do_add_transactions(block, [], _, _, _), do: block
 
   defp do_add_transactions(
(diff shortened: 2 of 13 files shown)


.dialyzer.ignore-warnings
```diff
@@ -132,6 +132,8 @@ ain.Block':put_state/2. The success typing is (#{'__struct__':='Elixir.Blockchai
 }) -> #{'__struct__':='Elixir.Blockchain.Block', 'header':=#{'__struct__':='Elixir.Block.Header', 's
 tate_root':=_, _=>_}, _=>_}
 
+apps/blockchain/lib/blockchain/block.ex:736: Function create_receipt/6 will never be called
```
MattMSumner: I'm confused by this message. Why does it think that this function will never be called?

README.md
```diff
@@ -101,7 +101,7 @@ Ethereum common tests are created for all clients to test against. We plan to pr
    - [GeneralStateTests](https://github.com/ethereum/tests/tree/develop/GeneralStateTests) 1174/1180 = 99.5% passing
 - [x] Byzantium
    - [BlockchainTests](https://github.com/ethereum/tests/tree/develop/BlockchainTests) (Includes GeneralStateTests) 4778/4959 = 96.4% passing
-   - [GeneralStateTests](https://github.com/ethereum/tests/tree/develop/GeneralStateTests) 4564/4632= 98.5% passing
+   - [GeneralStateTests](https://github.com/ethereum/tests/tree/develop/GeneralStateTests) 4574/4754= 96.2% passing
```
MattMSumner: Interesting, the percentage went down but the number of tests increased? Did we resync recently? Was it part of this work?

apps/blockchain/lib/blockchain/transaction.ex
```diff
@@ -255,10 +266,18 @@ defmodule Blockchain.Transaction do
 
       # Note, we only want to take the first 3 items from the tuples,
```
MattMSumner: Is this still true now that we're taking all four items?

apps/blockchain/lib/blockchain/transaction.ex
```diff
@@ -255,10 +266,18 @@ defmodule Blockchain.Transaction do
...
       # as designated Θ_3 in the literature Θ_3
-      {state, remaining_gas_, sub_state_, _output} = Contract.message_call(params)
+      {state, remaining_gas_, sub_state_, status} =
```
MattMSumner: What does the extra underscore mean in `remaining_gas_` and `sub_state_`? While we're here how would you feel about removing it? If it's a naming collision, how would you feel about extracting a function to call here?

apps/blockchain/lib/blockchain/block.ex
```diff
@@ -734,7 +730,24 @@ defmodule Blockchain.Block do
       |> put_receipt(trx_count, receipt, db)
       |> put_transaction(trx_count, trx, db)
 
-    do_add_transactions(updated_block, transactions, db, trx_count + 1, config)
+    do_add_transactions(updated_block, transactions, db, trx_count + 1, chain)
+  end
+
+  defp create_receipt(block_header, new_state, total_gas_used, logs, tx_status, chain) do
```
ayrat555: It seems receipts are not fully implemented yet. We are not using bloom filters. Also, I think we need to store them somewhere

germsvel: Yeah, receipts aren't fully implemented yet. There's an issue open for handling those -> https://github.com/poanetwork/mana/issues/66

apps/blockchain/lib/blockchain/transaction.ex
germsvel: I think it's just naming collision. I'll update these to different names. 👍 good catch

README.md
germsvel: We must've resynced recently, and the README wasn't updated. The resync wasn't part of this PR. 

But I ran the `generate_state_tests` task and it showed different numbers. It's hard to tell from just looking at this if more or less tests passed because of the changes in this PR. I can split this into a separate PR if it makes more sense.

apps/blockchain/lib/blockchain/block.ex
germsvel: We should probably add an issue to implement/store receipts. I imagine that's a part of Frontier's implementation that we never got around to doing. I wouldn't want to add that in this PR, since this is just meant to add the transaction status of `0` or `1` to the receipt.

README.md
germsvel: Decided against updating this since no tests passed. 082d0ecc1323c431751c4768d4d4a5e5499caf0f

apps/blockchain/lib/blockchain/transaction.ex
germsvel: Updated this in 0ff7ee6ae5abc3a06d32cd9d9d43e36f24ec8fa7

apps/blockchain/lib/blockchain/transaction.ex
germsvel: Updated this in 0ff7ee6ae5abc3a06d32cd9d9d43e36f24ec8fa7

.dialyzer.ignore-warnings
germsvel: I believe this warning was fixed by the word we did in `execute_with_validation`

Issue #167, opened 2018-06-14
Eth Wire Protocol


germsvel:
We are currently handling the encrypted handshake and deriving secrets needed to frame the rest of the messages. 

Issue #166 will handle the DEVp2p handshake (and other messages). Once that handshake is successful, we can handle subprotocols. 

The main subprotocol is the Ethereum Wire Protocol (https://github.com/ethereum/wiki/wiki/Ethereum-Wire-Protocol). We should handle this first (both v62 and v63), and then deal with Whisper (issue #105) and Swarm (issue #104) later.

masonforest:
Thanks for writing these! It's nice to have a better idea of where we are and where we are going.

I'd say implementing Whisper and Swarm should be postponed until other functionality is implemented because as far as I can tell they aren't part of the core protocol.

@igorbarinov you mentioned Whisper in Cincinnati. Does POA  require the Whisper protocol?

Swarm seems like a separate project all together. I don't see any reference to it in Parity.

It's too bad so much of this is up in the air: part of the nature of decentralization I guess.

germsvel:
Yeah, from my point of view (which is still limited) I think we can defer Whisper and Swarm until much later. I just mentioned them in this issue because I knew we had issues for them, and I know they're mentioned around.

masonforest:
cool, yeah good to have them on the Radar 👍

hayesgm:
This has been completed.

PR #410, opened 2018-09-06
Handle Disconnect, Ping, and Status packets

germsvel:
This is part of https://github.com/poanetwork/mana/issues/166 and https://github.com/poanetwork/mana/issues/167.

Summary
========

The networking layer was not handling Disconnect, Ping, and Status packages. The first two are DEVp2p and the Status packet is the first packet that must be sent for the Ethereum Wire Protocol (after the encrypted and DEVp2p handshakes).

Details
========

Adding Subscribers to Connection struct
-----------------------------------------

First we add `subscribers` to `ExWire.P2P.Connection` struct. For a while I was resisting doing this since I wanted `P2P` to be a pure module without having to know about the socket connection. Ideally it could just respond something like `{:send, packet}` and the `P2P.Server` would handle the actual communication to the socket.

The current implementation of `P2P`, however, is not pure, and it has knowledge of the socket (it's part of the `Connection` struct). So we are including `subscriptions` in there for now, especially since there is a function (`notify_subscribers`) which assumes the struct will have the subscribers in there.

This needs to be handled because our networking layer is technically broken without this part. 

Adds a rough implementation of handling `Packet.Disconnect`, `Packet.Ping`, and `Packet.Status`:
-----------------------------------

* `Disconnect` is a `DEVp2p` message. It can be sent by a peer at any point after the encrypted handshake. It signifies that a disconnect is imminent and that we should disconnect. So we should terminate the TCP connection.

* `Ping` (and `Pong`) are `DEVp2p` messages that should not be delegated to the subscribers. That is, they are related to the networking side of things (ie. "is this node alive?") and not to the blockchain side of things (e.g. "GetBlockHeaders").

* `Packet.Status` is an eth wire protocol packet, and as such, it is possible it should not reside in `P2P` but that it should live with the rest of the packets that are delegated to the subscribers. But since it is the first packet that must come from all eth wire packets, and since it requires a response, we include it here for now. It's also important to note that the status message requires fields like `total_difficulty` and `best_hash`, values which we do not yet know about our blockchain, so we are simply mirroring what we get from the other node for now.

Other changes
-------------

* Makes PeerSupervisor (previously a regular supervisor) into a DynamicSupervisor. This means it can dynamically start children nodes, which will be outbound `ExWire.P2P.Server`s. We add a `PeerSupervisor.start_child/1` function that takes in a peer enode url to start the outbound server.

### DIFF
--- a/apps/ex_wire/config/dev.exs
+++ b/apps/ex_wire/config/dev.exs
@@ -2,7 +2,10 @@ use Mix.Config
 
 config :ex_wire,
   network_adapter: {ExWire.Adapter.UDP, NetworkClient},
-  sync: false,
+  sync: true,
+  private_key:
+    <<10, 122, 189, 137, 166, 190, 127, 238, 229, 16, 211, 182, 104, 78, 138, 37, 146, 116, 90,
+      68, 76, 86, 168, 24, 200, 155, 0, 99, 58, 226, 211, 30>>,
   discovery: true,
   node_discovery: [
     network_adapter: {ExWire.Adapter.UDP, NetworkClient},
--- a/apps/ex_wire/.iex.exs
+++ b/apps/ex_wire/.iex.exs
@@ -1,0 +1,1 @@
+alias ExWire.{PeerSupervisor}
--- a/apps/ex_wire/lib/ex_wire.ex
+++ b/apps/ex_wire/lib/ex_wire.ex
@@ -20,7 +20,7 @@ defmodule ExWire do
         db = MerklePatriciaTree.DB.RocksDB.init(db_name())
 
         [
-          worker(ExWire.PeerSupervisor, [ExWire.Config.bootnodes()]),
+          supervisor(ExWire.PeerSupervisor, [:ok]),
           worker(ExWire.Sync, [db])
         ]
       else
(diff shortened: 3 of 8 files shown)


apps/ex_wire/config/dev.exs
```diff
@@ -2,7 +2,10 @@ use Mix.Config
 
 config :ex_wire,
   network_adapter: {ExWire.Adapter.UDP, NetworkClient},
-  sync: false,
+  sync: true,
+  private_key:
+    <<10, 122, 189, 137, 166, 190, 127, 238, 229, 16, 211, 182, 104, 78, 138, 37, 146, 116, 90,
```
germsvel: I wasn't sure if I should leave this hardcoded key here. 

It is often a source of confusion _not_ to have it, because without this we will generate a random key every time we interact with a message from a peer. That means we'll use one key to encrypt secrets and one key to decrypt them. And it is a difficult bug to trace.

apps/ex_wire/config/dev.exs
```diff
@@ -2,7 +2,10 @@ use Mix.Config
 
 config :ex_wire,
   network_adapter: {ExWire.Adapter.UDP, NetworkClient},
-  sync: false,
+  sync: true,
```
germsvel: I also left the `sync` option to true since it seems like a more sensible default in my mind, but I can set that back to `false` if that makes more sense to people.

PR #422, opened 2018-09-12
use correct client name

ayrat555:


apps/blockchain/README.md
```diff
@@ -1,4 +1,4 @@
-# Mana-Ethereum Blockchain [![CircleCI](https://circleci.com/gh/exthereum/blockchain.svg?style=svg)](https://circleci.com/gh/exthereum/blockchain)
```
MattMSumner: I'm guessing we're removing this here since we have the badge in the main README?

ayrat555: yep

PR #423, opened 2018-09-12
Turns EVM.address => Account.Address in interface

germsvel:
Block number 177610 in the mainnet has an account that is pushed into
the stack as an integer. The account when represented as an integer does
not have a leading zero, but when represented as hex, it has `00` as its
first two characters.

This led to a subtle bug, where the balance of the account in question
seemed to be zero, when in reality we were checking the wrong account's
balance.

To fix this issue, we ensure that the addresses coming from the EVM (via
the account interface) are all turned into
`Blockchain.Account.Addresses` which must have `160` bits. If they are
short, as was the case in the bug, we pad them.

Additionally:
------------

The SubState's touched accounts should have the binary format for the
address of an account. The `create` system operation was passing the
result of turning an address and nonce into an unsigned integer into the
substate. It should have been passing the `exec_env.address` instead.

### DIFF
--- a/apps/evm/lib/evm/operation/system.ex
+++ b/apps/evm/lib/evm/operation/system.ex
@@ -77,18 +77,20 @@ defmodule EVM.Operation.System do
       end
 
     # Note if was exception halt or other failure on stack
-    result =
+    new_address =
       if status == :ok do
         nonce = AccountInterface.get_account_nonce(exec_env.account_interface, exec_env.address)
 
         Address.new(exec_env.address, nonce)
       else
-        0
+        <<0>>
       end
+
+    new_address_for_machine_state = :binary.decode_unsigned(new_address)
 
     machine_state = %{
       machine_state
-      | stack: Stack.push(machine_state.stack, result),
+      | stack: Stack.push(machine_state.stack, new_address_for_machine_state),
         gas: n_gas + remaining_gas
     }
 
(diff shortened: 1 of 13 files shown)


apps/evm/lib/evm/operation/system.ex
```diff
@@ -97,7 +97,7 @@ defmodule EVM.Operation.System do
     sub_state =
       n_sub_state
       |> SubState.merge(sub_state)
-      |> SubState.add_touched_account(result)
+      |> SubState.add_touched_account(exec_env.address)
```
ayrat555: Why did you change touched account here?

germsvel: The SubState's touched accounts should have the binary format for the address of an account. This operation was passing the `result` which is an unsigned integer, when it should have been passing the `exec_env.address` instead.

ayrat555: but we should add to touched accounts a created address, not address from exec env

germsvel: Oh that's right. We need to change this to the created address, but it should be the binary format. I know there's a similar `Contract.Address.new` in the blockchain that does the same thing. We really should have one way of creating new addresses, but we'll look into that. Good catch!

germsvel: We updated this in https://github.com/poanetwork/mana/pull/423/commits/8872f8fb557939d9ec1788952d3f520344dc7632

PR #426, opened 2018-09-13
Create2 [EIP-1014]

ayrat555:
Resolves https://github.com/poanetwork/mana/issues/421

Remaining failing create2 tests are related to `sstore` gas cost changes in EIP 1283 which we didn't implement yet


.dialyzer.ignore-warnings
```diff
@@ -39,26 +39,28 @@
...
+apps/evm/lib/evm/gas.ex:308: Function operation_cost/3 has no local return
+apps/evm/lib/evm/gas.ex:308: The call 'Elixir.EVM.Gas':operation_cost(__@1::any(),__@2::any(),__@3::any(),'nil') breaks the contract (atom(),['Elixir.EVM':val()],['Elixir.EVM':val()],'Elixir.EVM.MachineState':t()) -> t() | 'nil'
+apps/evm/lib/evm/gas.ex:573: Function gas_cost_for_nested_operation/2 will never be called
 apps/evm/lib/evm/machine_state.ex:53: Function subtract_gas/2 has no local return
 apps/evm/lib/evm/operation/environmental_information.ex:114: Function calldataload/2 has no local return
 apps/evm/lib/evm/operation/environmental_information.ex:117: The call 'Elixir.EVM.Helpers':decode_signed(binary()) breaks the contract (integer() | 'nil') -> 'Elixir.EVM':val() | 'nil'
 apps/evm/lib/evm/vm.ex:98: The pattern 'continue' can never match the type {'halt','invalid_instruction' | 'stack_underflow' | 'undefined_instruction'}
+apps/evm/lib/evm/operation/system.ex:24: Function create/2 has no local return
```
masonforest: I remember seeing these errors when @germsvel and I were pairing. @germsvel did we end up figuring out what causes these? We should probably try to avoid adding dialyzer warnings if we can 🙂

apps/evm/lib/evm/address.ex
```diff
@@ -46,6 +46,36 @@ defmodule EVM.Address do
...
+    binary_init_code =
+      if is_binary(init_code) do
+        init_code
+      else
+        :binary.encode_unsigned(init_code)
+      end
+
+    code_hash = Keccak.kec(binary_init_code)
+
+    (<<255>> <> binary_address <> binary_salt <> code_hash)
```
masonforest: would it make sense to pull `<<255>>` out into a module variable or a constant?

apps/evm/lib/evm/operation/system.ex
```diff
@@ -297,7 +231,10 @@ defmodule EVM.Operation.System do
   `GeneralStateTests/stSystemOperationsTest/suicideSendEtherPostDeath.json`.
   """
   @spec selfdestruct(Operation.stack_args(), Operation.vm_map()) :: Operation.op_result()
-  def selfdestruct([refund_address], %{exec_env: exec_env, sub_state: sub_state}) do
+  def selfdestruct(
+        [refund_address],
+        %{exec_env: exec_env, sub_state: sub_state}
+      ) do
```
germsvel: Did the formatter do this? Can it be a single line?

.dialyzer.ignore-warnings
germsvel: I think it usually means that the function might raise an error. Note that it could be a function call that `create/2` is making within it that would raise the error, but the point remains. I think this is telling us that somewhere in `create/2` there's a possibility of an error being raised and so the function wouldn't have a local return. 

I agree that we should avoid adding dialyzer warnings if we can, especially when we're dealing with `evm` code. @ayrat555 did you try to figure out why this is now showing up when it wasn't before? Maybe the new code path added has the possibility of raising an error?

apps/blockchain/lib/blockchain/interface/account_interface.ex
```diff
@@ -382,6 +382,7 @@ defimpl EVM.Interface.AccountInterface, for: Blockchain.Interface.AccountInterfa
           EVM.MachineCode.t(),
           integer(),
           Block.Header.t(),
+          EVM.address(),
```
germsvel: Since you're passing `nil` in the doc test above, I think dialyzer will immediately say this is incorrect (even if we're suppressing the warning) because this argument should be `EVM.address() | nil` . Should we add the `nil` to the typespec or does it make sense to never pass `nil` into this?

ayrat555: no, we can't.  because created account address is only passed if a contract is being created with opcode ( from evm). Contract creation transactions pass nil

.dialyzer.ignore-warnings
ayrat555: Currently, I'm trying to figure out what causes this warning

PR #428, opened 2018-09-13
Update common-tests

MattMSumner:


apps/blockchain/test/blockchain_test.exs
```diff
@@ -251,9 +186,6 @@ defmodule BlockchainTest do
       "Byzantium" ->
         Chain.load_chain(:byzantium_test, config)
 
-      "Constantinople" ->
```
germsvel: This might be why you're not getting any constantinople tests to run, right?

PR #430, opened 2018-09-14
fix output in message call

ayrat555:
We did not check `out_size` parameter when writing output after message calls. This value can be zero so we shouldn't write anything to the memory and change `last_return_data`


apps/blockchain/test/blockchain/state_test.exs
```diff
@@ -54,7 +52,16 @@ defmodule Blockchain.StateTest do
       "stCreate2/Create2OOGafterInitCodeReturndataSize",
       "stCreate2/Create2OOGafterInitCodeRevert",
       "stCreate2/Create2OOGafterInitCodeRevert2",
+      "stCreate2/Create2OnDepth1023",
```
ayrat555: these tests are not related to fix. they are from tests submodule update

PR #432, opened 2018-09-18
Update dialyxir

MattMSumner:


.tool-versions
```diff
@@ -1,3 +1,3 @@
 rust 1.26.0
-elixir 1.7.2
-erlang 20.0
+elixir 1.7.2-otp-21
```
MattMSumner: I've updated these to match what we have on circle to try and reproduce locally running dialyzer without emitting any warnings.

mix.exs
```diff
@@ -18,6 +18,8 @@ defmodule Mana.MixProject do
       start_permanent: Mix.env() == :prod,
       dialyzer: [
         ignore_warnings: ".dialyzer.ignore-warnings",
+        plt_add_apps: [:mix],
+        plt_add_deps: :transitive,
```
MattMSumner: This could be removed. My understanding is that it's ensuring we include types from dependencies.

mix.exs
```diff
@@ -18,6 +18,8 @@ defmodule Mana.MixProject do
       start_permanent: Mix.env() == :prod,
       dialyzer: [
         ignore_warnings: ".dialyzer.ignore-warnings",
+        plt_add_apps: [:mix],
```
MattMSumner: This was added to remove some annoying warnings about mix tasks.

.tool-versions
ayrat555: I think it's wrong because we can now only use otp 20

ayrat555: til https://github.com/urbint/rox/pull/15 is merged

germsvel: The weird thing is that CI uses otp-21 and all the tests pass. I guess the tests are using `ets` instead of `rox`?

ayrat555: yep @germsvel

Issue #431, opened 2018-09-18
EIP-1283: Net gas metering for SSTORE operations 


ayrat555:
https://eips.ethereum.org/EIPS/eip-1283

Issue #416, opened 2018-09-09
Work with cache instead of storage in evm


ayrat555:
Currently, we're committing changes to storage on every evm operation that works with storage. A better solution would be to maintain some kind of cache and commit this cache after a successful evm execution.

This approach may be the solution to https://github.com/poanetwork/mana/issues/229

PR #433, opened 2018-09-19
 EIP-1283

ayrat555:
Resolves https://github.com/poanetwork/mana/issues/431

I created a cache for account storage modification in evm, So now we won't modify Merkle Patricia trie storage on sstore operations and only commit account storage changes on successful transaction execution ( for both message calls and contract creations).

### DIFF
--- a/apps/blockchain/lib/blockchain/interface/account_interface.ex
+++ b/apps/blockchain/lib/blockchain/interface/account_interface.ex
@@ -3,12 +3,24 @@ defmodule Blockchain.Interface.AccountInterface do
   Defines an interface for methods to interact with contracts and accounts.
   """
 
+  alias Blockchain.Account.Address
+
+  @type cache :: %{
+          Address.t() => %{
+            key: integer(),
+            initial_value: integer(),
+            current_value: integer()
+          }
+        }
+
   @type t :: %__MODULE__{
-          state: EVM.state()
+          state: EVM.state(),
+          cache: cache()
         }
 
   defstruct [
-    :state
+    :state,
+    :cache
   ]
 
   @doc """
@@ -28,7 +40,8 @@ defmodule Blockchain.Interface.AccountInterface do
   @spec new(EVM.state()) :: t
   def new(state) do
     %__MODULE__{
-      state: state
+      state: state,
+      cache: %{}
     }
   end
 end
@@ -185,7 +198,16 @@ defimpl EVM.Interface.AccountInterface, for: Blockchain.Interface.AccountInterfa
           {:ok, integer()} | :account_not_found | :key_not_found
   def get_storage(account_interface, evm_address, key) do
     address = Account.Address.from(evm_address)
-    Account.get_storage(account_interface.state, address, key)
+
+    address_cache = Map.get(account_interface.cache, address)
+
+    if address_cache do
+      value = Map.get(address_cache, :current_value)
+
+      if is_nil(value), do: :key_not_found, else: {:ok, value}
+    else
+      Account.get_storage(account_interface.state, address, key)
+    end
   end
 
   @spec account_exists?(AccountInterface.t(), EVM.address()) :: boolean()
@@ -225,7 +247,18 @@ defimpl EVM.Interface.AccountInterface, for: Blockchain.Interface.AccountInterfa
     if Account.get_account(account_interface.state, address) do
       updated_state = Account.put_storage(account_interface.state, address, key, value)
 
-      %{account_interface | state: updated_state}
+      cache_change = %{address => %{current_value: value, key: key}}
+
+      updated_cache =
+        Map.merge(account_interface.cache, cache_change, fn _k,
+                                                            old_address_cache,
+                                                            new_address_cache ->
+          Map.merge(old_address_cache, new_address_cache, fn _k, _value, new_value ->
+            new_value
+          end)
+        end)
+
+      %{account_interface | state: updated_state, cache: updated_cache}
     else
       account_interface
     end
@@ -238,7 +271,18 @@ defimpl EVM.Interface.AccountInterface, for: Blockchain.Interface.AccountInterfa
     if Account.get_account(account_interface.state, address) do
       updated_state = Account.remove_storage(account_interface.state, address, key)
 
-      %{account_interface | state: updated_state}
+      cache_change = %{address => %{current_value: nil, key: key}}
+
+      updated_cache =
+        Map.merge(account_interface.cache, cache_change, fn _k,
+                                                            old_address_cache,
+                                                            new_address_cache ->
+          Map.merge(old_address_cache, new_address_cache, fn _k, _value, new_value ->
+            new_value
+          end)
+        end)
+
+      %{account_interface | state: updated_state, cache: updated_cache}
     else
       account_interface
     end
--- a/apps/blockchain/lib/blockchain/interface/account_interface.ex
+++ b/apps/blockchain/lib/blockchain/interface/account_interface.ex
@@ -2,20 +2,12 @@ defmodule Blockchain.Interface.AccountInterface do
   @moduledoc """
   Defines an interface for methods to interact with contracts and accounts.
   """
-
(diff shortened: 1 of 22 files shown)


apps/blockchain/lib/blockchain/interface/account_interface/cache.ex
```diff
@@ -0,0 +1,71 @@
+defmodule Blockchain.Interface.AccountInterface.Cache do
+  alias Blockchain.Account.Address
+  defstruct cache: %{}
+
+  @type key_cache() :: %{
+          integer() => %{current_value: integer() | :deleted, initial_value: integer() | nil}
```
ayrat555: while working on https://github.com/poanetwork/mana/issues/416, I will try using `ets` tables as cache storage instead of simple maps

germsvel:
@ayrat555 do you think you could split the cache work and the EIP work into two separate PRs? It would make it easier to review.

ayrat555:
@germsvel I don't think so because for EIP-1283 we should have access to the current value. It's not possible without cache

apps/evm/lib/evm/refunds.ex
```diff
@@ -85,18 +86,70 @@ defmodule EVM.Refunds do
...
+      initial_value != current_value && initial_value != 0 && current_value == 0 ->
+        -15_000
+
+      initial_value != current_value && initial_value != 0 && new_value == 0 ->
+        15_000
+
+      initial_value != current_value && initial_value == new_value && initial_value == 0 ->
+        19_800
+
+      initial_value != current_value && initial_value == new_value && initial_value != 0 ->
```
masonforest: Might be out of the scope of this PR/overkill but this is getting pretty complex. Do we want to pull it out into `refunds/sstore.ex` or something?

apps/evm/lib/evm/refunds.ex
```diff
@@ -85,18 +86,70 @@ defmodule EVM.Refunds do
 
   # `SSTORE` operations produce a refund when storage is set to zero from some non-zero value.
   def refund(:sstore, [key, new_value], _machine_state, _sub_state, exec_env) do
```
masonforest: Might be out of the scope of this PR/overkill but this is getting pretty complex. Do we want to pull it out into `refunds/sstore.ex` or something?

apps/blockchain/lib/blockchain/interface/account_interface.ex
```diff
@@ -185,7 +208,26 @@ defimpl EVM.Interface.AccountInterface, for: Blockchain.Interface.AccountInterfa
           {:ok, integer()} | :account_not_found | :key_not_found
   def get_storage(account_interface, evm_address, key) do
     address = Account.Address.from(evm_address)
-    Account.get_storage(account_interface.state, address, key)
+    cached_value = Cache.get_current_value(account_interface.cache, address, key)
+
+    case cached_value do
+      nil -> Account.get_storage(account_interface.state, address, key)
+      :deleted -> :key_not_found
```
germsvel: We're returning `:key_not_found` to match the behavior of the `Account.get_storage` when a key isn't found, right? If so, I wonder if there's a better way to consolidate the knowledge of `:key_not_found` somewhere else? For example, maybe the `Account` module itself makes use of the cache if it finds one?

apps/blockchain/test/blockchain/interface/account_interface/cache_test.exs
```diff
@@ -0,0 +1,127 @@
+defmodule Blockchain.Interface.AccountInterface.CacheTest do
+  use ExUnit.Case, async: true
+
+  alias Blockchain.Interface.AccountInterface.Cache
+
+  alias Blockchain.Interface.AccountInterface.Cache
```
germsvel: I think this `alias` is duplicate. You have the same one on line 4

apps/blockchain/test/blockchain/interface/account_interface/cache_test.exs
```diff
@@ -0,0 +1,127 @@
...
+
+  describe "update_current_value/4" do
+    test "sets current_value" do
+      cache = %Cache{cache: %{}}
```
germsvel: I don't think you need `cache: %{}` inside. That's the default for the `Cache` struct, right? You should be able to just do `%Cache{}`.

apps/blockchain/test/blockchain/interface/account_interface/cache_test.exs
```diff
@@ -0,0 +1,127 @@
...
+      result = Cache.update_current_value(cache, address, key, value)
+      expected_result = %{<<1>> => %{2 => %{current_value: 6}}}
+
+      assert result.cache == expected_result
+    end
+  end
+
+  describe "add_initial_value/4" do
+    test "sets initial_value" do
+      cache = %Cache{cache: %{}}
```
germsvel: same comment here about not needing `%{}` inside since it's the default

apps/blockchain/lib/blockchain/interface/account_interface/cache.ex
```diff
@@ -0,0 +1,71 @@
...
+          cache: account_cache()
+        }
+
+  @spec get_current_value(t(), Address.t(), integer()) :: integer() | nil
+  def get_current_value(cache_struct, address, key) do
+    get_key_value(cache_struct.cache, address, key, :current_value)
+  end
+
+  @spec get_initial_value(t(), Address.t(), integer()) :: integer() | nil
+  def get_initial_value(cache_struct, address, key) do
```
germsvel: This is stylistic, so feel free to ignore, but what do you think of renaming these two functions that start with `get_*` into the same names without the `get_*`? 

```
Cache.get_current_value -> Cache.current_value

Cache.get_initial_value -> Cache.initial_value
```

apps/evm/lib/evm/gas.ex
```diff
@@ -347,23 +347,27 @@ defmodule EVM.Gas do
...
-    case ExecEnv.get_storage(exec_env, key) do
-      :account_not_found ->
-        @g_sset
-
-      :key_not_found ->
-        if new_value != 0 do
+    if Configuration.eip1283_sstore_gas_cost_changed?(exec_env.config) do
+      eip1283_sstore_gas_cost([key, new_value], exec_env)
+    else
+      case ExecEnv.get_storage(exec_env, key) do
```
germsvel: What do you think about extracting all this logic into a private function? That way you're dealing with the same level of abstraction in the `operation_cost` function. For example,

```elixir
def operation_cost(:sstore, [key, new_value], _machine_state, exec_env) do
  if Configuration.eip1283_sstore_gas_cost_changed?(exec_env.config) do 
    eip1283_sstore_gas_cost([key, new_value], exec_env)
else
  # you might be able to think of a better name for this function
   basic_sstore_gas_cost([key, new_value], exec_env)
end
```

apps/evm/lib/evm/gas.ex
```diff
@@ -598,4 +602,32 @@ defmodule EVM.Gas do
       {:original, original_cost}
     end
   end
+
+  defp eip1283_sstore_gas_cost([key, new_value], exec_env) do
```
germsvel: 👍

apps/evm/lib/evm/refunds/sstore.ex
```diff
@@ -0,0 +1,73 @@
...
+
+  # Refund given (added into refund counter) when the storage value is set to zero from non-zero.
+  @storage_refund 15_000
+
+  @spec refund({integer(), integer()}, ExecEnv.t()) :: integer()
+  def refund({key, new_value}, exec_env) do
+    if Configuration.eip1283_sstore_gas_cost_changed?(exec_env.config) do
+      eip1283_sstore_refund([key, new_value], exec_env)
+    else
+      case ExecEnv.get_storage(exec_env, key) do
```
germsvel: Love that we have extracted this into its own module. What do you think of extracting this `case` statement into a private function in this module so you have the same level of abstraction for the `refund` function?

Issue #424, opened 2018-09-12
Sync block 179098


germsvel:
https://github.com/poanetwork/mana/pull/423 allows for syncing past block 177610. We are now failing to verify block 179098.

Issue #229, opened 2018-07-06
Inconsistent state on storage clearing in sstore operation


ayrat555:
When clearing values in any account's storage, it affects other accounts.

The problem in `Trie.remove/2` method

hayesgm:
@ayrat555 Was this addressed / fixed by #435?

hayesgm:
What's the right solution here? What *should* we do?

ayrat555:
@hayesgm I think the problem is caused by the fact that we're storing all account storages in the same db so collisions occur when storage tries share the same node. Maybe we somehow can store them in a way that won't cause collisions.

PR #435, opened 2018-09-19
Do not remove storage in Account.del_account

germsvel:
Closes https://github.com/poanetwork/mana/issues/424

What?
=====

The deletion of storage currently has problems. We have encountered an
issue in block `179098` where we are trying to load values from storage
(via `SLOAD`) but the storage was previously deleted (in block `116172`)
because the account in that block has the _exact same storage root_ as
the account in block `179098`. Since deleting the storage of an account
means deleting the trie found at the storage root of that account,
deleting the storage in block `116172` removed the storage for block
`179098`.

Data from block `116172`:
------------------------

transaction number:
"0x47ba29b3a6598bc3f3bc79627f87bfee909edf7090314efac26fa58c8f81e916"

Account that was self destroyed:
"0x49e77af7f52eec4b042a52b957c96df1d7c1d31b"

Data from block `179098`:
------------------------
transaction number:
"0x65ac7e42338d8dee38ecb0525bee45bbdf8b3ad77c4478797adf166d3c7bf9b5"

Account that performed SLOAD and returned 0 incorrectly:
"0x9168414b724033d9ab0bc6dba06e26c851190953"

Co-authored-by: Matthew Sumner <matt.m.sumner@gmail.com>

### DIFF
--- a/apps/blockchain/lib/blockchain/account.ex
+++ b/apps/blockchain/lib/blockchain/account.ex
@@ -218,9 +218,7 @@ defmodule Blockchain.Account do
       nil ->
         state
 
-      acc ->
-        storage_trie = Trie.new(state.db, acc.storage_root)
-        Trie.Storage.delete(storage_trie)
+      _acc ->
         Trie.remove(state, Keccak.kec(address))
     end
   end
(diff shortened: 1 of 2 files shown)


ayrat555:
Are you sure that the problem was in storage deletion?

MattMSumner:
We're relatively confident that's the problem but we won't know until we see the sync_with_infura script get further down the chain.

germsvel:
@ayrat555 I still think we need to address the issue of storage raised by https://github.com/poanetwork/mana/issues/229

PR #438, opened 2018-09-21
EIP-1283 improvements

ayrat555:
Changes:
- Clean cache code
- Add eip1283 tests from  https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1283.md#test-cases
    - fix found bug
- Clean refunds, gas cost code

### DIFF
--- a/apps/blockchain/test/blockchain/contract/message_call_test.exs
+++ b/apps/blockchain/test/blockchain/contract/message_call_test.exs
@@ -160,4 +160,322 @@ defmodule Blockchain.Contract.MessageCallTest do
       assert output == :failed
     end
   end
+
+  # https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1283.md#test-cases
+  describe "eip1283 sstore gas cost changes" do
+    setup do
+      db = MerklePatriciaTree.Test.random_ets_db(:eip1283_test)
+
+      state =
+        db
+        |> Trie.new()
+        |> Account.put_account(<<0x10::160>>, %Account{balance: 1_000_000})
+        |> Account.put_account(<<0x20::160>>, %Account{balance: 1_000_000})
+
+      config = EVM.Configuration.Constantinople.new()
+      updated_config = %{config | eip1283_sstore_gas_cost_changed: true}
+
+      available_gas = 10_000_000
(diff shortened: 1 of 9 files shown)


apps/evm/lib/evm/refunds/sstore.ex
```diff
@@ -35,23 +39,35 @@ defmodule EVM.Refunds.Sstore do
...
+      initial_value == current_value && new_value == 0 ->
         15_000
 
-      initial_value != current_value && initial_value != 0 && current_value == 0 ->
-        -15_000
+      true ->
+        first_refund =
+          cond do
+            initial_value != 0 && current_value == 0 ->
+              -15_000
```
masonforest: can we pull these out into module variables like we did with the other gas values?

apps/blockchain/test/blockchain/contract/message_call_test.exs
```diff
@@ -160,4 +160,322 @@ defmodule Blockchain.Contract.MessageCallTest do
       assert output == :failed
     end
   end
+
+  # https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1283.md#test-cases
```
germsvel: 🎉

mix.exs
```diff
@@ -38,7 +38,7 @@ defmodule Mana.MixProject do
   defp deps do
     [
       {:ex_doc, "~> 0.19", only: :dev, runtime: false},
-      {:dialyxir, "~> 1.0.0-rc.3", only: [:dev, :test], runtime: false},
```
germsvel: Any reason why this change is in here? I think Dialyxir's new warnings are far better.

MattMSumner: If you downgraded to get the output to put into the ignore file, on the newer version you can run `mix dialyzer --format dialyzer` to get the old format.

apps/evm/lib/evm/refunds/sstore.ex
ayrat555: done

mix.exs
ayrat555: yep, exactly that was a problem for me. But I returned the new version

PR #440, opened 2018-09-24
return account_interface instead of state

ayrat555:
Part of https://github.com/poanetwork/mana/issues/416

To work with cache instead of permanent storage in evm we need to add accounts state (balance changes etc) caching to `AccountInterface`. 

Currently, we are returning state with applied changes from `CreateContract` and `MessageCall` modules.
 
Changes:
- return account_interface from `CreateContract` and `MessageCall` to apply changes in `Transaction` module


ayrat555:
@germsvel I think in our situation `AccountInterface` is an interface to the world state. `state` field is an implementation detail of the world state (in `evm` it's map, in `blockchain` it's merkle trie)

PR #441, opened 2018-09-24
cache storage operations

ayrat555:
with https://github.com/poanetwork/mana/pull/440 resolves https://github.com/poanetwork/mana/issues/416

Changes: 
- Do not modify Merkle Patricia Tree during EVM execution (it's read-only)
- Commit storage changes only on successful EVM execution
- Use temporary storage in EVM

### DIFF
--- a/apps/blockchain/lib/blockchain/contract/create_contract.ex
+++ b/apps/blockchain/lib/blockchain/contract/create_contract.ex
@@ -197,7 +197,7 @@ defmodule Blockchain.Contract.CreateContract do
     if params.new_account_address do
       params.new_account_address
     else
-      sender_account = Account.get_account(params.account_interface.state, params.sender)
+      {sender_account, _} = AccountInterface.account(params.account_interface, params.sender)
       Account.Address.new(params.sender, sender_account.nonce)
     end
   end
--- a/apps/blockchain/lib/blockchain/contract/create_contract.ex
+++ b/apps/blockchain/lib/blockchain/contract/create_contract.ex
@@ -51,7 +51,7 @@ defmodule Blockchain.Contract.CreateContract do
   def execute(params) do
     original_account_interface = params.account_interface
     contract_address = new_account_address(params)
-    {account, _} = AccountInterface.account(original_account_interface, contract_address)
+    account = AccountInterface.account(original_account_interface, contract_address)
 
     if is_nil(account) || Account.uninitialized_contract?(account) do
       result = {rem_gas, _, _, output} = create(params, contract_address)
@@ -129,7 +129,7 @@ defmodule Blockchain.Contract.CreateContract do
 
   @spec init_account(t, EVM.address()) :: AccountInterface.t()
   defp init_account(params, address) do
-    {account, _code} = AccountInterface.account(params.account_interface, address)
+    account = AccountInterface.account(params.account_interface, address)
 
     account_interface =
       if is_nil(account) do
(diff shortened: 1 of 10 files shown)


apps/blockchain/lib/blockchain/interface/account_interface.ex
```diff
@@ -37,9 +40,144 @@ defmodule Blockchain.Interface.AccountInterface do
...
+    %{account_interface | cache: updated_cache}
+  end
+
+  @spec transfer_wei!(t(), Address.t(), Address.t(), EVM.Wei.t()) :: t()
+  def transfer_wei!(account_interface, from, to, wei) do
+    {from_account, from_account_code} = account(account_interface, from)
+
+    cond do
+      wei < 0 ->
+        raise("wei transfer cannot be negative")
```
masonforest: Is there a reason to use `raise` over  returning an `:error/:ok` tuple here?

apps/blockchain/test/blockchain/interface/account_interface_test.exs
```diff
@@ -2,4 +2,233 @@ defmodule Blockchain.Interface.AccountInterfaceTest do
...
+        state
+        |> Account.reset_account(address)
+        |> AccountInterface.new()
+
+      {result, nil} = AccountInterface.account(account_interface, address)
+
+      assert result == %Blockchain.Account{
+               balance: 0,
+               code_hash:
+                 <<197, 210, 70, 1, 134, 247, 35, 60, 146, 126, 125, 178, 220, 199, 3, 192, 229,
```
masonforest: might be nice to name this `empty_keccak` or something as a helper method or a module variable

apps/blockchain/lib/blockchain/interface/account_interface.ex
ayrat555: this method is replacing `Account.transfer!` method that is used in AccountInterface protocol implementation. `Account.transfer!` also raises errors

apps/blockchain/lib/blockchain/contract/create_contract.ex
```diff
@@ -119,28 +119,26 @@ defmodule Blockchain.Contract.CreateContract do
...
     }
 
     EVM.VM.run(params.available_gas, exec_env)
   end
 
-  @spec init_account(t, EVM.address()) :: EVM.state()
+  @spec init_account(t, EVM.address()) :: AccountInterface.t()
   defp init_account(params, address) do
-    account = Account.get_account(params.account_interface.state, address)
+    {account, _code} = AccountInterface.account(params.account_interface, address)
```
germsvel: Why does `AccountInterface.account` return an `{account, code}` tuple? It seems like it should just return an account, right? As far as I can tell, most places just ignore the `code` part of the tuple.

apps/blockchain/lib/blockchain/interface/account_interface.ex
```diff
@@ -37,9 +40,144 @@ defmodule Blockchain.Interface.AccountInterface do
...
+    |> Cache.commit(account_interface.state)
+    |> new()
+  end
+
+  @spec commit_storage(t()) :: t()
   def commit_storage(account_interface) do
-    Cache.commit(account_interface.cache, account_interface.state)
+    new_state = Cache.commit_storage(account_interface.cache, account_interface.state)
+
+    new(new_state, %Cache{accounts_cache: account_interface.cache.accounts_cache})
```
germsvel: Should we have functions in `Cache` that allow us to create/update the accounts_cache? That way, the `Cache` module is solely responsible for manipulating the data. What do you think?

apps/blockchain/lib/blockchain/contract/create_contract.ex
ayrat555: I added a method that returns only account

apps/blockchain/lib/blockchain/interface/account_interface.ex
ayrat555: @germsvel This method was unused. I remove it. Cache is committed only with `AccountInterface.commit/1` which delegates to `Cache.commit`

PR #443, opened 2018-09-25
Add a `sync_with_file` mix task

masonforest:


apps/blockchain/scripts/sync_from_file.ex
```diff
@@ -0,0 +1,84 @@
...
+  SyncFromFile allows users to sync the blockchain from a file exported by parity.
+
+  You can make an export with the following command:
+
+  `parity export blocks ./export-blocks-mainnet.bin`
+
+  And then run `sync_with_file` to verify the chain:
+
+  `mix run apps/blockchain/scripts/sync_from_file.ex export-blocks-mainnet.bin`
+
```
germsvel: I think you have an extra space here

apps/blockchain/scripts/sync_from_file.ex
```diff
@@ -0,0 +1,84 @@
+require Logger
+
+defmodule SyncFromFile do
```
germsvel: I've been wondering, would it make sense to write these as mix tasks? Then we (and others) can get access to the docs more easily. Mix tasks have docs when you do, 

```
# uses @shortdoc
mix help 

# or

# users @moduledoc
mix help task_name
```

What do you think?

apps/blockchain/lib/mix/tasks/sync_from_file.ex
```diff
@@ -0,0 +1,83 @@
...
+  @moduledoc """
+  SyncFromFile allows users to sync the blockchain from a file exported by parity.
+
+  You can make an export with the following command:
+
+  `parity export blocks ./export-blocks-mainnet.bin`
+
+  And then run `sync_with_file` to verify the chain:
+
+  `mix run apps/blockchain/scripts/sync_from_file.ex export-blocks-mainnet.bin`
```
germsvel: Should this now be `mix sync_from_file`? Should we change the name to something like `mix sync.from_file` or is that worse?

PR #444, opened 2018-09-25
Create common_tests.run mix task

germsvel:
What changed?
============

Creates `mix common_tests.run` mix task to run a single blockchain test file, with an optional hardfork.

This task uses much of the same logic as does the `blockchain_test.exs` test. So we extract a `BlockchainTestRunner` module to house all that common logic related to running a single blockchain test file.

Why?
===

When we debug failing blockchain tests, it is often hard to run a single test. It is possible to do so by modifying certain pieces of the `test/blockchain_test.exs` file, but that requires some knowledge of the structure of the test setup, and it is both cumbersome and time consuming. With this new mix task, we can simply specify the file we need to run and it will run that test.

A note on mix task
==================

The advantage of creating this helpful script as a `mix` task (rather than just creating it as we have created other scripts) is that it allows us to have some clear documentation.

- `mix help` will list all the task available in the project, including `common_tests.run`. It will use the `@shortdoc` for that.

screen shot 2018-09-25 at 9 07 20 am

- `mix help common_tests.run` will give a full description of the mix task, using the information found in `moduledoc`.

screen shot 2018-09-25 at 9 07 41 am

### DIFF
--- a/apps/blockchain/test/blockchain_test.exs
+++ b/apps/blockchain/test/blockchain_test.exs
@@ -3,10 +3,7 @@ defmodule BlockchainTest do
 
   import EthCommonTest.Helpers
 
-  alias Blockchain.{Blocktree, Account, Transaction, Chain}
-  alias MerklePatriciaTree.Trie
-  alias Blockchain.Account.Storage
-  alias Block.Header
+  alias Blockchain.Chain
 
   doctest Blockchain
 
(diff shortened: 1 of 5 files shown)


apps/blockchain/test/blockchain_test.exs
```diff
@@ -117,11 +118,7 @@ defmodule BlockchainTest do
     tests
     |> Stream.reject(&known_fork_failures?(&1, fork))
     |> Enum.flat_map(fn json_test_path ->
-      json_test_path
-      |> read_test()
-      |> Stream.filter(&fork_test?(&1, fork))
-      |> Stream.map(&run_test/1)
-      |> Enum.filter(&failing_test?/1)
+      BlockchainTestRunner.run(json_test_path, fork)
```
ayrat555: good code reuse :+1:

ayrat555:
I think we need the same task for state tests

germsvel:
Yeah, getting one for state tests would be great! I might add that soon. I was just adding this because I was dealing with some blockchain tests not passing in a branch, so I wanted to improve this.

PR #447, opened 2018-09-25
Implement EIP-1234

germsvel:
This resolves https://github.com/poanetwork/mana/issues/436

Summary
=======

This implements EIP-1234 which updates the difficulty calculation and the block reward for Constantinople.

More details
===========

Thus far we have been using Parity's chain config files to generate configuration for a chain. We update those files to include the values for Constantinople.

In this change, however, Parity updates not only some values in the json configuration but also how they are stored. In particular, the block rewards and the bomb delay factor can now be a single value or json objects whose key is the block_number of the fork in question and the value is the thing looked for. So, for example, we could find this in a block_reward,

``` block_reward: {
"0" => 300000000,
"1700000" => 2000000
}
```

We make updates to our `Chain` module to account for this.


MattMSumner:
This adds some previously passing tests to the failing tests lists per https://github.com/ethereum/tests/issues/512

PR #448, opened 2018-09-25
Save state every 1000 blocks in SyncWithInfura

masonforest:
Before we were only saving the state when validation failed. This was
fine early on when failures were more common but now validation errors
are few and father between. Now we save state periodically so if the
task fails due to a catastrophic failure (such as a memory overflow) we
won't have to resync.

### DIFF
--- a/apps/blockchain/scripts/sync_with_infura.ex
+++ b/apps/blockchain/scripts/sync_with_infura.ex
@@ -1,6 +1,7 @@
 require Logger
 
 defmodule SyncWithInfura do
+  @save_block_interval 1000
   require Logger
 
   def setup() do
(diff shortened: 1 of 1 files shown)


apps/blockchain/scripts/sync_with_infura.ex
```diff
@@ -1,6 +1,7 @@
 require Logger
 
 defmodule SyncWithInfura do
+  @save_block_interval 1000
```
ayrat555: Maybe we can add a couple of saving points for debugging purposes?

masonforest: Like save at certain blocks like `500000`? If we want to stop at a certain block we've been adding an if statement inside `add_block_to_tree`. Could we achieve the same goal by doing that?

PR #449, opened 2018-09-25
Add mix state_tests.run task

MattMSumner:
What changed?
============

Adds `mix state_tests.run` task to enable us to run a single state test
file quickly. The mix task also takes an optional `--fork` flag to only
run tests in that hardfork.

This task is the counterpart to the blockchain task `mix
common_tests.run` for state tests.

Since this mix task uses much of the same logic as the
`test/blockchain/state_test.exs` we extract that into `StateTestRunner`
module.

### DIFF
--- a/apps/blockchain/test/blockchain/state_test.exs
+++ b/apps/blockchain/test/blockchain/state_test.exs
@@ -1,9 +1,6 @@
 defmodule Blockchain.StateTest do
   alias MerklePatriciaTree.Trie
-  alias Blockchain.{Account, Transaction}
-  alias Blockchain.Interface.AccountInterface
-  alias Blockchain.Account.Storage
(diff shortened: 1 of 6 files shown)


masonforest:
This will make running these so much easier 🎉

apps/blockchain/test/blockchain/state_test.exs
```diff
@@ -186,6 +171,31 @@ defmodule Blockchain.StateTest do
     |> is_nil()
   end
 
+  def configuration(hardfork) do
```
ayrat555: maybe we can reuse the method from https://github.com/poanetwork/mana/blob/master/apps/blockchain/lib/blockchain/chain.ex#L163 ?

PR #454, opened 2018-09-27
Use EIP161 chain config for EIP158

germsvel:
It turns out we have been using `EIP150`'s chain configuration for `EIP158` (during tests). A more accurate chain to use would be `EIP161` since `EIP161` is what superseded `EIP158`. We should also consider renaming EIP150 -> TangerineWhistle and EIP58 (EIP161) -> SpuriousDragon, but we leave that for future work.

This change does not have an effect in number of tests passed since most of the values that differ between EIP150 and EIP158 are being sourced via the `EVM.Configuration.EIP158` and not via the chain struct.

### DIFF
--- a/apps/blockchain/lib/blockchain/chain.ex
+++ b/apps/blockchain/lib/blockchain/chain.ex
@@ -148,7 +148,7 @@ defmodule Blockchain.Chain do
         load_chain(:eip150_test, config)
 
       "EIP158" ->
-        load_chain(:eip150_test, config)
+        load_chain(:eip161_test, config)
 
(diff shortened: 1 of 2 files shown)


germsvel:
@ayrat555 are you saying you're on board with switching to these? 

`EIP150` -> `TangerineWhistle`, 
`EIP158` -> SpuriousDragon

ayrat555:
@germsvel yep

germsvel:
@ayrat555 I'll merge this as is, since it's ready to go, and open a separate PR with those other changes (since I haven't done those yet).

Issue #450, opened 2018-09-26
store Receipts in the storage


ayrat555:
Related issue - https://github.com/poanetwork/mana/issues/66

PR #455, opened 2018-09-27
[WIP] Receipts

ayrat555:
Resolves https://github.com/poanetwork/mana/issues/450


apps/blockchain/lib/blockchain/transaction.ex
```diff
@@ -362,23 +365,36 @@ defmodule Blockchain.Transaction do
...
+      if Header.mined_by?(header, address) do
+        Account.reset_account(state, address)
       else
         Account.del_account(new_state, address)
       end
     end)
   end
 
+  @spec maybe_reset_coinbase(EVM.state(), EVM.SubState.t(), Header.t()) :: EVM.state()
+  defp maybe_reset_coinbase(state, sub_state, header) do
```
ayrat555: In the current implementation, we're resetting the coinbase account in `clean_up_accounts_marked_for_destruction` if the coinbase was self-destructed. But receipts expect state with the coinbase deleted not resetted. So we should reset coinbase after setting `root_hash` of the receipt.

Currently this method is not used, it will be used after turning on receipts validation in `HollisticValidity`

apps/blockchain/lib/blockchain/transaction.ex
```diff
@@ -439,6 +455,10 @@ defmodule Blockchain.Transaction do
     end
   end
 
+  defp create_receipt(state_root_hash, gas_used, logs, _status_code) do
```
germsvel: We're not using the `status_code` here, should we be using it? If I recall correctly, the `state_root` needed to change based on whether or not the transaction succeeded or failed.

apps/blockchain/lib/blockchain/transaction/receipt.ex
```diff
@@ -25,6 +40,34 @@ defmodule Blockchain.Transaction.Receipt do
           logs: EVM.SubState.logs()
         }
 
+  @spec new(EVM.trie_root(), EVM.Gas.t(), EVM.SubState.logs()) :: t()
+  def new(state_root, gas_used, logs) do
+    bloom_filter =
+      logs
+      |> Enum.reduce(Bloom.empty(), fn log_entry, acc ->
```
germsvel: Is this something that could reside within `Bloom` itself? Maybe something like` Bloom.from_logs(logs)`?

apps/blockchain/test/blockchain/block_test.exs
```diff
@@ -337,11 +337,22 @@ defmodule Blockchain.BlockTest do
 
       assert Enum.count(block.transactions) == 1
 
-      expected_receipt = %Transaction.Receipt{
-        bloom_filter: "",
+      expected_receipt = %Blockchain.Transaction.Receipt{
+        bloom_filter:
+          <<0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
```
germsvel: Is there an easier way to represent this? Maybe something like `<<0::2048>>`? (not sure if this is something like `256 * 8 = 2048` or something else. I'm just guessing here.)

apps/blockchain/test/blockchain/block_test.exs
```diff
@@ -337,11 +337,22 @@ defmodule Blockchain.BlockTest do
...
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0>>,
         cumulative_gas: 28_180,
         logs: [],
-        state: block.header.state_root
+        state:
+          <<61, 115, 119, 182, 91, 209, 64, 217, 127, 241, 158, 23, 15, 12, 238, 234, 204, 16, 1,
```
germsvel: Why is this no longer equal to `block.header.state_root`? If it is now equal to some other value, could we make that the expectation? The reason why I bring this point is that when we use raw values like this one in tests, and someone else comes looking later, it is really hard to understand what this test is expecting. Clearly, it expects this binary to be there. But why this binary and where did it come from?

apps/blockchain/test/blockchain/transaction/receipt/bloom_test.exs
```diff
@@ -0,0 +1,14 @@
+defmodule Blockchain.Transaction.Receipt.BloomTest do
+  use ExUnit.Case
+  alias Blockchain.Transaction.Receipt.Bloom
+
+  describe "contains?/2" do
+    test "adds value to new filter and checks if it is in filter" do
```
germsvel: Should this just say `"checks if value is in filter"` since it's a test under `contains?/2` or does the function also add a value to the bloom filter?

apps/blockchain/test/blockchain/transaction/receipt_test.exs
```diff
@@ -18,4 +18,226 @@ defmodule Blockchain.Transaction.ReceiptTest do
...
+        33, 208, 147, 66, 139, 9, 108, 165, 1, 167, 205, 26, 116, 8, 85, 167, 151, 111, 192, 160,
+        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+        0, 0, 247, 148, 220, 244, 33, 208, 147, 66, 139, 9, 108, 165, 1, 167, 205, 26, 116, 8, 85,
+        167, 151, 111, 192, 160, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+        0, 0, 0, 0, 0, 0, 0, 0, 0, 0>>
+
+    assert encoded == expectected_result
+  end
+
+  test "decodes receipt with multiple different logs" do
```
germsvel: Not sure if it would be helpful, but what do you think of adding a test that decodes a bloom filter, then encodes the result, and we check that they are the same? It would in a sense test the same thing these tests are testing, but it also adds the clarity that the decoding/encoding are inverse operations, right? Anyway, just an idea. Love the thorough testing here though. 👍

apps/blockchain/lib/blockchain/transaction.ex
ayrat555: yes, it's not used yet. I was gonna add in the next PRs.

apps/blockchain/test/blockchain/transaction/receipt_test.exs
ayrat555: @germsvel I think the first test tests it https://github.com/poanetwork/mana/blob/72f5d2d97edae204a3469618b18082bf55a968c2/apps/blockchain/test/blockchain/transaction/receipt_test.exs#L14

apps/blockchain/test/blockchain/block_test.exs
ayrat555: @germsvel Actually, it still equals to `block.header.state_root`. My bad

apps/blockchain/test/blockchain/block_test.exs
ayrat555: @germsvel I added default value to the `Receipt` struct

apps/blockchain/lib/blockchain/block/holistic_validity.ex
```diff
@@ -82,7 +82,6 @@ defmodule Blockchain.Block.HolisticValidity do
       |> check_transactions_root_validity(child_block, block)
       |> check_gas_used(child_block, block)
 
-    # TODO: Check receipts after resolving https://github.com/poanetwork/mana/issues/66
     # |> check_receipts_root_validity(child_block, block)
```
masonforest: Do we have an issue for enabling `check_receipts_root_validity` here or will that be included in [530: store Receipts in the storage](https://github.com/poanetwork/mana/issues/450)

ayrat555: I created an issue https://github.com/poanetwork/mana/issues/459

PR #462, opened 2018-10-01
Configure the EVM based on the block number

masonforest:
What changed
============

Before we were configuring the EVM manually. Now we use the correct
configuration based on the block number of the block we are verifying.

Other changes:
* Add named hardfork block numbers to blockchain configurations.

Parity configuration files have block numbers set for each EIP.
Configuring block numbers based on named hardforks is simpler an maps
better to how `EVM.Configuration` is set up.

* Rename EVM.Configuration.EIP158 to SpuriousDragon for consistency


apps/blockchain/lib/blockchain/chain.ex
```diff
@@ -161,6 +164,32 @@ defmodule Blockchain.Chain do
     end
   end
 
+  @doc """
+  Get the EVM configuration based on the chain and block number
+  """
+  def evm_config(chain = %__MODULE__{}, block_number \\ nil) do
```
ayrat555: I didn't get the idea. shouldn't all methods in configurations now accept block_number? or block_number will be stored in the configuration before passing it to evm?

masonforest: The correct EVM configuration [is chosen](https://github.com/poanetwork/mana/blob/mf-use-correct-evm-config/apps/blockchain/lib/blockchain/block.ex#L725) before processing a transaction based on block number. The idea being that our configuration won't change mid-transaction.

When processing a transaction any call to a configuration method will call the correct fork module `Configuration.Homestead`/ `Configuration.Frontier` etc.

apps/evm/lib/evm/configuration.ex
```diff
@@ -17,25 +17,25 @@ defmodule EVM.Configuration do
...
-  # EIP150
+  # TangerineWhistle
   @callback extcodecopy_cost(t) :: integer()
 
-  # EIP150
+  # TangerineWhistle
   @callback balance_cost(t) :: integer()
 
-  # EIP150
+  # TangerineWhistle
```
ayrat555: What do you think about leaving EIP numbers in comments?

masonforest: Ah yep, I had just done a find and replace 👍 -> [Leave EIP numbers in comments](https://github.com/poanetwork/mana/pull/462/commits/5a1ab993ce2ba64a4f5ac9724b1157cbc232dd2a)

Issue #456, opened 2018-09-28
Refactor `Blocktree.do_get_path_to_root`


masonforest:
Currently `Blocktree.do_get_path_to_root` calls itself recursively to build a path of hashes to the root of the tree.  This could probably be rewritten more efficiently possibly making use of `Enum.reduce`.

ayrat555:
https://github.com/poanetwork/mana/issues/446

Issue #446, opened 2018-09-25
Look into `blocktree` overflowing memory


masonforest:
`sync_with_infura` crashed at block `492326` with the following error

````
03:00:09.061 [info]  Verified Block 492326
eheap_alloc: Cannot allocate 1898305688 bytes of memory (of type "old_heap").
````

I think that's because the entire block tree is stored in a nested map.

That means if we have half a million blocks we have half a million nested maps.

Blocks are added to the blocktree [here](https://github.com/poanetwork/mana/blob/master/apps/blockchain/scripts/sync_with_infura.ex#L15) and get added to the map [here](https://github.com/poanetwork/mana/blob/master/apps/blockchain/lib/blockchain/blocktree.ex#L204).

There may be a way to store this map in the database instead. If not we may need to require clients to have greater system requirements or add swap memory.

Here's specs on the RAM of the server this crashed on:

````
              total        used        free      shared  buff/cache   available
Mem:           3.9G         60M        1.6G         40M        2.2G        3.5G
Swap:            0B          0B          0B
````

PR #464, opened 2018-10-02
do not keep all blocks in memory

ayrat555:
fixes https://github.com/poanetwork/mana/issues/456
fixes https://github.com/poanetwork/mana/issues/446

We don't have to keep in memory all blocks from the beginning. I think this data structure was needed when the best block was assigned by traversing full blocktree. But the algorithm of finding the best block was changed in https://github.com/poanetwork/mana/pull/303. Now, to find the best block we need only the current block and previous best block.

### DIFF
--- a/apps/blockchain/lib/blockchain/blocktree.ex
+++ b/apps/blockchain/lib/blockchain/blocktree.ex
@@ -2,18 +2,6 @@ defmodule Blockchain.Blocktree do
   @moduledoc """
   Blocktree provides functions for adding blocks to the
   overall blocktree and forming a consistent blockchain.
-
-  We have two important issues to handle after we get a new
-  unknown block:
-
-  1. Do we accept the block? Is it valid and does it connect to
-     a known parent?
-
-  2. After we've accepted it, is it (by total difficulty) the canonical block?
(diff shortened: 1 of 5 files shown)


masonforest:
Looks good! I'm always in for removing code if we can!

I still haven't fully wrapped my head around what happens in this case:
                           
                    A
                   /    \
                  B     C
                          |
                         D

First block A comes in, it's the best Block. Block B and C come in at the same time. B has higher difficulty so B is now the best block. Some of our peers never saw B so they mine on C. One of them mines a block on C and D comes in. I think that means we have to switch the best block to D because that's the longer chain. In regards to this PR: wouldn't we have discarded C already at this point because previously B was the best block?

Also, not related to this PR, do we need to keep state from previous blocks for this case? I asked @5chdn about that [here](https://poanetwork.slack.com/archives/GA89BNBT7/p1526309572000791). Slack or Hangouts might be a better channel to discuss that but I wanted to mention it here.

ayrat555:
@masonforest 
Firstly, we store all valid blocks.
When `D` blocks comes we retrieve its parent - `C`. If `D`.difficulty > `B`.difficulty, `D` will become the best block

Issue #459, opened 2018-10-01
validate receipts root hash when adding blocks


ayrat555:
We don't validate receipts root https://github.com/poanetwork/mana/blob/master/apps/blockchain/lib/blockchain/block/holistic_validity.ex#L86

PR #465, opened 2018-10-02
validate receipts

ayrat555:
resolves https://github.com/poanetwork/mana/issues/459


apps/blockchain/lib/blockchain/transaction.ex
```diff
@@ -211,19 +211,49 @@ defmodule Blockchain.Transaction do
 
     {expended_gas, refund} = calculate_gas_usage(tx, remaining_gas, sub_state)
 
+    {account_interface_after_receipt, receipt} =
+      if empty_contract_creation?(tx) &&
+           Configuration.for(chain.evm_config).clean_touched_accounts?(chain.evm_config) do
+        account_interface_after_execution = AccountInterface.commit(updated_account_interface)
```
MattMSumner: It looks like there are some order dependence stuff here that I'm not quite sure I understand.

Here we're creating the account first, creating the receipt and then doing clean up and refunds.

In the other branch it looks like we clean up accounts and refunds then commit the account? After that we make the receipt.

Does the receipt depend on the account state? Should we always be creating the receipt at the end? Is it important that the account is created first in the empty contract case?

apps/blockchain/lib/blockchain/transaction.ex
```diff
@@ -211,19 +211,49 @@ defmodule Blockchain.Transaction do
 
     {expended_gas, refund} = calculate_gas_usage(tx, remaining_gas, sub_state)
 
+    {account_interface_after_receipt, receipt} =
+      if empty_contract_creation?(tx) &&
```
germsvel: There seems to be a lot of repetition in this `if else` condition, but it seems like the _timing_ of when we perform the `AccountInterface.commit` matters. Is that correct? 

If so, is there a different way we could express this? I have a hard time understanding why we need to commit before we pay and refund gas and clean accounts in the `if` statement and after in the `else` statement.

PR #467, opened 2018-10-02
EIP-155 signature schema

MattMSumner:
Why:

We should be validating signatures using v = 27 | 28 | CHAIN_ID * 2 + 35
| CHAIN_ID * 2 + 36 but we are currently only using v = 27 | 28.

This PR:

Adds the additional valid values for v. This means we need to pass
around the chain instead of only evm_config for validating sender
signatures.

### DIFF
--- a/apps/blockchain/scripts/sync_with_infura.ex
+++ b/apps/blockchain/scripts/sync_with_infura.ex
@@ -104,8 +104,8 @@ defmodule SyncWithInfura do
       end
 
     Blockchain.Block.add_ommers(block, ommers)
-  catch
-    :exit, error ->
+  rescue
+    error ->
       Logger.info("error:  #{inspect(error)}")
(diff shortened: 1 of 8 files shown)


apps/blockchain/scripts/sync_with_infura.ex
```diff
@@ -6,7 +6,8 @@ defmodule SyncWithInfura do
 
   def setup() do
     db = MerklePatriciaTree.DB.RocksDB.init(db_name())
-    chain = Blockchain.Chain.load_chain(:foundation)
+    config = EVM.Configuration.EIP158.new()
```
MattMSumner: Maybe I should remove these changes...

germsvel: Yeah, I think we should default the `sync_with_infura` to mainnet for now

PR #470, opened 2018-10-03
remove ropsten.dat file

ayrat555:
This file contains the first 11 blocks of Ropsten and was used for debugging. 

We passed these blocks so now this file is not needed


masonforest:
🎉 

Should we remove [the test](https://github.com/poanetwork/mana/blob/master/apps/blockchain/test/ropsten_test.exs) as well?

ayrat555:
@masonforest thanks. I removed it

PR #471, opened 2018-10-03
Create SyncWithInfura mix task

### LINKED ISSUE #453: Convert  SyncWithInfura into a mix task
Currently `SyncWithInfura` is [a script](https://github.com/poanetwork/mana/blob/master/apps/blockchain/scripts/sync_with_infura.ex).

We should convert it to a mix task like `Sync.ToFile` so we get the mix documentation that comes with that.

bradleystachurski:
Closes #453 

# What changed?

Creates a mix task `sync.with_infura` with the same functionality as `apps/blockchain/scripts/sync_with_infura.ex`. This script has been removed since it can now be run as a mix task.

@andogro This PR also updates the `README.md`, which appears to be your domain. Feel free to make any edits, but please note that changes in `README.md` should also be reflected in the documentation in `apps/blockchain/lib/mix/tasks/sync/with_infura.ex`.

# Why is this needed?

`sync_with_infura` is used frequently. Newcomers to the project may benefit from the additional mix documentation from running `mix help sync.with_infura`. Also, `mix sync.with_infura` is more tidy than `mix run apps/blockchain/scripts/sync_with_infura.ex`.


bradleystachurski:
`dialyzer` fails with the following message:

```
apps/blockchain/lib/mix/tasks/sync/with_infura.ex:96:call
The call:
Ethereumex.HttpClient.eth_get_block_by_number(<<_::16, _::size(8)>>, true)

will never return since it differs in arguments with
positions 2 from the success typing arguments:

(binary(), binary())
________________________________________________________________________________
done (warnings were emitted)
Exited with code 2
```

The call to `Ethereumex.HttpClient.get_block_by_number/3` is the same call that was made in the `sync_with_infura` script, so it seems odd to me that it fails for the new mix task.

I see a couple potential remedies.

1. Change the `typespec` in `Ethereumex.HttpClient.get_block_by_number/2` from
```
      @spec eth_get_block_by_number(binary(), binary(), keyword()) :: {:ok, map()} | error
```
to
```
      @spec eth_get_block_by_number(binary(), boolean(), keyword()) :: {:ok, map()} | error
```

2. Add `apps/blockchain/lib/mix/tasks/sync/with_infura.ex` to `.dialyzer.ignore-warnings`

Does anyone have an opinion on which solution to choose or have a different idea to fix?

bradleystachurski:
Okey dokey, thanks for the feedback.

PR #473, opened 2018-10-04
Make BlockInterface into a behaviour

germsvel:
Description
===========

Makes the `EVM.Interface.BlockInterface` into a behaviour, and updates `Blockchain.Interface.BlockInterface` and `EVM.Interface.Mock.MockBlockInterface` into modules to satisfy that behaviour.

While we're doing this, we take the opportunity to rename the modules since `Interface` does not seem to fit as well. We rename `BlockInterface` as `BlockHeaderInfo` since the behaviour is about retrieving information from the block header.

We also create some helper functions in `EVM.BlockHeaderInfo` to make dynamic dispatch easier.

Why?
---

We were using a protocol to abstract the notion of what a `BlockInterface` is. But a protocol is not the correct abstraction for a polymorphic module that responds to the same functions. For that we want to use behaviours.

Other notes
-----------

While doing this update, we noticed that the function `get_block_by_hash` in the protocol was not used at all (only privately in one of the implementations). So we remove that from the protocol (now behaviour).


apps/blockchain/lib/blockchain/block_header_info.ex
```diff
@@ -0,0 +1,93 @@
...
+  def get_block_header(block_header_info) do
+    block_header_info.block_header
+  end
+
+  @doc """
+  Returns a block that is `n` blocks before the current block.
+
+  ## Examples
+
+      iex> db = MerklePatriciaTree.Test.random_ets_db()
```
ayrat555: What do you think about moving this "example" to a test file?

.dialyzer.ignore-warnings
```diff
@@ -45,6 +34,8 @@ apps/evm/lib/evm/gas.ex:309: Function operation_cost/3 has no local return
 apps/evm/lib/evm/gas.ex:309: The call 'Elixir.EVM.Gas':operation_cost(__@1::any(),__@2::any(),__@3::any(),'nil') breaks the contract (atom(),['Elixir.EVM':val()],['Elixir.EVM':val()],'Elixir.EVM.MachineState':t()) -> t() | 'nil'
 apps/evm/lib/evm/gas.ex:576: Function gas_cost_for_nested_operation/2 will never be called
 apps/evm/lib/evm/machine_state.ex:53: Function subtract_gas/2 has no local return
+apps/evm/lib/evm/machine_state.ex:55: The pattern {'changed', _cost@1, _new_call_gas@1} can never match the type number() | {'original',number()}
```
MattMSumner: Could this possibly be a real error? I'm not sure what exposed this but it does look like the spec for Gas.cost would never match this case?
```ex
# apps/evm/lib/evm/gas.ex:136
  @spec cost(MachineState.t(), ExecEnv.t(), keyword()) :: t | nil | {atom(), t | nil}

# apps/evm/lib/evm/machine_state.ex:54
    case Gas.cost(machine_state, exec_env, with_status: true) do
      {:changed, cost, new_call_gas} ->

```

apps/evm/test/evm/interface/block_interface_test.exs
```diff
@@ -0,0 +1,34 @@
+defmodule EVM.BlockHeaderInfoTest do
```
MattMSumner: How would you feel about renaming this file to have behavior instead of interface in the name?

.dialyzer.ignore-warnings
germsvel: Yeah, I think it makes sense to look into these and see if they're real errors

apps/evm/test/evm/interface/block_interface_test.exs
germsvel: Oh no! I missed the name of the file? I thought I double checked that. I'll definitely change that

apps/blockchain/lib/blockchain/block_header_info.ex
germsvel: @ayrat555 we already have a very similar test in the test file. Would you prefer removing this example altogether or should we keep it?

apps/evm/test/evm/interface/block_interface_test.exs
germsvel: Updated in 95df82539e1ea1c1c3706123d5f1359aafdebea3

.dialyzer.ignore-warnings
germsvel: After further research, I think we should look at these errors, but not in relation to this PR. I think they are unrelated to this. I just improved a typespec that surfaced these errors.

apps/blockchain/lib/blockchain/block_header_info.ex
ayrat555: yes, I think we should get rid of these large doctests

germsvel: Sounds good to me! Updated in e0efead483f9cf729d1da10376a41b654508951a

PR #476, opened 2018-10-05
increase `CREATE2` memory cost

ayrat555:
This PR implements changes to [EIP1014](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1014.md#gas-cost)
`The CREATE2 has the same gas schema as CREATE, but also an extra hashcost of GSHA3WORD * ceil(len(init_code) / 32), to account for the hashing that must be performed.`

### DIFF
--- a/apps/evm/lib/evm/gas.ex
+++ b/apps/evm/lib/evm/gas.ex
@@ -219,7 +219,8 @@ defmodule EVM.Gas do
   end
 
   def memory_cost(:create2, [_value, in_offset, in_length, _salt], machine_state) do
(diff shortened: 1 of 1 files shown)


apps/blockchain/scripts/generate_state_tests.ex
```diff
@@ -66,11 +65,10 @@ defmodule GenerateStateTests do
       failing_tests = length(completed_tests[:failing][hardfork])
       total_tests = passing_tests + failing_tests
 
-      IO.puts(
```
masonforest: nice, this always bugged me

PR #482, opened 2018-10-09
use ipc client for synchronisation

ayrat555:
There is a [new feature](https://github.com/exthereum/ethereumex/blob/master/CHANGELOG.md#040) in `ethereumex`. Now we can connect to a ethereum node using `IpcClient`. It uses unix sockets so it is faster and potentially more secure


apps/blockchain/config/dev.secret.exs.example
```diff
@@ -1,2 +1,6 @@
 use Mix.Config
-config :ethereumex, url: "https://mainnet.infura.io/<your api key here>"
+
+config :ethereumex,
+  url: "https://mainnet.infura.io/<your api key here>",
+  ipc_path: "/.local/share/io.parity.ethereum/jsonrpc.ipc",
```
MattMSumner: Will this path work on OSX? Assuming parity is installed via home-brew?

ayrat555: yep, it should work. parity's file should be locate in "~/.local/share/io.parity.ethereum/jsonrpc.ipc"
ipc server is started by default

Issue #479, opened 2018-10-08
Fix consensus bug at block #1237678 on mainnet


masonforest:
We’re failing to match consensus at block [#1237678](https://etherscan.io/block/1237678)

The only contract transaction is [0x1066b](https://etherscan.io/tx/0x1066b9764832dc6dc710c3e131ee54c68c1b3f1e21de6b3dc8f97ee68ff3f190) and I’m almost certain that’s the culprit.

We loose consensus at program counter position `1202` where on the canonical chain 15 gas is used on the MSTORE operation and in our implementation only `6` gas is used.

If you replace line [271 of gas.ex](https://github.com/poanetwork/mana/blob/master/apps/evm/lib/evm/gas.ex#L271) with the following:

    memory_expansion_value = if machine_state.program_counter == 1202  do
      memory_expansion_value(machine_state.active_words, offset, length, machine_state) + 3
    else
      memory_expansion_value = memory_expansion_value(machine_state.active_words, offset, length, machine_state)
    end

The block will validate.

I’ve used [Geth transaction tracing](https://github.com/ethereum/go-ethereum/wiki/Tracing:-Introduction) and wrote [a script](https://gist.github.com/masonforest/12f651445f3e9406c83fba2c7d82c69f) to compare our implementation with the Geth tracing logs. Here’s the [raw trace file](https://gist.github.com/masonforest/00b6bdb43582ca2b5e5fba22e4dc6a69).

You can test this by logging into the PoA testing server and running `sync_with_infura` on the branch “mf-fix-block-1237678”

The geth trace has 128 bytes of additional zeroed memory when the MSTORE is made when the program counter is at `1202`. Our implementation doesn’t have that. Possibly memory allocation is suppose to happen at a different time when validating blocks?

I’ll post progress here if I discover any additional information. If anyone else discovers anything please do the same.

masonforest:
Fixed by f68b87a57e00183c8547933a826a4ff0f4597def

PR #486, opened 2018-10-09
Fix memory error at 1417555

masonforest:
[Transaction 0x9d1b51c65d0624985e41447a7966bf7287a08dc9334b85dac9e0a1ab5dd7a9c1](https://etherscan.io/tx/0x9d1b51c65d0624985e41447a7966bf7287a08dc9334b85dac9e0a1ab5dd7a9c1) calls a `CODECOPY` with offset > 0xffffffffe0
causing the memory to overflow. Since the data size zero the write can
be skipped all together.

### DIFF
--- a/apps/evm/lib/evm/memory.ex
+++ b/apps/evm/lib/evm/memory.ex
@@ -71,6 +71,9 @@ defmodule EVM.Memory do
           MachineState.t()
   def write(machine_state, offset_bytes, original_data, size \\ nil)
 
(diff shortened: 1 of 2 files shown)


germsvel:
@masonforest is this a fix for https://github.com/poanetwork/mana/issues/479? Or was that somewhere else (since this has a different block number)?

masonforest:
@germsvel this was a later block and I forgot to create an issue for it. I'll close #479 as it's fixed now 👍

PR #495, opened 2018-10-11
Fix dialyzer warnings for AccountRepo implementations

germsvel:
This fixes the majority of the ignored dialyzer warnings for `blockchain/account/repo.ex`, `blockchain/account/repo/cache.ex`, `evm/mock/mock_account_repo.ex`, `evm/account_repo.ex`. They were ignored as part of https://github.com/poanetwork/mana/pull/494. 

Description
===========

We start by removing the ignored warnings for `blockchain/account/repo` and `evm/mock/mock_account_repo`. This triggered a cascade of fixes that spanned more modules that just those two. In fact, some warnings led us to understand that some of the functions were not used at all:

Removals
--------

- Remove add_wei from EVM.AccountRepo. It is not called from the evm app and thus does not need to be part of the behaviour. It remains as part of the `Blockchain.Account.Repo`.

- Remove AccountRepo.new_contract_address callback. It is not used.

- Remove unused AccountRepo.message_call callback. It is not used.

Other changes
-------------

- Update correct typespec from EVM.state() to Trie.t() in several places in `Account.Repo`.

- Correct the return type of EVM.AccountRepo.transfer

- We expose better types from `Account.Repo.Cache` to be used by `Account.Repo`.

- We address several incorrect types in `Account.Repo.Cache` that triggered several dialyzer warnings in `Account.Repo`. the Cache's types are fairly inaccurate in their representation of all the `nil` possibilities. Based on how the cache is used from the `Account.Repo` module, we make some updates that more accurately represent reality. For example, a function in `Account.Repo` calls the caching to update updating an account with `{nil, nil}` in order to "delete" the account from cache.

This behavior seems oddly intentional based on a failing test suggesting that we should not retrieve the account which has been deleted from storage since that would ignore the fact that it was deleted (but not yet persisted).

We should find a better way to handle that rather than having to do a manual insert of `{nil, nil}` to signify "this account has been deleted in cache, so do not look for it in storage".

### DIFF
--- a/.dialyzer.ignore-warnings
+++ b/.dialyzer.ignore-warnings
@@ -16,7 +16,7 @@ apps/evm/lib/evm/builtin.ex:255: Guard test __@3::'true' =:= 'false' can never s
 apps/evm/lib/evm/builtin.ex:255: Guard test __@1::'true' =:= 'false' can never succeed
 
 -------------------------------
-# TODO: rewrite mocked protocol typespecs
+# Machine code
 -------------------------------
 lib/evm/machine_code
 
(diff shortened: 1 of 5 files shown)


.dialyzer.ignore-warnings
```diff
@@ -86,6 +86,8 @@ tate_root':=_, _=>_}, _=>_}
 # using erlang 21 while we are using erlang 20. But we cannot yet update to
 # erlang 21 because we need rox to update
 apps/blockchain/lib/blockchain/block.ex:739: Function create_receipt/6 will never be called
+apps/blockchain/lib/blockchain/account/repo.ex:47: Function commit/1 has no local return
+apps/blockchain/lib/blockchain/account/repo.ex:50: The call 'Elixir.Blockchain.Account.Repo':new(#{<<_:160>>=>#{'balance':=integer(), 'code':=binary(), 'nonce':=integer(), 'storage':=#{'__struct__':='Elixir.MerklePatriciaTree.Trie', 'db':={atom(),_}, 'root_hash':=binary()}}}) will never return since it differs in the 1st argument from the success typing arguments: (#{'__struct__':='Elixir.MerklePatriciaTree.Trie', 'db':={atom(),_}, 'root_hash':=binary()})
```
germsvel: These are the last two warnings that I did not tackle because they seem to be going deeper into other modules. I will see if they have a quick fix, but since the issue wasn't apparent to me, I decided to leave them here for now

PR #497, opened 2018-10-12
fix writing to memory

ayrat555:
fixes https://github.com/poanetwork/mana/issues/491

Changes:
- We weren't considering `out_size` parameter when we were writing to the memory after message call. If it is smaller than output's size than we should truncate output
- Truncating logic in `Memory` module was wrong

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,13 +129,13 @@ Ethereum common tests are created for all clients to test against. We plan to pr
   - [GeneralStateTests](https://github.com/ethereum/tests/tree/develop/GeneralStateTests) 1026/1026 = 100% passing
 - [x] Homestead
   - [BlockchainTests](https://github.com/ethereum/tests/tree/develop/BlockchainTests) (Includes GeneralStateTests) 2231/2231 = 100% passing
-  - [GeneralStateTests](https://github.com/ethereum/tests/tree/develop/GeneralStateTests) 2059/2061 = 99.9% passing
+  - [GeneralStateTests](https://github.com/ethereum/tests/tree/develop/GeneralStateTests) 2061/2061 = 100.0% passing
 - [x] EIP150
   - [BlockchainTests](https://github.com/ethereum/tests/tree/develop/BlockchainTests) (Includes GeneralStateTests) 1275/1275 = 100% passing
-  - [GeneralStateTests](https://github.com/ethereum/tests/tree/develop/GeneralStateTests) 1109/1113 = 99.6% passing
+  - [GeneralStateTests](https://github.com/ethereum/tests/tree/develop/GeneralStateTests) 1113/1113 = 100.0% passing
 - [x] EIP158
   - [BlockchainTests](https://github.com/ethereum/tests/tree/develop/BlockchainTests) (Includes GeneralStateTests) 1232/1233 = 99.9% passing
-  - [GeneralStateTests](https://github.com/ethereum/tests/tree/develop/GeneralStateTests) 1176/1181 = 99.6% passing
+  - [GeneralStateTests](https://github.com/ethereum/tests/tree/develop/GeneralStateTests) 1180/1181 = 99.9% passing
 - [x] Byzantium
   - [BlockchainTests](https://github.com/ethereum/tests/tree/develop/BlockchainTests) (Includes GeneralStateTests) 4915/4959 = 99.1% passing
   - [GeneralStateTests](https://github.com/ethereum/tests/tree/develop/GeneralStateTests) 4758/4784 = 99.5% passing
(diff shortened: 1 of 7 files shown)


apps/evm/lib/evm/message_call.ex
```diff
@@ -199,7 +199,7 @@ defmodule EVM.MessageCall do
           %{machine_state | last_return_data: output}
 
         true ->
-          machine_state = Memory.write(machine_state, out_offset, output)
+          machine_state = Memory.write(machine_state, out_offset, output, out_size)
```
germsvel: Is there a test that we could add for this? It seems like this was an important piece that was overlooked, and in terms of making sure we don't re-introduce this bug in the future, it would be nice to have a test that covers this. What do you think?

apps/evm/test/evm/memory_test.exs
```diff
@@ -3,9 +3,31 @@ defmodule EVM.MemoryTest do
   use ExUnit.Case, async: true
   doctest EVM.Memory
 
-  test "returns the machine_state unchanged if data size is zero" do
-    machine_state = %MachineState{}
-    updated_machine_state = Memory.write(machine_state, 256, <<>>, 0)
-    assert machine_state == updated_machine_state
+  describe "write/4" do
```
germsvel: 👍 I like these tests!

apps/evm/lib/evm/memory.ex
```diff
@@ -79,13 +79,17 @@ defmodule EVM.Memory do
...
   def write(machine_state = %MachineState{}, offset_bytes, original_data, size) do
     data =
-      if size do
-        original_data
-        |> :binary.decode_unsigned()
-        |> rem(size * EVM.word_size())
-        |> :binary.encode_unsigned()
-      else
-        original_data
+      cond do
```
masonforest: This is more clear what's going on here too 👍

apps/evm/lib/evm/message_call.ex
ayrat555: @germsvel done

apps/blockchain/test/blockchain/state_test.exs
```diff
@@ -58,16 +55,12 @@ defmodule Blockchain.StateTest do
...
-      "stRevertTest/RevertOpcodeMultipleSubCalls"
+      "stCreate2/returndatasize_following_successful_create"
     ],
     "Frontier" => [],
-    "Homestead" => ["stRevertTest/RevertOpcodeMultipleSubCalls"],
-    "SpuriousDragon" => [
-      "stRevertTest/RevertOpcodeMultipleSubCalls",
-      "stSpecialTest/failed_tx_xcf416c53"
-    ],
-    "TangerineWhistle" => ["stRevertTest/RevertOpcodeMultipleSubCalls"]
```
germsvel: Yes! Opcode multiple subcalls 🔥

apps/evm/test/evm/message_call_test.exs
```diff
@@ -142,4 +143,37 @@ defmodule EVM.MessageCallTest do
...
   end
+
+  test "truncates message call's output if output's byte size is bigger than output_params" do
+    code = build(:machine_code, operations: [:push1, 1, :push1, 3, :add, :push1, 0x00, :mstore])
+
+    recipient = %{address: <<0x9::160>>, code: code}
+
+    account_interface =
+      build(:mock_account_interface)
+      |> MockAccountInterface.add_account(recipient.address, %{balance: 10, code: recipient.code})
```
germsvel: Is the `AcountInterface` stuff still in? Did I fail to remove it?

apps/evm/lib/evm/operation/stack_memory_storage_and_flow.ex
```diff
@@ -65,7 +65,12 @@ defmodule EVM.Operation.StackMemoryStorageAndFlow do
   """
   @spec mstore8(Operation.stack_args(), Operation.vm_map()) :: Operation.op_result()
   def mstore8([offset, value], %{machine_state: machine_state}) do
-    machine_state = Memory.write(machine_state, offset, value, EVM.byte_size())
+    byte_value =
+      value
+      |> rem(8 * EVM.word_size())
+      |> :binary.encode_unsigned()
```
germsvel: This was added since the last time I reviewed this. Just curious, did this fix more things?

README.md
```diff
@@ -129,13 +129,13 @@ Ethereum common tests are created for all clients to test against. We plan to pr
   - [GeneralStateTests](https://github.com/ethereum/tests/tree/develop/GeneralStateTests) 1026/1026 = 100% passing
 - [x] Homestead
   - [BlockchainTests](https://github.com/ethereum/tests/tree/develop/BlockchainTests) (Includes GeneralStateTests) 2231/2231 = 100% passing
-  - [GeneralStateTests](https://github.com/ethereum/tests/tree/develop/GeneralStateTests) 2059/2061 = 99.9% passing
+  - [GeneralStateTests](https://github.com/ethereum/tests/tree/develop/GeneralStateTests) 2061/2061 = 100.0% passing
```
germsvel: 🎉

apps/evm/lib/evm/operation/stack_memory_storage_and_flow.ex
ayrat555: @germsvel this logic was moved from `Memory.write`. It's needed for getting one byte from value

Issue #506, opened 2018-10-16
Fix consensus bug at block #544113 on ropsten


ayrat555:
it fails with `:gas_used_mismatch, :state_root_mismatch`

ayrat555:
fixed by https://github.com/poanetwork/mana/pull/507

PR #507, opened 2018-10-16
fix substate deduplication

ayrat555:
While working on https://github.com/poanetwork/mana/issues/506 it was discovered that substate deduplication works wrong

We used `Enum.dedup` for `selfdestruct_list` and `touched_accounts`
deduplication. It's wrong because it deduplicates only consecutive
values. MapSet has required functionality for deduplication

### DIFF
--- a/apps/evm/lib/evm/sub_state.ex
+++ b/apps/evm/lib/evm/sub_state.ex
@@ -13,12 +13,12 @@ defmodule EVM.SubState do
     Refunds
   }
 
-  defstruct selfdestruct_list: [],
-            touched_accounts: [],
+  defstruct selfdestruct_list: MapSet.new(),
+            touched_accounts: MapSet.new(),
             logs: [],
             refund: 0
 
-  @type address_list :: [EVM.address()]
+  @type address_list :: MapSet.t()
   @type logs :: [LogEntry.t()]
   @type refund :: EVM.Wei.t()
 
@@ -37,14 +37,14 @@ defmodule EVM.SubState do
   @doc """
   Checks whether the given `sub_state` is empty.
   """
-  def empty?(sub_state), do: %{sub_state | touched_accounts: []} == empty()
+  def empty?(sub_state), do: %{sub_state | touched_accounts: MapSet.new()} == empty()
 
   @doc """
   Adds log entry to substate's log entry list.
 
   ## Examples
 
-      iex> sub_state = %EVM.SubState{selfdestruct_list: [], logs: [], refund: 0}
+      iex> sub_state = %EVM.SubState{}
       iex> sub_state |> EVM.SubState.add_log(0, [1, 10, 12], "adsfa")
       %EVM.SubState{
         logs: [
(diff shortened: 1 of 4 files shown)


apps/evm/lib/evm/sub_state.ex
```diff
@@ -100,28 +98,27 @@ defmodule EVM.SubState do
...
 
     refund =
       sub_state1.refund + sub_state2.refund -
-        (Enum.count(selfdestruct_list) - Enum.count(dedup_selfdestruct_list)) *
-          Refunds.selfdestruct_refund()
+        (Enum.count(sub_state1.selfdestruct_list) + Enum.count(sub_state2.selfdestruct_list) -
+           Enum.count(selfdestruct_list)) * Refunds.selfdestruct_refund()
 
-    touched_accounts = Enum.dedup(sub_state1.touched_accounts ++ sub_state2.touched_accounts)
+    touched_accounts = MapSet.union(sub_state1.touched_accounts, sub_state2.touched_accounts)
```
germsvel: 👍

apps/evm/lib/evm/sub_state.ex
```diff
@@ -100,28 +98,27 @@ defmodule EVM.SubState do
...
-    dedup_selfdestruct_list = Enum.dedup(selfdestruct_list)
+    selfdestruct_list = MapSet.union(sub_state1.selfdestruct_list, sub_state2.selfdestruct_list)
     logs = sub_state1.logs ++ sub_state2.logs
 
     refund =
       sub_state1.refund + sub_state2.refund -
-        (Enum.count(selfdestruct_list) - Enum.count(dedup_selfdestruct_list)) *
-          Refunds.selfdestruct_refund()
+        (Enum.count(sub_state1.selfdestruct_list) + Enum.count(sub_state2.selfdestruct_list) -
+           Enum.count(selfdestruct_list)) * Refunds.selfdestruct_refund()
```
germsvel: I don't know the specification for how we calculate refunds, but this seems a bit odd to me. I'm guessing we should be doing the following: 

> Given sets `a` and `b`, the combined refund should be `refund a` + `refund b` - `refund of duplicate values`

Is that right? 

But we seem to be doing, 
```
a.refund + b.refund - ((a.count + b.count - (a union b).count) * refund_amount)
```
won't ` a.count + b.count` be the same as `(a union b).count`, making that `0 * refund_amount`? 

I think we might just want the _intersection_, right? 

```
a.refund + b.refund - ((a intersection b).count) * refund_amount)
```

ayrat555: `a union b`  has only unique values from a and b. so `a.count + b.count != (a union b).count`. Your variant seem clearer so I will use it

ayrat555: @germsvel done

apps/evm/lib/evm/sub_state.ex
```diff
@@ -100,28 +98,30 @@ defmodule EVM.SubState do
...
-    selfdestruct_list = sub_state1.selfdestruct_list ++ sub_state2.selfdestruct_list
-    dedup_selfdestruct_list = Enum.dedup(selfdestruct_list)
+    selfdestruct_list = MapSet.union(sub_state1.selfdestruct_list, sub_state2.selfdestruct_list)
     logs = sub_state1.logs ++ sub_state2.logs
 
-    refund =
-      sub_state1.refund + sub_state2.refund -
-        (Enum.count(selfdestruct_list) - Enum.count(dedup_selfdestruct_list)) *
-          Refunds.selfdestruct_refund()
+    common_refund =
```
germsvel: 👍 good name change

PR #508, opened 2018-10-16
Add Blockchain.Account.Inspector

germsvel:
Description
===========

Adds a `Blockchain.Account.Inspector` module to more easily inspect the contents of an account. This is mostly for debugging purposes, and it is not expected that it would be used in production. It takes an account address (hex) and prints out the balance, nonce, storage root, and code hash.

### DIFF
--- a/apps/blockchain/lib/blockchain/account/inspector.ex
+++ b/apps/blockchain/lib/blockchain/account/inspector.ex
@@ -1,0 +1,41 @@
+defmodule Blockchain.Account.Inspector do
+  @moduledoc """
+  Module to inspect account information
+  """
+
+  alias Blockchain.Account
+
+  @doc """
+  Prints a visual representation of several accounts
+  """
+  def inspect_accounts(state, account_addresses) do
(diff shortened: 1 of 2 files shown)


apps/blockchain/lib/blockchain/account/inspector.ex
```diff
@@ -0,0 +1,41 @@
...
+  Prints a visual representation of several accounts
+  """
+  def inspect_accounts(state, account_addresses) do
+    Enum.map(account_addresses, &inspect_account(state, &1))
+  end
+
+  @doc """
+  Prints a visual representation of an account
+  """
+  def inspect_account(state, account_address) do
```
ayrat555: will it be used only with a committed state?

germsvel: Yeah. We could improve this as we go. I was just extracting this because I thought it might be helpful as tooling since we're all debugging issues with blocks. I've been using this for what I'm currently debugging.

Issue #370, opened 2018-08-21
Fix "GeneralStateTests/stSpecialTest/failed_tx_xcf416c53_d0g0v0.json" test


germsvel:
There are blockchain and state test counterparts to this. One fix should fix both of them.

PR #509, opened 2018-10-16
Touch RIPEMD account on failure

germsvel:
Relates to https://github.com/poanetwork/mana/issues/370

This fixes GeneralStateTests/stSpecialTest/failed_tx_xcf416c53 in state tests for SpuriousDragon and Byzantium, and it fixes it in blockchain tests for SpuriousDragon. Strangely, it does not solve it for Byzantium. I will continue working on a Byzantium fix, but I think this can be committed now as a unit of work.

Description
===========

During the empty account cleanup that ensued after the Spurious Dragon hardfork, a consensus issue arose. Geth was not reverting removed empty accounts (un-deleting them as it where) when a call was OOG. Parity was reverting the removal.

During cleanup, it was discovered that there was one more idiosyncrasy: both Geth and Parity had failed to revert the clean up of an empty account when an inner call was out-of gas. The account in question was account 3, the precompiled contract RIPEMD.

Since both Geth and Parity had removed the account and had not reverted that, it was deemed that it should remain that way. So the `failed_tx_xcf416c53` test is testing that very thing. It's testing transaction [0xcf416c536ec1a19ed1fb89e4ec7ffb3cf73aa413b3aa9b77d60e4fd81a4296ba](https://etherscan.io/tx/0xcf416c536ec1a19ed1fb89e4ec7ffb3cf73aa413b3aa9b77d60e4fd81a4296ba#internal)

What's our fix?
==============

Since we implicitly revert on a call failure by simply not returning a new sub state or exec env, we modify the call failure to return a new sub state with the account touched if it is the RIPEMD contract. It is an ugly fix for this, but I am unsure as to how else to handle it without changing the logic of how we revert state on a call failure.

More details
============

If more details are needed, please see:

- The announcement of the consensus bug: https://blog.ethereum.org/2016/11/25/security-alert-11242016-consensus-bug-geth-v1-4-19-v1-5-2/

In there you can see,

> Details: Geth was failing to revert empty account deletions when the
transaction causing the deletions of empty accounts ended with an out-of-gas
exception. An additional issue was found in Parity, where the Parity client
incorrectly failed to revert empty account deletions in a more limited set of
contexts involving out-of-gas calls to precompiled contracts; the new Geth
behavior matches Parity’s, and empty accounts will cease to be a source of
concern in general in about one week once the state clearing process finishes.

- This helpful comment by Holiman in EIP 716: https://github.com/ethereum/EIPs/issues/716#issuecomment-330824096

- Geth introducing fix: https://github.com/ethereum/go-ethereum/commit/12d654a6fc4580f9194a931032ebf0e1b1927279#diff-2433aa143ee4772026454b8abd76b9ddR93

- Aleth implementation of fix: https://github.com/ethereum/aleth/blob/8daa6e1c56e874e88e894d96f89f1ff8c9412a63/libethereum/Executive.cpp#L314

### DIFF
--- a/apps/evm/lib/evm/message_call.ex
+++ b/apps/evm/lib/evm/message_call.ex
@@ -161,7 +161,7 @@ defmodule EVM.MessageCall do
        ) do
     result = failed_call(message_call)
 
-    if message_call_to_rip_md?(message_call) do
+    if message_call_to_rip_emd?(message_call) do
       sub_state =
         message_call.current_sub_state
         |> SubState.add_touched_account(message_call.recipient)
@@ -172,8 +172,8 @@ defmodule EVM.MessageCall do
     end
   end
 
-  defp message_call_to_rip_md?(message_call) do
-    message_call.recipient == <<3::160>>
+  defp message_call_to_rip_emd?(message_call) do
+    message_call.recipient == EVM.Builtin.Rip160.contract_address()
   end
 
   defp finalize_reverted_message_call(
--- a/apps/evm/lib/evm/message_call.ex
+++ b/apps/evm/lib/evm/message_call.ex
@@ -2,7 +2,7 @@ defmodule EVM.MessageCall do
   alias EVM.{AccountRepo, ExecEnv, Memory, Builtin, VM, Functions, MachineState, SubState}
 
   @moduledoc """
-  Describes a message call function that used for all call operations (call, delegatecall, callcode, staticcall).
+  Describes a message call function that is used for all call operations (call, delegatecall, callcode, staticcall).
   """
   defstruct [
     :current_exec_env,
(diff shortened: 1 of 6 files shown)


ayrat555:
What do you think about updating tests status in README?

apps/evm/lib/evm/message_call.ex
```diff
@@ -155,6 +155,27 @@ defmodule EVM.MessageCall do
...
+        |> SubState.add_touched_account(message_call.recipient)
+
+      Map.put(result, :sub_state, sub_state)
+    else
+      result
+    end
+  end
+
+  defp message_call_to_rip_md?(message_call) do
+    message_call.recipient == <<3::160>>
```
ayrat555: can we move `<<3::160>>` to module attribute or maybe to the module with rip builtin?

germsvel: Yeah, very good idea.

germsvel: Updated in 1150407669ec07803c7d8615524f2d4a9a9c62ec

PR #516, opened 2018-10-17
Derive current block from tree in sync_with_infura

germsvel:
Description
===========

We have been saving both a "current_block_hash" and "current_block_tree" in the database every 1000 blocks. The problem comes when a failure occurs at a block that is a multiple of 1000. In that case, it appears as though we have a corrupted db, but in reality all that has happened is that we saved a "current_block_hash" too early (before actually validating the block), and next time we search for that block in our database, it cannot be found.

So we do two things here to improve that,

1. We derive the current block hash from the current block tree. Since the tree has the best block in it, that is all we need.

2. We move the saving of the current block tree to after we have verified and added the block (instead of before). That way, when we try to run the script again, we will have the correct version of the tree saved.

### DIFF
--- a/apps/blockchain/scripts/sync_with_infura.ex
+++ b/apps/blockchain/scripts/sync_with_infura.ex
@@ -14,19 +14,20 @@ defmodule SyncWithInfura do
   def add_block_to_tree(db, chain, tree, n) do
     next_block = get_block(n)
 
-    if rem(n, @save_block_interval) == 0 do
-      Logger.info("Saved progress at block #{n}")
-      MerklePatriciaTree.DB.put!(db, "current_block_hash", Blockchain.Block.hash(next_block))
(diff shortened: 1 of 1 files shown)


apps/blockchain/scripts/sync_with_infura.ex
```diff
@@ -153,4 +144,14 @@ tree =
       Blockchain.Blocktree.new_tree()
   end
 
+current_block =
```
masonforest: If you wanna go one step further:

It looks like we're only using `current_block` to get the block number on line`157`

So I think we could actually do:

    current_block_number = if is_nil(tree.best_block), do: 0, else: tree.best_block.header.number
    SyncWithInfura.add_block_to_tree(db, chain, tree, current_block_number)

germsvel: Oh yeah, that is a good. I'll add that to my dao fork changes

PR #518, opened 2018-10-18
fix output in revert during create

ayrat555:
We should return output from revert during create and save it to last_return_data

this pr should be merged after https://github.com/poanetwork/mana/pull/517


apps/blockchain/test/blockchain/contract/create_contract_test.exs
```diff
@@ -90,7 +90,7 @@ defmodule Blockchain.Contract.CreateContractTest do
         block_header: %Block.Header{nonce: 1}
       }
 
-      {:error, {account_repo, 0, sub_state}} = Contract.create(params)
+      {:error, {account_repo, 0, sub_state, _ourput}} = Contract.create(params)
```
masonforest:
```suggestion
      {:error, {account_repo, 0, sub_state, _output}} = Contract.create(params)
```

PR #521, opened 2018-10-19
fix Byzantium blockchain tests

ayrat555:
We weren't running Byzantium blockchain tests because `failing_tests` list contained an empty string (`" "`)

That' why it wasn't failing with `receipt_root_mismatch` without https://github.com/poanetwork/mana/pull/520

This should be merged after https://github.com/poanetwork/mana/pull/520. Because now tests are failing with `receipts_root_mismatch`


apps/blockchain/test/blockchain_test.exs
```diff
@@ -18,8 +18,14 @@ defmodule BlockchainTest do
     "Homestead" => [],
     "TangerineWhistle" => [],
     "SpuriousDragon" => [],
-    "Byzantium" => String.split(@failing_byzantium_tests, "\n"),
-    "Constantinople" => String.split(@failing_constantinople_tests, "\n"),
+    "Byzantium" =>
+      @failing_byzantium_tests
+      |> String.split("\n")
+      |> Enum.reject(fn path -> path == "" || path == " " end),
```
masonforest: I feel like this could be cleaner. Could we update the script that writes this file not to write a newline instead? Or can we go back to listing the failing tests inline?

ayrat555: @masonforest  good idea. We can rid of external files because not many failing tests left

PR #530, opened 2018-10-23
Update license to dual MIT / Apache 2.0

andogro:
The provisional GNU GPL license is too strong for this library and should be replaced with a more permissive one. MIT/Apache-2.0 has been chosen as a suitable candidate.


andogro:
All current contributors have agreed to this change. The software was originally published under MIT and previous contributions not association with current contributors were made under the MIT license.

PR #531, opened 2018-10-23
Improve EVM.Logger output

germsvel:
As I work on debugging why we're failing to sync the mainnet, I have found these changes to be helpful. I am unsure whether or not this is something we want to use, since I know the EVM.Logger's original intent was to compare the output with parity. But I have found that most of my debugging is done against Geth's stacktraces in [Etherescan](https://etherscan.io). A sample stack trace can be found [here](https://etherscan.io/vmtrace?txhash=0xec32243bad69039271ae01362ec4916be9fbf546e5356db420b1267ac1a038a3#raw).

Description
===========

A lot of our debugging is happening as we compare stack traces against geth's vm traces found in Etherescan.

To that end, we make the following changes so that debugging is easier:

1. Add a `step` key to `MachineState` so that we know how many cycles the vm has executed. This is very helpful in comparing vm traces line-by-line against geth, since they keep a step count. We increment the step at the end of each `VM.cycle`.

2. Log the program counter as a number. For some reason, we were logging the program counter as hex. Etherscan shows it as a number, and it is much easier to reason about.

3. Log stack as hex. Etherescan shows the stack as hex numbers, not as integers. Comparing the stack against that will be much easier and will no longer require manual translation of values.

4. We update the EVM readme to suggest placing the EVM.Logger calls closer to the beginning of the `cycle` function so that values more closely match Etherscan's values (e.g. gas used would be equivalent to next cycle's gas in Etherscan if it's printed at the end of the cycle).

Comparisons
==========

Before
-------

screen shot 2018-10-23 at 10 33 48 am

After
------

screen shot 2018-10-23 at 10 24 42 am

### DIFF
--- a/apps/evm/lib/evm/logger.ex
+++ b/apps/evm/lib/evm/logger.ex
@@ -7,7 +7,11 @@ defmodule EVM.Logger do
   """
   @spec log_stack(MachineState.t()) :: MachineState.t()
   def log_stack(machine_state) do
-    Logger.debug(fn -> "Stack: #{inspect(machine_state.stack)}" end)
+    stack =
+      machine_state.stack
+      |> Enum.map(&stack_value_string/1)
+
+    Logger.debug(fn -> "Stack: #{inspect(stack)}" end)
     machine_state
   end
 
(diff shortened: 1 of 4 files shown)


apps/evm/lib/evm/logger.ex
```diff
@@ -39,42 +43,42 @@ defmodule EVM.Logger do
...
-
-        Logger.debug(fn -> "       | #{i}: 0x#{value_string}" end)
+        value_string = stack_value_string(value)
+        Logger.debug(fn -> "       | #{i}: #{value_string}" end)
       end)
     end
   end
 
+  defp current_step(machine_state) do
+    machine_state.step + 1
```
masonforest: Nitpick: adding 1 here feels a bit odd. Can we move the call to `log_opcode_and_gas_left` after the step is incremented? Or maybe initialize step to 1?

germsvel: Love the nitpick because it felt very weird. I'll change this.

PR #485, opened 2018-10-09
Update Ethereum common tests

MattMSumner:
This updates the ethereum common tests. Also updates the readme with the passing test numbers.

PR #533, opened 2018-10-24
turn on eip1283

ayrat555:
eip1283 (sstore gas cost changes) was disabled because common tests
weren't ready

Passing Constantinople state tests 5309/5311= 100.0%
Passing Constantinople blockchain tests 5347/5349 = 100.0%

### DIFF
--- a/apps/evm/lib/evm/configuration.ex
+++ b/apps/evm/lib/evm/configuration.ex
@@ -115,8 +115,8 @@ defmodule EVM.Configuration do
       "Byzantium" ->
         EVM.Configuration.Byzantium.new()
 
-      "Constantinople" ->
-        EVM.Configuration.Constantinople.new()
+      # "Constantinople" ->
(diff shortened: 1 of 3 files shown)


apps/evm/lib/evm/configuration.ex
```diff
@@ -115,8 +115,8 @@ defmodule EVM.Configuration do
       "Byzantium" ->
         EVM.Configuration.Byzantium.new()
 
-      "Constantinople" ->
-        EVM.Configuration.Constantinople.new()
+      # "Constantinople" ->
```
MattMSumner: Should this be commented out?

ayrat555: @MattMSumner  currently, Constantinople tests are passing on the latest `develop` common tests branch. It should be commented out in your PR https://github.com/poanetwork/mana/pull/485

PR #535, opened 2018-10-24
Simplify extcodecopy conditional logic

germsvel:
Resolves https://github.com/poanetwork/mana/issues/528

Description
============

We were failing to sync past block 2325632 on the mainnet because transaction `0xec32243bad69039271ae01362ec4916be9fbf546e5356db420b1267ac1a038a3` has an `extcodecopy` operation with positive `size` but the code is empty.

In order to fix that, we remove the conditionals on `extcodecopy` that returned an unchanged `machine_state`. That conditional tested three possibilities with logical ORs, meaning any of these three would return an unchanged `machine_state`:

1. size == 0 -> this is already handled by the Memory.write operation, so we remove the condition.

2. size + mem_offset > EVM.max_int() -> reading the yellow paper, it seems that this condition was wrong. Under `excodecopy` in the appendices outlining Environmental Information, it says:

> The additions in μs[2] + i are not subject to the 2^256 modulo.

3. (code_offset == 0 && account_code == "") -> this condition was the one failing the transaction mentioned above because even though the code offset was `0` and the account code was empty, the size was still positive, meaning we actually needed to write into memory. We add a test for this case to prevent future regression of this condition.

### DIFF
--- a/apps/evm/test/evm/operation/environmental_information_test.exs
+++ b/apps/evm/test/evm/operation/environmental_information_test.exs
@@ -52,6 +52,27 @@ defmodule EVM.Operation.EnvironmentalInformationTest do
   end
 
   describe "extcodecopy" do
+    test "writes empty code to memory if size is positive" do
+      address = <<1::160>>
+      mem_offset = code_offset = 0
+      size = 120
+      bit_size = size * 8
+      code = <<>>
+
+      params = [address, mem_offset, code_offset, size]
+
+      exec_env = %EVM.ExecEnv{
+        account_repo: %EVM.Mock.MockAccountRepo{account_map: %{address => %{code: code}}}
+      }
+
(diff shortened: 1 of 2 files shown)


apps/evm/test/evm/operation/environmental_information_test.exs
```diff
@@ -52,6 +52,27 @@ defmodule EVM.Operation.EnvironmentalInformationTest do
   end
 
   describe "extcodecopy" do
+    test "writes empty code to memory if size is positive" do
+      address = <<1::160>>
+      mem_offset = code_offset = 0
+      size = 120
+      bit_size = 120 * 8
```
MattMSumner: How would you feel about chaining this to:
```suggestion
      bit_size = size * 8
```

PR #536, opened 2018-10-24
Make EVM.Configuration simpler

germsvel:
This is a suggested PR. If it doesn't seem like a good idea, I'd be happy to close this.

The problem
============

The EVM.Configuration files are full of repetitive implementations of functions that simply call the underlying key in a struct,

```elixir
def contract_creation_cost(config), do: config.contract_creation_cost
```

or functions that use a fallback config to call another module's
implementation,

```elixir
def extcodesize_cost(config), do: Frontier.extcodesize_cost(config.fallback_config)
```

This is a lot of duplication both in the functions we write but also in the calling code. We are doing the following a lot,

```elixir
Evm.Configuration.for(config).contrac_creation_cost(config)
```

when we could simply be doing,

```elixir
config.contract_creation_cost
```

In addition, it is currently difficult to understand the full configuration for a hardfork as a single snapshot, since a configuration like Byzantium can have SpuriousDragon's configuration nested within it, and it have TangerineWhistle nested within it, and it have Homestead, and it have Frontier:

```elixir
%Configuration.Byzantium{
 # fields
 fallback_config: %SpuriousDragon{
   # fields
   fallback_config: %TangerineWhistle{
     # fields
     fallback_config: %Homestead{
       # fields
       fallback_config: %Frontier{
         # fields
       }
     }
   }
  }
}
```

The solution
============

Since most of the configuration options are just static values, like boolean flags or some integer, we can simply use a "callback config" at the moment we are generating the structs. So for example, the `Homestead` `defstruct` definition can call `Frontier.new()` and override the values needed to generate its own struct.

This has the benefit of removing repetition in function definitions (with two exceptions), it improves ease of legibility of a configuration since we only have a single struct instead of having structs within structs within structs, and it also opens the possibility of future configurations that a user may want to define which do not match any single hardfork.

Note: with this solution we do lose the ability to have functions with `?` to help us understand whether or not the configuration returns a boolean. But with `has_` and `should_` prefixes to the configuration options to indicate that, I think we can make that tradeoff.

### DIFF
--- a/apps/blockchain/lib/blockchain/contract/create_contract.ex
+++ b/apps/blockchain/lib/blockchain/contract/create_contract.ex
@@ -83,7 +83,7 @@ defmodule Blockchain.Contract.CreateContract do
           EVM.address()
         ) :: Repo.t()
   defp increment_nonce_of_touched_account(account_repo, config, address) do
-    if EVM.Configuration.for(config).increment_nonce_on_create?(config) do
+    if config.increment_nonce_on_create do
       Repo.increment_account_nonce(account_repo, address)
     else
       account_repo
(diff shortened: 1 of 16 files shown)


apps/blockchain/lib/blockchain/contract/create_contract.ex
```diff
@@ -83,7 +83,7 @@ defmodule Blockchain.Contract.CreateContract do
           EVM.address()
         ) :: Repo.t()
   defp increment_nonce_of_touched_account(account_repo, config, address) do
-    if EVM.Configuration.for(config).increment_nonce_on_create?(config) do
```
ayrat555: This looked weird to me from the beginning. The new version looks much cleaner

MattMSumner:
👍

Issue #538, opened 2018-10-25
Improve working with cache


ayrat555:
1. `current_value` in storage cache is not used and always read from Merkle storage
2. Try to implement multiple levels of cache 
     - Currently, we commit changes after every transaction
     - I think it's possible to have Merkle Patricia storage cache and commit changes on block finalisation

PR #542, opened 2018-10-26
Mekle Patricia Trie Cache

ayrat555:
We need at least two levels of cache. Right now we have one level that was implemented in https://github.com/poanetwork/mana/pull/433, https://github.com/poanetwork/mana/pull/441, https://github.com/poanetwork/mana/pull/458. 

Now we're committing changes after every transaction. This is ok if there are no invalid blocks. But in real network we will be receiving invalid blocks and we shouldn't write their transaction data to db. 

The idea is to have `CachingTrie`: all storage modification will be made to in-memory trie and saved to disk after successful block validation. Nodes that are not present in in-memory trie will be read from disk.
This will help to validate state root without saving storage modifications to disk.

This PR implements `CachingTrie`. I will replace regular `Trie` with `CachingTrie` in the next PR

Part of https://github.com/poanetwork/mana/issues/538

### DIFF
--- a/apps/merkle_patricia_tree/lib/merkle_patricia_tree/storage_behaviour.ex
+++ b/apps/merkle_patricia_tree/lib/merkle_patricia_tree/storage_behaviour.ex
@@ -1,0 +1,28 @@
+defmodule MerklePatriciaTree.StorageBehaviour do
+  @moduledoc """
+  Defines functions for fetching and updating nodes.
+  """
+
+  alias MerklePatriciaTree.Trie
+  alias MerklePatriciaTree.Trie.Node
+
+  @type t :: struct()
+
+  @callback fetch_node(t) :: Node.trie_node()
+
+  @callback put_node(node, t) :: nil | binary()
+
+  @callback remove_key(t, Trie.key()) :: t
+
+  @callback update_key(t(), Trie.key(), ExRLP.t() | nil) :: t
+
+  @callback get_key(t(), Trie.key()) :: t()
+
+  @callback into(binary(), t) :: t
+
+  @callback store(t) :: t
+
+  def storage(implementation) do
+    implementation.__struct__
+  end
+end
--- a/apps/merkle_patricia_tree/lib/merkle_patricia_tree/storage_behaviour.ex
+++ b/apps/merkle_patricia_tree/lib/merkle_patricia_tree/storage_behaviour.ex
@@ -12,11 +12,11 @@ defmodule MerklePatriciaTree.StorageBehaviour do
 
   @callback put_node(node, t) :: nil | binary()
 
-  @callback remove_key(t, Trie.key()) :: t
+  @callback remove_key(t, Trie.key()) :: Node.trie_node()
 
   @callback update_key(t(), Trie.key(), ExRLP.t() | nil) :: t
 
-  @callback get_key(t(), Trie.key()) :: t()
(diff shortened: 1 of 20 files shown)


apps/merkle_patricia_tree/lib/merkle_patricia_tree/storage_behaviour.ex
```diff
@@ -0,0 +1,28 @@
+defmodule MerklePatriciaTree.StorageBehaviour do
```
germsvel: What do you think about dropping the `Behaviour` and make it just `MerklePatriciaTree.Storage`? I don't think the specificity adds value here

apps/merkle_patricia_tree/lib/merkle_patricia_tree/storage_behaviour.ex
```diff
@@ -0,0 +1,28 @@
...
+
+  @callback remove_key(t, Trie.key()) :: t
+
+  @callback update_key(t(), Trie.key(), ExRLP.t() | nil) :: t
+
+  @callback get_key(t(), Trie.key()) :: t()
+
+  @callback into(binary(), t) :: t
+
+  @callback store(t) :: t
```
germsvel: Should this be a `@callback`? The cache implementation doesn't seem to implement this, so I would say that we should probably move this out of the behaviour and just let it be a function in the `Trie` module.

apps/merkle_patricia_tree/lib/merkle_patricia_tree/trie/helper.ex
```diff
@@ -40,4 +40,9 @@ defmodule MerklePatriciaTree.Trie.Helper do
   def get_binary(l) do
     for x <- l, into: <<>>, do: <<x::4>>
   end
+
+  # Encodes `x` in RLP if it isn't already encoded.
+  # And it is definitely not encoded if it is `<<>>` or not a binary (e.g. array).
+  def rlp_encode(x) when not is_binary(x) or x == <<>>, do: ExRLP.encode(x)
```
germsvel: `<<>>` is a binary, right? So is the `or x == <<>>` superfluous here?

apps/merkle_patricia_tree/lib/merkle_patricia_tree/storage_behaviour.ex
ayrat555: `CachingTrie` implements `store` https://github.com/mana-ethereum/mana/blob/9704ade786d567e34e4e703532d706a2ae6dedf7/apps/merkle_patricia_tree/lib/merkle_patricia_tree/caching_trie.ex#L89

germsvel: My bad. I missed that.

apps/merkle_patricia_tree/lib/merkle_patricia_tree/trie/helper.ex
ayrat555: Actually, `is_binary` does not match `<<>>`

germsvel: That's weird. I just tried it on `iex`, 

```elixir
iex> is_binary(<<>>)
true
```

ayrat555: you're right

PR #543, opened 2018-10-27
adding distillery as the release manager and ensuring the release starts

InoMurko:
- I've added Distillery dependency that we can use as the OTP release process.
- I've removed all Mix.env() runtime calls, because Mix isn't available in a OTP release, making the release unable to start.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -494,12 +494,6 @@ jobs:
       - run: mix local.hex --force
       - run: mix local.rebar --force
       - run: mix compile
-      - run:
-          name: Collecting artifacts
-          command: |
-            ls
-            ls rel/
-          when: always
       - run: mix release --verbose --env prod
       - run:
           name: Collecting artifacts
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -494,7 +494,7 @@ jobs:
       - run: echo 'export PATH=~/.cargo/bin:$PATH' >> $BASH_ENV
       - run: mix local.hex --force
       - run: mix local.rebar --force
-      - run: mix release --verbose
+      - run: MIX_ENV=dev mix release --verbose
       - store_artifacts:
           path: _build/dev/rel/mana/
           destination: _build/dev/rel/mana/
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -478,6 +478,23 @@ jobs:
 
       - run: mix credo --only readability,consistency,warning,refactor
 
+  release:
+    docker:
+      # Ensure .tool-versions matches
+      - image: circleci/elixir:1.7.2
+        environment:
+          MIX_ENV: dev
+
+    working_directory: ~/app
+
+    steps:
+      - attach_workspace:
+          at: .
+
+      - run: mix release
+
+      - run: mix credo --only readability,consistency,warning,refactor
+
 workflows:
   version: 2
   primary:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -489,10 +489,15 @@ jobs:
 
     steps:
       - attach_workspace:
-          at: .
-
-      - run: mix release
-
+           at: .
+      - run: curl https://sh.rustup.rs -sSf | sh -s -- -y
+      - run: echo 'export PATH=~/.cargo/bin:$PATH' >> $BASH_ENV
+      - run: mix local.hex --force
+      - run: mix local.rebar --force
+      - run: mix release --verbose
+      - store_artifacts:
+          path: _build/dev/rel/mana/
+          destination: _build/dev/rel/mana/
 
 workflows:
   version: 2
@@ -506,6 +511,8 @@ workflows:
       - dialyzer:
           requires: [build]
       - test:
+          requires: [build]
+      - release:
           requires: [build]
       - Frontier:
           requires: [build]
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -493,7 +493,6 @@ jobs:
 
       - run: mix release
 
-      - run: mix credo --only readability,consistency,warning,refactor
 
 workflows:
   version: 2
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -490,11 +490,12 @@ jobs:
     steps:
       - attach_workspace:
            at: .
-      - run: curl -s https://static.rust-lang.org/rustup.sh | sh -s -- --channel=nightly
+      - run: curl https://sh.rustup.rs -sSf | sh -s -- -y
       - run: echo 'export PATH=~/.cargo/bin:$PATH' >> $BASH_ENV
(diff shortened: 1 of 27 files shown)


rel/config.exs
```diff
@@ -0,0 +1,59 @@
...
+  # dev mode.
+  set dev_mode: true
+  set include_erts: true
+  set cookie: :"zGokf%I2ZvCe6)?qkO}rlf)[*&<GefmD}w*$5Gv/6VNO6`x&?Y!t^i>M[/zJ.q9j"
+end
+
+environment :prod do
+  set include_erts: true
+  set include_src: false
+  set cookie: :"%POy.W>cr1@.Qe217jz;d.m0Uhx]$h8Ae$3.i*,U;jH@]U^T}&sLkg12;,5dc)P5"
```
hayesgm: Should we load cookie from somewhere else or just not include it at all?

apps/ex_wire/lib/ex_wire/config.ex
```diff
@@ -115,6 +115,11 @@ defmodule ExWire.Config do
     get_env!(:node_discovery)
   end
 
+  @spec get_environment() :: atom() | ArgumentError
+  def get_environment() do
```
hayesgm: Why don't we just use `Mix.env()` here, since the configuration looks to be identical and it's less surprising. I assume you're preparing for the issue of having multiple "environments" under prod, such as rinkeby, etc? But I'm not sure if having a `Mix.env()` and `Config.get_environment()` would necessarily would even in that case. I think we're best sticking with `Mix.env()` here and adding `Config.get_environment()` when we have a stronger case to separate it out. Thoughts?

.circleci/config.yml
```diff
@@ -210,6 +210,22 @@ jobs:
...
+          MIX_ENV: dev
+
+    working_directory: ~/app
+
+    steps:
+      - attach_workspace:
+          at: .
+
+      - run: mix release
+
```
hayesgm: We should add the release tar.gz as an artifact so it's stored by CircleCI

apps/ex_wire/lib/ex_wire/config.ex
InoMurko: The problem with the Mix.* namespace is that it's not available in a OTP release. So if you have environment based conditionals, I don't know of other ways of doing this.

rel/config.exs
InoMurko: At this moment I think the purpose of having the release built in CI is that we're ensuring it doesn't get broken. Any tarball release that might deploy the tarball would want to overwrite the cookie anyway.

Setting it to something that doesn't look generated might make more sense.

InoMurko: In the longer run, having acceptance tests that would run against a _release_ would prove the release _working_ as well.

.circleci/config.yml
InoMurko: It would make sense to pack the artefacts + logs once we have acceptance test run against it.

InoMurko:
Sorting issues with RocksDB...

apps/ex_wire/lib/ex_wire/config.ex
hayesgm: So isn't it okay to use then if it gets baked in at compile that? Generating a production release will mean that it's always set to prod, which is what we would expect?

InoMurko: It gets evaluated only for .exs configuration files but not inside modules. 
You can also make dev releases that are quicker or custom (like staging).

hayesgm:
@InoMurko I was taking a look at this last night-- it's definitely having a problem compiling the `rox` dependency, breaking during a rust build of a C++ artifact? But, the bigger question is actually: it works fine building test (`MIX_ENV=test mix compile`)-- is it not building that artifact in test or is the environment different?

InoMurko:
Hey @hayesgm! Being at devcon I'm a little busy.
What happens is that rox wants to optimise the cargo build for a release. As per https://hexdocs.pm/rustler/basics.html and the mix.exs of the rox dependency:
`  defp rustler_crates do
    [
      rox_nif: [path: "native/rox_nif", cargo: :system, default_features: false, features: [], mode: :release],
    ]
  end`

What I think it happens is that the compilation eats the all available memory in the container executing the job. I've been aligning with @atoulme to increase the CI plan that would allow us to increase the memory size - it's only 4gb at the moment. That should solve it. I've tried compiling a release on a bigger linux instance and it worked.

InoMurko:
That's my current take (with limited time at devcon and the output being internal compiler error: Killed with exit code 4). Tried also rustc nightly build and it comes up the same.

InoMurko:
Managed to get this working by swapping the RocksDB dependency used with https://gitlab.com/barrel-db/erlang-rocksdb .

That pointed me to the fact that db handles are not being closed. Probably because the previous :rox dependency didn't support that. This awesome work from Benoit does, we should implement that in a later stage.

Please review again @hayesgm @masonforest @atoulme and others.

hayesgm:
Great work getting this building. Do you know if it's possible to build for multiple architectures? E.g. one for linux, one for macOS and one for Windows?

InoMurko:
I’m not 100% for windows. Everything else should work.

hayesgm:
On MacOS:

```bash
$ ./bin/mana
./erts-10.0.5/bin/erl: line 13: ./erts-10.0.5/bin/erlexec: cannot execute binary file
./erts-10.0.5/bin/erl: line 13: ./erts-10.0.5/bin/erlexec: Undefined error: 0
Unusable Erlang runtime system! This is likely due to being compiled for another system than the host is running
```

InoMurko:
Oh I wasn’t sure what you meant. You took the release artifact from Circle? You would have to build it separately. Afaik, circle allows docker, mac but not windows.

hayesgm:
Yeah, it'd be cool if we could build for macOS, for instance, in a docker container. But for now, this is great and we should merge to master!

.credo.exs
```diff
@@ -87,7 +87,7 @@
         {Credo.Check.Readability.MaxLineLength, priority: :low, max_length: 120},
         {Credo.Check.Readability.ModuleAttributeNames},
         {Credo.Check.Readability.ModuleDoc, false},
-        {Credo.Check.Readability.AliasOrder},
```
ayrat555: This line should be uncommented I think

.credo.exs
```diff
@@ -87,7 +87,7 @@
...
+        {Credo.Check.Readability.AliasOrder, false},
```
ayrat555: Is there any reason to disable this check?

InoMurko: It didn’t seem to add much value to me. Happy to put it back though!

MattMSumner: Might be worth leaving it as is for now and changing in a separate PR in case anyone has any strong feelings.

PR #545, opened 2018-10-27
Variety of Test Case Fixes for CI

hayesgm:
This patch includes a variety of fixes to test cases specifically for running CI.

1. We change from CircleCI's parallelism feature to workflows. Workflows better support the multiple-container model and provide better results in general. We run the previous jobs (tests, credo, etc) and we run one job per hard-fork. Example Results: https://circleci.com/workflow-run/5afec9b8-bea0-4a95-a8c0-c507ae5d640d
2. We stop using the macro `define_common_tests`, and begin to simply run all of the common tests in a standard Elixir test (similar to how `blockchain_test.exs` works). While there's more work to do on consolidation, this fixes the issue of large initial heap usage (fixing error 137 from CircleCI and allowing us to use standard sized instances).
3. We do not fail if `dev.secret.exs` does not exist. The file is not strictly required for most operations and therefore should not be a load failure (esp. since not having it even breaks `mix deps.get`).
4. We bump up to libsecp256k1 version 10 which uses mix build and generally reduces the burden of building libsecp256k1.
5. We exclude network and slow tests by default from our test runner.

Sorry that the PR is a little one, but it took all of these changes together to get a fully passing CI build.


InoMurko:
This is great. Thanks!

apps/blockchain/test/blockchain/block_test.exs
```diff
@@ -9,40 +9,47 @@ defmodule Blockchain.BlockTest do
...
-          accounts: get_test_accounts(test["alloc"])
-        }
+  test "eth common tests" do
+    common_tests = EthCommonTest.Helpers.test_files("GenesisTests")
+
+    for test_path <- common_tests do
+      test_data = EthCommonTest.Helpers.read_test_file(test_path)
+
+      for {internal_test_name, test} <- test_data do
+        # TODO: Enum.member? why?
```
ayrat555: I think it's better to create GitHub issues instead of leaving `TODO` comments in the code. Also, `TODO` comments are credo warnings (currently, it's disabled)

PR #547, opened 2018-10-28
Extract `run_common_tests` for Block and Transaction Test

hayesgm:
The goal of this patch is to re-unify `run_common_tests` through all of our test cases. Previously, we generated a `test do...` for each test, which caused a number of issues. `BlockchainTest` opted instead of run a single Elixir test that creates async tasks for each test and aggregates the results at the end. This patch seeks to move the rest of the testcases on to this style, which is significantly faster overall. We also start to extract "known_failures" to a .txt file so we have a clear indication of failing tests in a limited set of places.


apps/blockchain/lib/eth_common_test/helpers.ex
```diff
@@ -109,4 +111,100 @@ defmodule EthCommonTest.Helpers do
...
+  end
+
+  @spec run_common_tests(String.t(), (String.t(), %{} -> no_return())) :: no_return()
+  def run_common_tests(test_set_name, runner) do
+    # `common_tests` are simply a list of paths to tests for test_set_name
+    # these are top-level folder names in `ethereum_common_tests`
+    common_tests = EthCommonTest.Helpers.test_files(test_set_name)
+
+    known_failures =
+      case File.read(known_failures_path()) do
```
ayrat555: maybe we should use `File.read!`. Because current pattern matching can potentially skip files if a wrong path is provided

apps/blockchain/lib/eth_common_test/helpers.ex
```diff
@@ -109,4 +111,100 @@ defmodule EthCommonTest.Helpers do
...
     Path.join(System.cwd(), "/../../ethereum_common_tests")
   end
+
+  @spec known_failures_path :: String.t()
+  def known_failures_path do
+    Path.join(System.cwd(), "/test/support/known_failures.txt")
+  end
+
+  @spec run_common_tests(String.t(), (String.t(), %{} -> no_return())) :: no_return()
+  def run_common_tests(test_set_name, runner) do
```
ayrat555: can we split `run_common_tests` to more compact functions?

apps/blockchain/lib/eth_common_test/helpers.ex
hayesgm: The idea was that we always look for "known_failures.txt" and if it's not there, then we expect all tests to pass. In other words, the default is to run all tests, but you can include `known_failures.txt` if some common tests are not yet passing.

hayesgm: (Part of the reason for this is that one of the tests was ignoring a common test and it wasn't obvious until we dug into the file that it was skipping a core test. This is meant to hang a lampshade on the issues).

PR #548, opened 2018-10-28
Add a CLI for Syncing

hayesgm:
This patch begins to add a CLI for syncing the blockchain. We create a new sub-app `CLI` to host command-line functions, which is currently limited to syncing. Over time, we will expand this CLI to sync, as well as host an RPC server, etc. We allow syncing from HTTP or IPC (e.g. Infura or a local node).


ayrat555:
It would be nice to have at least a couple of tests. There are a couple of testing libraries that serve prepared responses

hayesgm:
@ayrat555 Agree with adding tests. I wanted to make sure the general set-up worked first and get high-level feedback before adding them.

apps/cli/lib/sync/rpc.ex
```diff
@@ -0,0 +1,230 @@
...
+  @spec setup(String.t() | nil) :: {:ok, state} | {:error, any()}
+  def setup(provider_url) do
+    provider_url = if provider_url, do: provider_url, else: @provider_url
+
+    {:ok, _started} = Application.ensure_all_started(:ethereumex)
+
+    state =
+      case URI.parse(provider_url) do
+        %URI{scheme: scheme} when scheme == "http" or scheme == "https" ->
+          Application.put_env(:ethereumex, :url, provider_url)
```
InoMurko: I think it's a good practice to match it against the expected return value, this case it's :ok

InoMurko:
There's a problem with this approach I think. Namely mix tasks and other tools related to mix are not available in an OTP release.
The way to do this (CLI) would be utilising escript. 
http://erlang.org/doc/man/escript.html
https://hexdocs.pm/mix/master/Mix.Tasks.Escript.Build.html

apps/cli/lib/cli.ex
```diff
@@ -0,0 +1,59 @@
...
+          current_block
+      end
+
+    sync_provider.add_block_to_tree(provider_state, db, chain, tree, current_block.header.number)
+  end
+
+  # The name of the database (e.g. for RocksDB) to store loaded blocks in.
+  @spec db_name() :: String.t()
+  defp db_name() do
+    env = Mix.env() |> to_string()
```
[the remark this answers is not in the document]
hayesgm: Well, I do think that this db name needs to be based on the chain more than the env. But I imagined `Mix.env()` would just always equal `production` in a production release. I do believe it should be `baked` into a release, as opposed to based on env vars.

apps/cli/lib/cli.ex
```diff
@@ -0,0 +1,59 @@
...
+  ./mana --sync --rpc --submit-transaction "{...}"
+  ```
+  """
+
+  @doc """
+  Initiates a sync with a given provider (e.g. a JSON-RPC client, such
+  as Infura). This is the basis of our "sync the blockchain" code.
+  """
+  def sync(sync_provider \\ CLI.Sync.Infura, provider_args \\ []) do
+    db = MerklePatriciaTree.DB.RocksDB.init(db_name())
```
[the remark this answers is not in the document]
hayesgm: Sure, I like fully qualified if it's used sparingly (and aliased if it's used a lot), but happy to change to our consistent format.

apps/cli/lib/sync/rpc.ex
hayesgm: Good call.

hayesgm:
@InoMurko Based on Distillery docs here: http://blog.firstiwaslike.com/elixir-deployments-with-distillery-running-ecto-migrations/, we can use a script to access the `CLI` module. So that means we'll switch from a mix task to a rel script, but the majority of the code here will remain the same.

InoMurko:
Is comparing migrations with a CLI sensible? Migration (or prestart hook) would run before your release starts and block it’s execution until it’s done and then start the res of the app. I imagine the CLI should be running in parallel with the rest of the application and talk to the main application. I might be missing something ...

hayesgm:
The suggestion of Distillery does not use hooks, etc, and instead suggests creating a simple bash script that wraps a call to an Elixir function in the release, e.g.

`rel/commands/migrate.sh`

```bash
#!/bin/sh

release_ctl eval --mfa "MyApp.ReleaseTasks.migrate/1" --argv -- "$@"
```

We could easily replace that call to `CLI.sync/1` or whatever appropriate function. This would be a strong entrypoint into the application.

I've tended to avoid escripts due to Elixir's notes on escripts here: https://hexdocs.pm/mix/master/Mix.Tasks.Escript.Build.html

> Escripts should be used as a mechanism to share scripts between developers and not as a deployment mechanism. For running live systems, consider using mix run or building releases. See the Application module for more information on systems life- cycles.

hayesgm:
Oh whoops, @InoMurko the link I sent earlier was the wrong post! I meant to link to the Distillery docs here: https://hexdocs.pm/distillery/guides/running_migrations.html

Yes, migrations from that old blog post were very much not the right solution.

hayesgm:
I'm going to continue to build on this, but at this point syncing is working for me with this mix task. I believe we should merge this now if it looks good, then next item will be integrating it with distillery once that PR is merged.

apps/cli/lib/sync.ex
```diff
@@ -0,0 +1,101 @@
...
+        tree,
+        block_number,
+        block_limit \\ :infinite,
+        highest_known_block_number \\ nil
+      ) do
+    case should_continue?(block_limit) do
+      :stop ->
+        {:ok, tree}
+
+      {:continue, next_block_limit} ->
```
InoMurko: might be good to extract or refactor this?

apps/cli/README.md
```diff
@@ -0,0 +1,21 @@
+# Cli
+
+**TODO: Add description**
```
masonforest: Can we remove this line/file? Or maybe put the CLI docs here? @andogro has been doing a great job helping us keep our documentation accurate and up to date. Let's remember to discuss our options next time we all meet.

apps/blockchain/lib/blockchain/blocktree.ex
```diff
@@ -76,4 +77,19 @@ defmodule Blockchain.Blocktree do
...
     %{blocktree | best_block: new_best_block}
   end
+
+  @doc """
+  Returns the best block in a tree, which is either the listed best block,
+  or it's the genesis block, which we create.
+  """
+  @spec get_best_block(t(), Chain.t(), DB.db()) :: {:ok, Block.t()} | {:error, any()}
+  def get_best_block(blocktree, chain, db) do
+    case blocktree.best_block do
```
masonforest: Would this work?

    if is_nil(blocktree.best_block) do
        {:ok, Genesis.create_block(chain, db)}
    else
      Block.get_block(block.block_hash, db)
    end

I've been going with `if` if there are two branches and `case`/`cond` if there are more then two branches. Happy to use `case/cond` over `if` all the time as long as we're consistent.

hayesgm: Sure, this came from the original sync script, happy to convert to `if`.

apps/cli/lib/sync.ex
hayesgm: Right now, I just wanted some progress to let this run indefinitely or exit. I think it's worth refactoring this once we realize how P2P syncing will work.

apps/cli/README.md
hayesgm: Yeah, I think getting some basic docs/info might be a good idea. I can write something simple.

PR #551, opened 2018-10-29
Adds Ethash dataset and cache sizes

germsvel:
This is part of https://github.com/mana-ethereum/mana/issues/541

Description
===========

This is the first step in adding the Ethash algorithm for Proof-of-work. We follow the implementation outlined in Appendix J of the Yellow Paper (Byzantium edition).

We introduce `dataset_size`, `cache_size` and `epoch` functions outlined in J.2. We also introduce cached sizes for the first 2048 epochs. This data comes from https://github.com/ethereum/wiki/wiki/Ethash#data-sizes and significantly speed up the retrieval of those sizes.


apps/blockchain/lib/blockchain/ethash.ex
```diff
@@ -0,0 +1,56 @@
...
+      nil -> calculate_cache_size(epoch)
+      found_cache -> found_cache
+    end
+  end
+
+  defp calculate_cache_size(epoch) do
+    prime(@j_cacheinit + @j_cachegrowth * epoch - @j_hashbytes, @j_hashbytes)
+  end
+
+  defp prime(x, y) do
```
masonforest: Is there a name that would be more descriptive then `prime`? Is this `greatest_common_factor`? Or `greatest_common_prime` or something?

apps/blockchain/lib/blockchain/ethash.ex
```diff
@@ -0,0 +1,56 @@
...
+      found_data_size -> found_data_size
+    end
+  end
+
+  defp calculate_dataset_size(epoch) do
+    prime(@j_datasetinit + @j_datasetgrowth * epoch - @j_mixbytes, @j_mixbytes)
+  end
+
+  def cache_size(epoch, cache \\ Ethash.Cache.cache_sizes()) do
+    case Enum.at(cache, epoch) do
```
masonforest: would this work?

    Enum.at(cache, epoch, calculate_cache_size(epoch))

apps/blockchain/lib/blockchain/ethash/cache.ex
```diff
@@ -0,0 +1,4106 @@
+defmodule Ethash.Cache do
+  @data_sizes [
+    1_073_739_904,
```
ayrat555: maybe we can move these values to text files?

apps/blockchain/lib/blockchain/ethash.ex
hayesgm: I believe that wouldn't short-circuit evaluation

apps/blockchain/lib/blockchain/ethash/cache.ex
hayesgm: Agree, this can be loaded from a file in app start-up

apps/blockchain/lib/blockchain/ethash.ex
hayesgm: from the document, they state:

> we always take the highest prime below the linearly growing threshold

apps/blockchain/lib/blockchain/ethash.ex
germsvel: We were actually wondering about that. I'll take another look to see if we could do that if it short-circuits evaluation. I think it would be nicer

apps/blockchain/lib/blockchain/ethash.ex
germsvel: Yeah, great idea. I'll update this to something more descriptive.

germsvel: @masonforest @hayesgm we just renamed it, but I'm not actually sure it's that much clearer. The difficult aspect of this is that the function doesn't actually return the highest prime number. It returns the largest size below the upper bound for which `x/y` is a prime number. Do you have any suggestions for other names or any thoughts on this name?

apps/blockchain/lib/blockchain/ethash.ex
MattMSumner: `Enum.at(cache, epoch) || calculate_cache_size(epoch)` may make the short circuiting behavior more obvious.

germsvel: ☝️ we added that in f827391ca05a1bd35352b66092040fc8aea17071. @masonforest your suggestion unfortunately executes `calculate_cache_size(epoch)` every time, so it doesn't short-circuit it (like Geoff mentioned)

apps/blockchain/lib/blockchain/ethash.ex
```diff
@@ -0,0 +1,65 @@
...
+  end
+
+  defp calculate_cache_size(epoch) do
+    largest_size(
+      @j_cacheinit + @j_cachegrowth * epoch - @j_hashbytes,
+      factor_size: @j_hashbytes
+    )
+  end
+
+  defp largest_size(upper_bound, factor_size: unit_size) do
```
MattMSumner: @masonforest we went back and forth a lot on this name. How do you feel about the names now?

apps/blockchain/lib/blockchain/ethash.ex
masonforest: I think I like `prime` over `largest_size`.

Other ideas:

`max_prime_quotient`?
`max_prime`?
`upper_bound_prime`?

apps/blockchain/lib/blockchain/ethash.ex
masonforest: I think I like `prime` over `largest_size`.

Other ideas:

`max_prime_quotient`?
`max_prime`?
`upper_bound_prime`?
`largest_size_prime`?

apps/blockchain/lib/blockchain/ethash.ex
germsvel: Unfortunately, `prime` is not actually a descriptive name because that's not what it returns. It actually returns numbers that aren't prime. It's just that the `x/y` has to be prime. it's like `x` needs to be prime _relative_ to a `y` factor or something like that.

apps/blockchain/lib/blockchain/ethash.ex
MattMSumner: The one hesitation we had with including prime in the name is that this method does not return a prime number. It returns the amount of bytes required so that when you split it up into blocks of memory (i.e. 128 or 64) you get a prime number for the amount you can store.

Some examples of what this returns include: 

```
1073739904
1082130304
1090514816
```

Which are all divisible by 128 so definitively not prime.

apps/blockchain/lib/blockchain/ethash.ex
hayesgm: I mean, `largest_bounded_relative_prime` with a comment above it describing what you just wrote would be fine. It's definitely not an intuitive function at all (as with many Ethereum functions...)

apps/blockchain/lib/blockchain/ethash/cache.ex
hayesgm: Sure, this means they would be baked into the compiled code? I think that's a fine solution.

PR #552, opened 2018-10-30
Trie with cache

ayrat555:
with https://github.com/mana-ethereum/mana/pull/542 replaces regular Trie with CachingTrie which doesn't write any data to disk

It commits cache after a successful block validation

Part of https://github.com/mana-ethereum/mana/issues/538


apps/blockchain/lib/blockchain/account.ex
```diff
@@ -6,6 +6,7 @@ defmodule Blockchain.Account do
 
   alias ExthCrypto.Hash.Keccak
   alias MerklePatriciaTree.Trie
```
MattMSumner: Since we're moving to use `TrieStorage` as the interface for interacting with the Trie, would it make sense to remove this and force account to interact with the Trie via `TrieStorage`?

apps/blockchain/lib/blockchain/account.ex
```diff
@@ -543,18 +544,20 @@ defmodule Blockchain.Account do
   @spec put_storage(EVM.state(), Address.t(), integer(), integer()) :: EVM.state()
```
MattMSumner: Does this spec need to be updated to match the new return type?

apps/blockchain/lib/blockchain/account.ex
```diff
@@ -543,18 +544,20 @@ defmodule Blockchain.Account do
...
-      updated_storage_trie = Storage.put(state.db, acct.storage_root, key, value)
+      {updated_storage_trie, updated_trie} = Storage.put(state, acct.storage_root, key, value)
+      root_hash = TrieStorage.root_hash(updated_storage_trie)
 
-      %{acct | storage_root: updated_storage_trie.root_hash}
+      {%{acct | storage_root: root_hash}, updated_trie}
     end)
   end
 
   @spec remove_storage(EVM.state(), Address.t(), integer()) :: EVM.state()
```
MattMSumner: Same here?

apps/blockchain/lib/blockchain/account.ex
```diff
@@ -614,7 +617,7 @@ defmodule Blockchain.Account do
   @spec clear_balance(EVM.state(), Address.t()) :: EVM.state()
```
MattMSumner: And here? Maybe the `EVM.state()` type has changed and I'm wrong?

apps/blockchain/lib/blockchain/account/storage.ex
```diff
@@ -9,35 +9,30 @@ defmodule Blockchain.Account.Storage do
   """
 
   alias ExthCrypto.Hash.Keccak
-  alias MerklePatriciaTree.{DB, Trie}
+  alias MerklePatriciaTree.TrieStorage
```
MattMSumner: Could these two lines be `alias MerklePatriciaTree.{TrieStorage, Trie}`? This could also be a case where we can remove the `Trie` alias and force account/storage to interact through `TrieStorage` if it makes sense.

apps/blockchain/lib/blockchain/block.ex
```diff
@@ -710,40 +711,45 @@ defmodule Blockchain.Block do
...
-  def add_transactions(block, transactions, db, chain) do
-    block
-    |> process_hardfork_specifics(chain, db)
-    |> do_add_transactions(transactions, db, chain)
-    |> calculate_logs_bloom()
+  def add_transactions(block, transactions, trie, chain) do
+    {updated_block, updated_trie} = process_hardfork_specifics(block, chain, trie)
+
+    {updated_block, updated_trie} =
+      do_add_transactions(updated_block, transactions, updated_trie, chain)
```
MattMSumner: If we updated the arguments that `do_add_transactions` and `calculate_logs_bloom` expects I think this could become:

```ex
def add_transactions(block, transactions, trie, chain) do
  process_hardfork_specifics(block, chain, trie)
  |> do_add_transactions(transactions, chain)
  |> calculate_logs_bloom
end

def do_add_transactions({updated_block, updated_trie}, transactions, chain) do
  # ...
end

def calculate_logs_bloom({updated_block, updated_trie}) do
  # ...
  {updated_block, updated_trie}
end
```

apps/blockchain/lib/blockchain/genesis.ex
```diff
@@ -45,7 +46,7 @@ defmodule Blockchain.Genesis do
           extra_data: "55555555555555555555555555555555",
           gas_limit: 16777216,
           parent_hash: <<0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0>>,
-          state_root: <<33, 123, 11, 188, 251, 114, 226, 213, 126, 40, 243, 60, 179, 97, 185, 152, 53, 19, 23, 119, 85, 220, 63, 51, 206, 62, 112, 34, 237, 98, 183, 123>>,
+          state_root: <<232, 107, 69, 139, 178, 253, 233, 218, 130, 174, 216, 128, 173, 120, 27, 24, 193, 116, 237, 198, 52, 187, 160, 193, 55, 110, 174, 93, 47, 62, 189, 214>>,
```
MattMSumner: Funny that the state root changed. Is this expected?

apps/merkle_patricia_tree/lib/merkle_patricia_tree/caching_trie.ex
```diff
@@ -51,13 +72,36 @@ defmodule MerklePatriciaTree.CachingTrie do
   @impl true
   def remove_key(caching_trie, key) do
     key_nibbles = Helper.get_nibbles(key)
+    trie_change = {:remove, caching_trie.in_memory_trie.root_hash, key}
```
MattMSumner: I wonder if this could be extracted into a function we could chain on the end of `updated_caching_trie` i.e.

```ex
caching_trie
|> fetch_node()
|> Destroyer.remove_key(key_nibbles, caching_trie)
|> put_node(caching_trie)
|> into(caching_trie)
|> store()
|> add_to_trie_changes(caching_trie.in_memory_trie.root_hash, key)
```

apps/merkle_patricia_tree/lib/merkle_patricia_tree/caching_trie.ex
```diff
@@ -66,17 +110,58 @@ defmodule MerklePatriciaTree.CachingTrie do
       remove_key(caching_trie, key)
     else
       key_nibbles = Helper.get_nibbles(key)
+      trie_change = {:update, caching_trie.in_memory_trie.root_hash, key, value}
```
MattMSumner: A similar thought here around extracting a function to tack onto the end of our function chain below https://github.com/mana-ethereum/mana/pull/552/files#diff-4ae4307fc5394786d95e029e0b1bb41eR125

apps/blockchain/lib/blockchain/genesis.ex
ayrat555: my bad. it was fixed with  remaining genesis tests

apps/blockchain/lib/blockchain/account.ex
ayrat555: this alias is used for types from Trie module

apps/blockchain/lib/blockchain/account.ex
ayrat555: signature didn't change

apps/blockchain/lib/blockchain/block.ex
ayrat555: I will update it in the next PR

PR #556, opened 2018-10-31
Add build-essential for debian/ubuntu installation

germsvel:
I am not a debian/ubuntu user, but when running the `bin/setup` script in a remote box, the installation failed once because `make` was not available and another time because `c++` was not available (when compiling `rox`).

To fix this, I know we can add the `build-essential` package which includes `make`, `gcc`, and `g++`. I don't believe the same package is available for fedora/centos, so I did not include that package for those two.

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -8,15 +8,15 @@ fancy_echo() {
 }
 
 install_linux_deps() {
-  local LINUX_DEPS="automake autoconf autogen libtool gcc curl"
+  local LINUX_DEPS="automake autoconf autogen libtool curl"
   local flavor=`grep "^ID=" /etc/os-release | cut -d"=" -f 2`
   echo $flavor
   case $flavor in
     debian|ubuntu)
-      sudo apt-get install -y $LINUX_DEPS libgmp-dev erlang-dev
+      sudo apt-get install -y $LINUX_DEPS build-essential libgmp-dev erlang-dev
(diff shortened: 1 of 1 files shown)


ayrat555:
While you're at it can you please remove `libsecp256k1` instructions from this script. It was fixed

hayesgm:
Note: pull #543 changes the dep from `rox` to an erlang rocksdb library. This may change the requirements here.

hayesgm:
Can we rebase this on master and verify it works as expected with the changes to rocksdb from rox?

germsvel:
@hayesgm I'm not sure I'll be able to test this properly again. I was only updating this as I was setting up a new remote box. I'm not really a `debian/ubuntu` user, so I don't have a machine on which I can test this again. Do we want to just close this or does it seem like worth merging as is?

ayrat555:
@hayesgm @germsvel what do you think about running this script in the `build` CircleCI workflow?

germsvel:
@ayrat555 so long as it doesn't make the build process longer, I'd be okay. So I think in practice that would mean so long as it is faster than Constantinople common tests, then we could do that.

germsvel:
I'm closing this in favor of https://github.com/mana-ethereum/mana/issues/587. I think we should have a way of knowing whether the changes work or not.

Issue #541, opened 2018-10-25
Implement Ethash (PoW) consensus algorithm


germsvel:
We need to implement the consensus algorithm mentioned in Section 11.5 of the Yellow Paper: 

![screen shot 2018-10-25 at 2 08 55 pm](https://user-images.githubusercontent.com/3245976/47520934-91897e80-d85f-11e8-82f1-4e40a7d0c569.png)

A full description is in Appendix J of the Yellow Paper. 

A helpful explanation of part of the algorithm can be found in http://diyhpl.us/~bryan/papers2/bitcoin/meh/hashimoto.pdf, and geth's implementation of the hashimoto algorithm can be found [here](https://github.com/ethereum/go-ethereum/blob/master/consensus/ethash/algorithm.go#L334)

PR #561, opened 2018-11-01
Implement cache calculation for Ethash

germsvel:
This is part of #541

Description
===========

Implements the cache calculation for Ethash outlined in Appendix J, section 3, of the Yellow Paper. Calculation of the `cache` requires an initial cache (already implemented) and the RandMemoHash algorithm, which is also implemented in this commit.


apps/blockchain/lib/blockchain/ethash/rand_memo_hash.ex
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  @spec second_index(integer(), integer(), Ethash.cache()) :: integer()
+  defp second_index(i, n, cache) do
+    <<header, _rest::binary>> = Enum.at(cache, i)
+    Integer.mod(header, n)
+  end
+
+  @spec binary_xor(binary(), binary(), binary()) :: binary()
+  defp binary_xor(a, b, acc \\ <<>>)
```
germsvel: There's probably a better place for this to live. I wasn't sure where it might be, so I included it here. But if others have better suggestions, I'd be happy to move it out of her.e

apps/blockchain/lib/blockchain/ethash.ex
```diff
@@ -80,10 +85,25 @@ defmodule Blockchain.Ethash do
     |> is_nil
   end
 
+  @doc """
+  Calculates the cache, c, outlined in Appendix J section J.3.2 of the Yellow
+  Paper, by performing the RandMemoHash algorithm n times (where n is 3 if it is
+  j_cacherounds)
+  """
+  @spec calculate_cache(cache(), integer()) :: cache()
+  def calculate_cache(initial_cache, number_of_rounds \\ @j_cacherounds)
```
InoMurko: whats the reason this is public?

hayesgm: Probably to allow unit testing. So is life with these things. Though you could move the function to a sub module to reduce exposure, it’s probably fine to have public.

apps/blockchain/lib/blockchain/ethash/rand_memo_hash.ex
```diff
@@ -0,0 +1,58 @@
+defmodule Blockchain.Ethash.RandMemoHash do
+  alias Blockchain.Ethash
+  alias ExthCrypto.Hash.Keccak
+
+  use Bitwise
+
+  @doc """
```
hayesgm: Awesome, will review more thoroughly but great work getting this done. We should make a note that none of our crypto algorithms are “safe” in the sense of resistance to timing or cache attacks, but since we’re not mining we should be okay with correct but not “safe.”

apps/blockchain/lib/blockchain/ethash.ex
InoMurko: Yeah, moving it to a separate module or testing the consequence is preferred I think.

hayesgm: I would argue, overall, that public/private functions aren't generally worth a refactor unless you expect people to be using your API and it would get in the way. Since this module itself is effectively private, I would argue we should keep it as it is unless it gets out of hand.

InoMurko: If the module is effectively private we can make sure that functions stay private too, I think. 
Static code analysis is also unable to report whether they’re being used or not which affects refactoring and maintainability. 
Chaning private functions also has a bigger side effect on tests, which shouldn’t be the case.

hayesgm: We should move this conversation off-line, but I tend to say let the original coder decide such issues.

germsvel: @InoMurko I am glad you're bringing up the point of public functions that shouldn't be! I totally agree with you. My approach right now is to leave some of these public _in the meantime_. 

I'm trying to implement the Ethash algorithm piece by piece so it's easier for others to review (and it is easier for me to understand too), but once the algorithm is fully implemented, I do intend to make all functions that should be private into private functions, and I'll remove the appropriate tests. 

As we go along, if it makes sense to extract some functionality into other modules, I'm happy to do that. I'm always on board of refactoring as we go. 

Does that work for you?

InoMurko: Great! 👍🏻

apps/blockchain/lib/blockchain/ethash/rand_memo_hash.ex
germsvel: So glad you mentioned that! It's exactly what I was looking for. Updated in d871d995c6244239a8d541bbd364e551438b0a67

apps/blockchain/lib/blockchain/ethash/rand_memo_hash.ex
```diff
@@ -0,0 +1,58 @@
...
+  ```
+  ERMH(x) =  [Ermh(x, 0), Ermh(x, 1), ..., Ermh(x, n - 1)]
+
+  where
+    Ermh(x, i) = KEC512(x'[(i - 1 + n) mod n] XOR x'[x'[i][0] mod n])
+    with x' = x except x'[j] = Ermh(x,j) for all j < i
+  ```
+  """
+  @spec hash(Ethash.cache()) :: Ethash.cache()
+  def hash(original_cache) when is_list(original_cache) do
```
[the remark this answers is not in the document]
germsvel: Good question. It's not really needed. It was more to clarify the api, but it seems overkill now that I'm looking at it again. I removed it in d275b9725bbeaf529bc2b14c59a57606b26210de

PR #564, opened 2018-11-01
Simple Sync Script from Release

hayesgm:
This patch adds a simple script on how to run sync after building a release. There's work to do to get these scripts running correctly in all environments, but this should be the easiest / most straight-forward process for calling into the CLI from a release. The framework comes from [distrillery - migrations](https://github.com/bitwalker/distillery/blob/master/docs/guides/running_migrations.md#custom-command). The biggest open issue is getting the correct build artifact (currently the $release var in the script and correct rocks db filename (currently a relative in CLI.Config).

You can run, for example:

```bash
rel/commands/sync.sh --chain ropsten
```

### DIFF
--- a/apps/cli/lib/config.ex
+++ b/apps/cli/lib/config.ex
@@ -15,6 +15,16 @@ defmodule CLI.Config do
   """
   @spec db_name(atom()) :: nonempty_charlist()
   def db_name(chain_id) do
-    'db/mana-' ++ Atom.to_charlist(chain_id)
+    "db/mana-#{Atom.to_string(chain_id)}"
+    |> relative_to_pwd()
+    |> String.to_charlist()
+  end
+
+  @spec relative_to_pwd(String.t()) :: String.t()
+  defp relative_to_pwd(rel) do
+    Path.join(
+      System.cwd!(),
+      "/../../../../#{rel}"
+    )
   end
 end
--- a/rel/commands/sync.sh
+++ b/rel/commands/sync.sh
@@ -1,0 +1,4 @@
+#!/bin/sh
+
+release="_build/dev/rel/mana/bin/mana"
+$release eval --mfa "Mix.Tasks.Sync.run/1" --argv -- "$@"


apps/cli/lib/config.ex
```diff
@@ -15,6 +15,16 @@ defmodule CLI.Config do
...
+    "db/mana-#{Atom.to_string(chain_id)}"
+    |> relative_to_pwd()
+    |> String.to_charlist()
+  end
+
+  @spec relative_to_pwd(String.t()) :: String.t()
+  defp relative_to_pwd(rel) do
+    Path.join(
+      System.cwd!(),
+      "/../../../../#{rel}"
```
MattMSumner: Is this dependent on local machine setup?

apps/cli/lib/config.ex
```diff
@@ -15,6 +15,16 @@ defmodule CLI.Config do
...
   @spec db_name(atom()) :: nonempty_charlist()
   def db_name(chain_id) do
-    'db/mana-' ++ Atom.to_charlist(chain_id)
+    "db/mana-#{Atom.to_string(chain_id)}"
+    |> relative_to_pwd()
+    |> String.to_charlist()
+  end
+
+  @spec relative_to_pwd(String.t()) :: String.t()
+  defp relative_to_pwd(rel) do
```
germsvel: Might [Path.relative_to_cwd/1](https://hexdocs.pm/elixir/Path.html#relative_to_cwd/1) be useful here?

rel/commands/sync.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+release="_build/dev/rel/mana/bin/mana"
+$release eval --mfa "Mix.Tasks.Sync.run/1" --argv -- "$@"
```
germsvel: Not super familiar with `distillery`, but I was perusing the docs. Could we use `release_ctl` here instead of the combination `release="_build/dev/rel/mana/bin/mana"` and `$release`? 

See [shell scripts](https://hexdocs.pm/distillery/extensibility/shell_scripts.html) and an example of a [custom command](https://hexdocs.pm/distillery/guides/running_migrations.html#custom-command)

PR #567, opened 2018-11-02
Storage optimisations

ayrat555:
Optimisations:
- we can optimize writing  key-value pairs (`{block_hash, block}`, `{account_code_hash, account_code}` etc) when committing raw in-memory cache (simple map of key-value pairs) by using batch updates
- instead of updating permanent trie by traversing Merkle Trie we can insert data from in-memory-trie by fetching all key-value pairs from ets table 

Resolves https://github.com/mana-ethereum/mana/issues/559 https://github.com/mana-ethereum/mana/issues/538

### DIFF
--- a/apps/merkle_patricia_tree/lib/merkle_patricia_tree/db/ets.ex
+++ b/apps/merkle_patricia_tree/lib/merkle_patricia_tree/db/ets.ex
@@ -4,14 +4,12 @@ defmodule MerklePatriciaTree.DB.ETS do
   is backed by :ets.
   """
 
-  alias MerklePatriciaTree.{DB, Trie}
-
   @behaviour MerklePatriciaTree.DB
 
   @doc """
   Performs initialization for this db.
   """
-  @spec init(DB.db_name()) :: DB.db()
+  @impl true
   def init(db_name) do
     :ets.new(db_name, [:set, :public, :named_table])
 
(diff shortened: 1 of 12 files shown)


apps/merkle_patricia_tree/lib/merkle_patricia_tree/db/ets.ex
```diff
@@ -4,14 +4,12 @@ defmodule MerklePatriciaTree.DB.ETS do
...
   """
 
-  alias MerklePatriciaTree.{DB, Trie}
-
   @behaviour MerklePatriciaTree.DB
 
   @doc """
   Performs initialization for this db.
   """
-  @spec init(DB.db_name()) :: DB.db()
```
InoMurko: whats the reasoning behind deleting specs?

ayrat555: @InoMurko `@impl true` annotation checks if the following function is the implementation of a behaviour

https://github.com/elixir-lang/elixir/issues/5734

InoMurko: how does this affect typespecs?

ayrat555: it replaces typespecs in the implementation with those in the behaviour

hayesgm: I'm actually a big fan of `@impl true` since it's possible that behaviors and functions get out of sync in regards to typespecs. This prevents that from happening.

PR #569, opened 2018-11-05
Get ExWire working against local peer

hayesgm:
**Note**: This is not ready for review, but I'm putting up the PR to track progress and get early design feedback.

This patch is for a variety of changes to get ExWire working against a local Geth/Parity node. This should allow us to connect to a Geth node, request block headers / block bodies and begin syncing. We are less concerned with the discovery protocol so far. Note: sync does not currently work since it depends on the old version of blocktree that stores the entire parent map in memory. Thus, we'll need to upgrade that to work.

We also refactor to slightly simplify some logic, improve typespecs and likely, in the process, break test cases.

This, again, is a work in progress, but we now can begin to sync the real chain using P2P. After we get the sync working, this PR will get some re-factoring / test case fixing.

Note: this will likely obviate the need for `CLI.Sync` and that script may be removed wholesale.

Status: we are currently syncing blocks up to block number 4036 on Ropsten. I'm guessing that that's the first block that needs to be split into multiple frames, and thus, the current issue.

Final comment: this (appears) to be way faster than RPC. So far, at least.


apps/cli/lib/cli.ex
```diff
@@ -37,17 +37,24 @@ defmodule CLI do
...
-          block_provider_state,
-          updated_trie,
-          chain,
-          blocktree,
-          current_block.header.number + 1,
-          :infinite,
-          highest_known_block_number
-        )
+        if true do
+          Application.put_env(:ex_wire, :initial_db, db)
```
InoMurko: one of the issues with this approach is side effects. Is there a way we could avoid this?
A common approach is to send params down the supervision tree.

apps/ex_wire/lib/ex_wire/node_discovery_supervisor.ex
```diff
@@ -1,34 +1,37 @@
...
 
     Supervisor.start_link(__MODULE__, params, name: supervisor_name)
   end
 
   def init(params) do
-    {udp_module, udp_process_name} = discovery_param(params, :network_adapter)
-    kademlia_name = discovery_param(params, :kademlia_process_name)
-    port = discovery_param(params, :port)
-    bootnodes = (params[:nodes] || nodes()) |> Enum.map(&Node.new/1)
+    kademlia_name = Keyword.get(params, :kademlia_process_name, KademliaState)
```
InoMurko: is there a reason this is configurable?

apps/cli/lib/cli.ex
hayesgm: Yeah, this was a lazy approach / didn't really solve what I wanted, but if we think about the CLI just starting sub-app components instead of running code itself (e.g. like what we're doing here), then it's more confusing to me how we pass arguments.

Basically, if we want to supervise ExWire when running CLI, do we choose which of the following options to configure ExWire:

1. `Application.ensure_all_started(:ex_wire)`, passing configuration by some global since there's no way to pass args in that call.
2. CLI is an application with a supervision tree, in it's tree we call: `CLI.start(...); ... children = [ ... ] ++ ExWire.get_child_specs(chain)` and monitor ExWire ourselves?
3. Start ExWire in such a way that it doesn't need any of this data, but where each process can be configured after it's started to do the work required.

Note: similar issues exist with configuration Ethereumex, which has a supervision tree and loads its state from env.

Thoughts?

apps/ex_wire/lib/ex_wire/node_discovery_supervisor.ex
hayesgm: Dunno, I just refactored to code to be clearer, but there's not an obvious reason why this process would be renamed by the caller.

apps/ex_wire/lib/ex_wire/config.ex
```diff
@@ -1,96 +1,123 @@
...
-    get_env(:caps)
+  @spec caps(Keyword.t()) :: [{String.t(), integer()}]
+  def caps(given_params \\ []) do
+    get_env(given_params, :caps)
   end
 
   @spec client_id() :: String.t()
-  def client_id do
+  def client_id() do
     version = Mix.Project.config()[:version]
```
InoMurko: this won't work in a release... should this be part of your PR or addressed later?

apps/ex_wire/lib/ex_wire/p2p/manager.ex
```diff
@@ -133,64 +140,74 @@ defmodule ExWire.P2P do
...
+    case packet_handle_response do
+      :ok ->
+        conn
 
-        {:disconnect, reason} ->
-          Packet.Disconnect.new(reason)
-      end
+      :activate ->
+        new_session =
+          if DEVp2p.session_active?(conn.session) do
```
InoMurko: maybe pulling logging out of the function makes more sense.

apps/ex_wire/lib/ex_wire/config.ex
InoMurko: Hm, yeah, interesting question. We haven't discussed this yet! A release configuration setts it's version:
`
release :mana do
  set version: "0.1.0"
  set applications: [
    :runtime_tools,
    blockchain: :permanent,
    cli: :permanent,
    evm: :permanent,
    ex_wire: :permanent,
    exth_crypto: :permanent,
    merkle_patricia_tree: :permanent
  ]
end
`
Now this part of configuration get's evaluated when building so you could potentially set it to:
set version: "#{System.get_env("RELEASE_VERSION")}" or even pull it out of a file or... whatever. Maybe, for consistency reasons having a file in root/priv isn't a bad solution? So if a git tag is a release, that commit needs to update the file too. WDYT?

apps/cli/lib/cli.ex
InoMurko: So yeah, I'm not saying overriding configuration is entirely bad. Though you need to identify top level modules/processes that are allowed to write to it. Also, it's not persisted.

There's a couple of options. One is the above, where you limit yourself where you do that - potentially in the application module or top level task module that starts the top level supervisors and workers, which allows you to pass down the configuration. 

One way you could potentially enforce that is using a ETS table (set to protected) per app where the owner of the table is the top level process. Everyone else just reads.

apps/ex_wire/lib/ex_wire/p2p/manager.ex
```diff
@@ -133,64 +140,74 @@ defmodule ExWire.P2P do
...
-    if DEVp2p.session_active?(session) do
-      do_handle_packet(packet, conn)
-      session
-    else
-      attempt_session_activation(session, packet)
+  @spec handle_packet(module(), Packet.packet(), Connection.t()) :: Connection.t()
+  defp handle_packet(packet_mod, packet, conn) do
+    packet_handle_response = packet_mod.handle(packet)
+
+    if not DEVp2p.session_active?(conn.session) && packet_handle_response != :activate do
```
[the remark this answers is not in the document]
hayesgm: Sure, I like that. pattern matching as truth tables is my favorite.

apps/ex_wire/lib/ex_wire/p2p/manager.ex
hayesgm: Good point, I'll try and reduce side-effects here and leave that to the caller maybe?

.dialyzer.ignore-warnings
```diff
@@ -1,75 +1,3 @@
--------------------------------
```
ayrat555: :clap:

InoMurko: Sweet!!!

apps/blockchain/lib/state.ex
```diff
@@ -1,6 +1,6 @@
-defmodule CLI.State do
+defmodule Blockchain.State do
```
ayrat555: what do you think about moving this module under `Blocktree` namespace. `Blockchain.Blocktree.State`

apps/cli/mix.exs
```diff
@@ -30,9 +30,13 @@ defmodule Cli.MixProject do
...
   defp deps do
     [
-      {:ethereumex, "~> 0.5.0"},
+      # Umbrella deps
       {:blockchain, in_umbrella: true},
+      {:ex_wire, in_umbrella: true, runtime: false},
       {:merkle_patricia_tree, in_umbrella: true},
+
+      # External deps
+      {:ethereumex, "~> 0.5.0"},
```
ayrat555: it should be updated to `0.5.1`. This version replaces Poison with Jason as a json library. Jason should be faster

apps/ex_wire/lib/ex_wire/struct/block_queue.ex
```diff
@@ -201,86 +161,92 @@ defmodule ExWire.Struct.BlockQueue do
...
+              block_queue.backlog,
+              block.header.parent_hash,
+              [block],
+              fn blocks -> [block | blocks] end
+            )
+
+          {block_tree, trie, updated_backlog, []}
+
+        {:invalid, reasons} ->
+          :ok =
```
InoMurko: I think this would fail if you raise log level, no?

InoMurko: I think this would fail if you raise log level, no?

hayesgm: From my two seconds of investigation on a local VM, seems ok:

**config**
```elixir
config :logger,
  level: :warn
```

**iex**
```elixir
iex(2)> Logger.debug("hi")
:ok
iex(3)> Logger.warn("hi")
:ok

03:01:12.289 [warn]  hi
```

apps/cli/mix.exs
hayesgm: done.

apps/blockchain/lib/state.ex
hayesgm: done.

apps/ex_wire/lib/ex_wire/struct/block_queue.ex
InoMurko: You're right, thats how Lager used to behave 🤕

PR #576, opened 2018-11-05
 Precompute `valid_jump_destinations`

masonforest:
Before we would calculate `valid_jump_destinations` on every call to
`valid_jump_dest?`. Now we calculate it once ahead of time.

### DIFF
--- a/apps/evm/lib/evm/machine_code.ex
+++ b/apps/evm/lib/evm/machine_code.ex
@@ -31,38 +31,7 @@ defmodule EVM.MachineCode do
   end
 
   @doc """
-  Returns true if the given new program_counter is a valid jump
-  destination for the machine code, false otherwise.
-
-  TODO: Memoize
-
-  ## Examples
-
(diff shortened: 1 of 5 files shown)


apps/evm/lib/evm/machine_code.ex
```diff
@@ -30,40 +30,9 @@ defmodule EVM.MachineCode do
     |> Operation.metadata()
   end
 
-  @doc """
-  Returns true if the given new program_counter is a valid jump
-  destination for the machine code, false otherwise.
-
-  TODO: Memoize
```
hayesgm: The only question here is: should we memoize versus eager calculate? I'm guessing that every contract worth its salt with either be very short (e.g. no code) or will use jumps (thus requiring us to calculate), but I wanted to entertain the idea that we shouldn't calculate jump destinations ever unless the code actually calls jump/jumpi.

masonforest: Yeah, good to consider.

If we memoized we'd also need to pass the state through somehow, though. I guess `is_invalid_jump_destination?` could return `exec_env` but that seems like overkill for optimizing an edge case. I'm gonna merge as-is and if we realize this is a bottle neck we can continue to optimize it.

PR #577, opened 2018-11-05
Rem poison

InoMurko:
Removed Poison as a dependency because we don't need two json handlers. Jason is faster and it's shipped in ethereumex and credo.


mix.exs
```diff
@@ -39,10 +37,9 @@ defmodule Mana.MixProject do
   defp deps do
     [
       {:ex_rlp, "~> 0.3.1"},
-      {:poison, "~> 4.0.1", runtime: false},
```
masonforest: What do you think about listing `Jason` as a dependency here just to be explicit?

PR #578, opened 2018-11-06
circle strict dialyzer checks

InoMurko:
Strict dialyzer checks that only allow issues from ex_wire.


.dialyzer.ignore-warnings
```diff
@@ -1,3 +1,75 @@
+-------------------------------
+ # TODO:: EX WIRE IS GOING THROUGH LARGE REFACTOR
```
InoMurko: @hayesgm I added ex_wire here

ayrat555: maybe we can leave `ex_wire` ignored? because we'll have to update dialyzer ignore file after every change if  this gets merged

InoMurko: afaik, what @hayesgm is doing is going to change much of typespecs in ex_wire. until then, I would prefer if we strictly don't add any more issues to that pile. if we add the whole app then we're ignoring potential increases.

InoMurko: when @hayesgm rebases, he'll need to clean the ignore file though.

PR #583, opened 2018-11-06
Extract EvmTestRunner to run common_tests

MattMSumner:
Why:

We would like to start profiling our evm code to look for areas where we
can improve performance. [Parity] has a [benchmark script] for
outputting how long each of the evm common_tests take to run

[Parity]: https://github.com/paritytech/parity-ethereum
[benchmark script]: https://github.com/paritytech/parity-ethereum/blob/f6dcca3ebb6341049b8fc4c890a463277c493bc4/scripts/evm_jsontests_bench.sh

This PR:

Extracts EvmTestRunner. This allows us to run the following type of code
for profiling:

```
% cd apps/evm
% MIX_ENV=test iex -S mix

iex(1)> test_json =
...(1)> File.read!("../../ethereum_common_tests/VMTests/vmPerformance/ackermann31.json") |>
...(1)> Jason.decode!() |>
...(1)> Map.get("ackermann31")
iex(2)> :fprof.apply(EvmTestRunner, :run, [test_json])
iex(3)> :fprof.profile
iex(4)> :fprof.analyse([callers: true, sort: :own, totals: true, details: true])
```

### DIFF
--- a/apps/evm/test/support/evm_test_runner.ex
+++ b/apps/evm/test/support/evm_test_runner.ex
@@ -1,0 +1,153 @@
+defmodule EVM.TestRunner do
+  import ExthCrypto.Math, only: [hex_to_bin: 1, hex_to_int: 1]
+
+  alias EVM.Mock.{MockAccountRepo, MockBlockHeaderInfo}
+  alias EVM.{ExecEnv, SubState, VM}
+
+  def run(json) do
+    exec_env = get_exec_env(json)
+    gas = hex_to_int(json["exec"]["gas"])
+    VM.run(gas, exec_env)
+  end
+
+  defp get_exec_env(json) do
+    %ExecEnv{
+      address: hex_to_bin(json["exec"]["address"]),
+      originator: hex_to_bin(json["exec"]["origin"]),
+      gas_price: hex_to_int(json["exec"]["gasPrice"]),
+      data: hex_to_bin(json["exec"]["data"]),
+      sender: hex_to_bin(json["exec"]["caller"]),
+      value_in_wei: hex_to_int(json["exec"]["value"]),
+      machine_code: hex_to_bin(json["exec"]["code"]),
+      account_repo: account_repo(json),
+      block_header_info: block_header_info(json)
+    }
+  end
+
+  defp block_header_info(json) do
+    if json["env"]["currentNumber"] == "0x00" do
+      single_block_header_info(json)
+    else
+      many_blocks_header_info(json)
+    end
+  end
+
+  defp single_block_header_info(json) do
+    build_block_header_info([
+      %Block.Header{
+        number: hex_to_int(json["env"]["currentNumber"]),
+        timestamp: hex_to_int(json["env"]["currentTimestamp"]),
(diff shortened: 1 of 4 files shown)


InoMurko:
whooaaa, this is great! do you think it would make sense to run this separately or after the tests in circle?
this way we could monitor potential performance degradations. wdyt?

apps/evm/test/support/evm_test_runner.ex
```diff
@@ -0,0 +1,110 @@
+defmodule EvmTestRunner do
+  import ExthCrypto.Math, only: [hex_to_bin: 1, hex_to_int: 1]
+
+  alias EVM.Mock.{MockAccountRepo, MockBlockHeaderInfo}
+
+  def run(json) do
```
germsvel: Not sure if this is desired, but I _think_ the other test runners accept a test name and retrieve the file. We could do that to keep in line with them, and it makes it easier to reuse the test runner with a rake task when we only wan to run a single test. What do you think?

apps/evm/test/support/evm_test_runner.ex
```diff
@@ -0,0 +1,110 @@
+defmodule EvmTestRunner do
```
InoMurko: is this the correct namespace?

apps/evm/test/support/evm_test_runner.ex
```diff
@@ -0,0 +1,110 @@
...
+  import ExthCrypto.Math, only: [hex_to_bin: 1, hex_to_int: 1]
+
+  alias EVM.Mock.{MockAccountRepo, MockBlockHeaderInfo}
+
+  def run(json) do
+    exec_env = get_exec_env(json)
+    gas = hex_to_int(json["exec"]["gas"])
+    EVM.VM.run(gas, exec_env)
```
InoMurko: could we alias this?

apps/evm/test/support/evm_test_runner.ex
```diff
@@ -0,0 +1,110 @@
...
+  end
+
+  defp get_exec_env(json) do
+    %EVM.ExecEnv{
```
InoMurko: and this too, credo will nag about it

andogro:
Any clarifications or additional instructions needed in test section of README (or additions to EVM README) for this change?

apps/evm/test/support/evm_test_runner.ex
MattMSumner: I like the idea of doing this for the consistency across test runners. I would like to keep reading files out of the is test_runner to eliminate it as a factor when benchmarking/running profiling. How would you feel if I opened a PR to remove reading files from the other test_runners?

MattMSumner:
@andogro not at this time. This is a small refactor to allow me to manually do some profiling. Once I've learnt more around what is useful to look at (and got all the performance tests running in a reasonable time) I'll likely a test to check for performance degradation as @InoMurko suggested. That will likely require some documentation changes.

apps/evm/test/support/evm_test_runner.ex
MattMSumner: Thanks, changed to `EVM.TestRunner`

PR #592, opened 2018-11-08
fix highest bit calculation for exp_mod builtin

ayrat555:
wrong highest bit calculation caused wrong gas cost.
Because of that we were failing `ropsten` chain syncing on block #2_444_903

fixes https://github.com/mana-ethereum/mana/issues/580

### DIFF
--- a/apps/evm/lib/evm/builtin/mod_exp.ex
+++ b/apps/evm/lib/evm/builtin/mod_exp.ex
@@ -110,7 +110,7 @@ defmodule EVM.Builtin.ModExp do
     end
   end
 
-  @spec highest_bit(binary(), non_neg_interger()) :: non_neg_integer()
+  @spec highest_bit(binary(), non_neg_integer()) :: non_neg_integer()
   defp highest_bit(_, 0), do: 0
 
   defp highest_bit(binary_number, _) do
--- a/apps/evm/lib/evm/builtin/mod_exp.ex
+++ b/apps/evm/lib/evm/builtin/mod_exp.ex
@@ -94,7 +94,9 @@ defmodule EVM.Builtin.ModExp do
   defp e_length_prime(e_length, e, _) when e == 0 and e_length <= 32, do: 0
 
   defp e_length_prime(e_length, e, _) when e_length <= 32 do
-    8 * (e_length - 32) + highest_bit(e)
+    bin_e = :binary.encode_unsigned(e)
+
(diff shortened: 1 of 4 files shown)


apps/evm/lib/evm/builtin/mod_exp.ex
```diff
@@ -92,27 +92,32 @@ defmodule EVM.Builtin.ModExp do
...
     end
   end
 
-  @spec highest_bit(non_neg_integer()) :: integer()
-  defp highest_bit(number) do
-    res =
-      number
-      |> :math.log2()
-      |> MathHelper.floor()
+  def highest_bit(_, 0), do: 0
```
InoMurko: could be private I think?

apps/evm/test/evm/builtin/mod_exp_test.exs
```diff
@@ -0,0 +1,25 @@
+defmodule EVM.Builtin.ModExpTest do
+  use ExUnit.Case, async: true
+
+  alias EVM.Builtin.ModExp
+
+  # https://github.com/mana-ethereum/mana/issues/580
+  test "calculates mod_exp (block #2_444_903, transaction #4 on ropsten)" do
```
InoMurko: great for providing a test!

apps/evm/lib/evm/builtin/mod_exp.ex
ayrat555: @InoMurko done

apps/evm/lib/evm/builtin/mod_exp.ex
```diff
@@ -92,27 +92,33 @@ defmodule EVM.Builtin.ModExp do
...
+      8 * (e_length - 32) + highest_bit(e_first_32_bytes_bin, e_length)
     end
   end
 
-  @spec highest_bit(non_neg_integer()) :: integer()
-  defp highest_bit(number) do
-    res =
-      number
-      |> :math.log2()
-      |> MathHelper.floor()
```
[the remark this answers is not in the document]
ayrat555:
```
  number
      |> :math.log2()
      |> MathHelper.floor()
```
is direct implementation of the equation from the Yellow Paper. It should calculate the number of leading zeroes

It was returning wrong value in #2_444_903 ( 254 instead of 253 ) I think it's precision error of `:math.log2` because the number was very large

PR #595, opened 2018-11-12
Fix DER Encoding Bugs

hayesgm:
This patch addresses a code quality issue that has continually bit us: a DER-encoded public key is `<<0x04, public_key:binary()>>`. The problem with that is, we often rely on the variable name to know whether or not the key was DER encoded or is the raw public key material. This has led to us using DER-encoded keys as standard keys or getting argument errors not giving DER-encoded keys to the `:crypto` module (whereas most Ethereum functions require non-DER encoded keys). We opt to now store DER-encoded keys as `{:public_der, public_key}` so that we cannot possibly pass the value incorrectly to a function (as dialyzer will catch those issues). This direct relates to us having a bug on this topic.


apps/evm/lib/evm/address.ex
```diff
@@ -38,7 +39,7 @@ defmodule EVM.Address do
   @doc """
   Returns an address given an address and a nonce.
   """
-  @spec new(integer(), integer()) :: binary()
+  @spec new(binary(), integer()) :: t()
```
ayrat555: binary() -> t() ?

hayesgm: The type here is an `t :: <<_::160>>`, which represents an Ethereum address. This is to make the type more specific.

germsvel:
Super late to review, but just wanted to say this is a great change! 🎉

PR #597, opened 2018-11-12
Add Fake Kademlia GenServer for Tests

hayesgm:
A few tests relied on having a global `Kademlia.Server` gen server running for the tests to pass. This made these tests very difficult as any changes to configuration caused them to fail, as well as issues with running tests asynchronously. This patch creates a simple fake Kademlia Server that makes sure that each test starts this server as it needs it to reduce global dependencies.


apps/ex_wire/test/support/ex_wire/fake_kademlia_server.ex
```diff
@@ -0,0 +1,25 @@
+defmodule ExWire.FakeKademliaServer do
```
ayrat555: We can create behaviour with kademlia functions and implement them here and in real kademlia module

hayesgm: Yeah, though this is a genserver, so it doesn't work quite well with traditional behaviours since the functions are mostly `handle_call`, etc.

PR #598, opened 2018-11-12
save unmodified accounts to the EVM cache

ayrat555:
Part of https://github.com/mana-ethereum/mana/issues/586

Currently, we are only saving modified account to the EVM cache, we can also save unmodified accounts that were read during EVM execution for faster access.

See also
https://www.badykov.com/ethereum/2018/11/10/storage-in-ethereum/

Update
I ran homestead tests (state and blockchain) using unmodified accounts cache and rocksdb as database. Tests finished in 442 seconds.
without cache tests ran 493.3 seconds


apps/blockchain/test/blockchain/account/repo/cache_test.exs
```diff
@@ -249,6 +249,34 @@ defmodule Blockchain.Account.Repo.CacheTest do
       assert found_code == code
     end
 
+    test "does not update account if account in cache is clean" do
+      ets_db = MerklePatriciaTree.Test.random_ets_db()
+      state = MerklePatriciaTree.Trie.new(ets_db)
+      address = <<1>>
+
+      # we need to create a blank account with storage
+      state_with_account = Account.reset_account(state, address)
```
MattMSumner: How would you feel about removing the above comment in favor of a different variable name? Something like:

```suggestion
      blank_account_with_storage = Account.reset_account(state, address)
```

apps/evm/lib/evm/gas.ex
```diff
@@ -134,36 +134,50 @@ defmodule EVM.Gas do
...
 
-  @spec cost_with_status(MachineState.t(), ExecEnv.t()) :: cost_with_status()
+  @spec cost_with_status(MachineState.t(), ExecEnv.t()) :: {ExecEnv.t(), cost_with_status()}
   def cost_with_status(machine_state, exec_env) do
     operation = MachineCode.current_operation(machine_state, exec_env)
     inputs = Operation.inputs(operation, machine_state)
-    operation_cost = operation_cost(operation.sym, inputs, machine_state, exec_env)
+
+    {updated_exec_env, operation_cost} =
+      case operation_cost(operation.sym, inputs, machine_state, exec_env) do
```
MattMSumner: Since we're passing in `exec_env` to `operation_cost`, it's surprising that it either returns just the cost or a tuple of the cost and exec env. Would it be possible to remove this case statement by having `operation_cost` consistently return `{exec_env, cost}`?

apps/evm/lib/evm/refunds/sstore.ex
```diff
@@ -14,23 +14,31 @@ defmodule EVM.Refunds.Sstore do
...
   defp basic_sstore_refund({key, new_value}, exec_env) do
-    case ExecEnv.get_storage(exec_env, key) do
-      {:ok, value} ->
-        if value != 0 && new_value == 0 do
-          @storage_refund
-        else
+    {updated_exec_env, current_value} = ExecEnv.storage(exec_env, key)
+
+    refund =
+      case current_value do
```
MattMSumner: Could `current_value` return an int? Then we could remove this case statement? i.e.

```ex
refund = if current_value != 0 && new_value == 0 do
  @storage_refund
else
  0
end
```

apps/evm/lib/evm/gas.ex
ayrat555: @MattMSumner most operation cost functions don't modify exec_env. But we can return unmodified exec_env.

apps/blockchain/test/blockchain/account/repo/cache_test.exs
ayrat555: good idea. I change it in my next PR

PR #600, opened 2018-11-12
json rpc inital commit

InoMurko:
This PR adds an umbrella app that serves JSONRPC2. It start off assembling the supervision tree. 
Depending on the configuration provided it starts a TCP supervison tree via Ranch (and opens a socket on the path) and a cowboy based webserver via plug that has different dispatchers (websockets enabled or disabled, http enabled or disabled or both).
Modules like _JSONRPC2.Request_, _JSONRPC2.Response_ serve as serialisers and deserializers for the protocol.
 
All servers register a handler _JSONRPC2.SpecHandler_ that will implement the ETH RPC methods.


apps/jsonrpc2/lib/jsonrpc2/spec_handler.ex
```diff
@@ -0,0 +1,5 @@
+defmodule JSONRPC2.SpecHandler do
+  use JSONRPC2.Server.Handler
+
+  def handle_request("web3_clientVersion", _), do: "Mana 0.0.1"
```
InoMurko: this is just a stub

apps/jsonrpc2/lib/jsonrpc2/server/handler.ex
```diff
@@ -0,0 +1,279 @@
...
+    id = Map.get(request, "id", :undefined)
+
+    if valid_request?(version, method, params, id) do
+      {method, params, id}
+    else
+      :invalid_request
+    end
+  end
+
+  defp parse(_) do
```
MattMSumner: 👍

apps/jsonrpc2/lib/jsonrpc2/server/handler.ex
```diff
@@ -0,0 +1,279 @@
...
+    :invalid_request
+  end
+
+  defp valid_request?(version, method, params, id) do
+    version == "2.0" and is_binary(method) and (is_list(params) or is_map(params)) and
```
MattMSumner: How would you feel about breaking up these lines a bit to improve grokability?
```suggestion
    version == "2.0" and
    is_binary(method) and
    (is_list(params) or is_map(params)) and
    (id in [:undefined, nil] or is_binary(id) or is_number(id))
```

apps/jsonrpc2/lib/jsonrpc2/servers/tcp.ex
```diff
@@ -0,0 +1,67 @@
+defmodule JSONRPC2.Servers.TCP do
+  @moduledoc """
+  A server for JSON-RPC 2.0 using a line-based TCP transport.
+  """
+
+  alias JSONRPC2.Servers.TCP.Protocol
+
+  @default_timeout 1000 * 60 * 60
```
MattMSumner: Love that this is broken up. 💯

apps/jsonrpc2/test/jsonrpc2/http_test.exs
```diff
@@ -0,0 +1,71 @@
+defmodule JSONRPC2.HTTPTest do
+  use ExUnit.Case, async: false
+  alias JSONRPC2.Clients.HTTP
+  alias JSONRPC2.Servers.WebSocketHTTP
+  alias JSONRPC2.SpecHandlerTest
+
+  setup_all do
+    port = :rand.uniform(65_535 - 1025) + 1025
```
MattMSumner: This is basically ensuring the port is between `1026` and `65_535`, correct?

apps/jsonrpc2/config/config.exs
```diff
@@ -0,0 +1,8 @@
+use Mix.Config
+
+config :jsonrpc2,
+  ipc: %{enabled: true, path: Enum.join([System.user_home!(), "/mana.ipc"])},
```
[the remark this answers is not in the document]
InoMurko: Wherever you would most likely have write/read access from your applications or others. I think thath's where Geth is storing it.
`$HOME/.ethereum/geth.ipc` or
`filepath.Join(os.TempDir(), "geth.ipc")`

apps/jsonrpc2/test/jsonrpc2/http_test.exs
InoMurko: yeah

apps/jsonrpc2/lib/jsonrpc2/server/handler.ex
InoMurko: Made a change, take a look please

apps/jsonrpc2/config/config.exs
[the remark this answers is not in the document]
MattMSumner: Would it make sense to have our socket file also in `.ethereum`? i.e. `$HOME/.ethereum/mana.ipc`

apps/jsonrpc2/lib/jsonrpc2/request.ex
```diff
@@ -0,0 +1,41 @@
...
+  """
+
+  @type request ::
+          {JSONRPC2.method(), JSONRPC2.params()}
+          | {JSONRPC2.method(), JSONRPC2.params(), JSONRPC2.id()}
+
+  @doc """
+  Returns a serialized `request` using `serializer`.
+
+  See the README for more information on serializers.
```
hayesgm: These could use `Examples` which would help clear up exactly how the function works.

apps/jsonrpc2/lib/jsonrpc2/request.ex
```diff
@@ -0,0 +1,41 @@
...
+  """
+  @spec serialized_request(request, module) :: {:ok, binary()} | {:error, any}
+  def serialized_request(request, serializer) do
+    serializer.encode(request(request))
+  end
+
+  @doc """
+  Returns a map representing the JSON-RPC2.0 request object for `request`.
+  """
```
hayesgm: Same here with examples, since the function is pretty straight-forward but the typespec doesn't give us many clues.

apps/jsonrpc2/lib/jsonrpc2/response.ex
```diff
@@ -0,0 +1,37 @@
...
+  JSON-RPC 2.0 Response object utilites.
+  """
+
+  @type id_and_response ::
+          {JSONRPC2.id() | nil,
+           {:ok, any} | {:error, code :: integer, message :: String.t(), data :: any}}
+
+  @doc """
+  Deserialize the given `response` using `serializer`.
+  """
```
hayesgm: Same comment as above on Examples

apps/jsonrpc2/lib/jsonrpc2/server/handler.ex
```diff
@@ -0,0 +1,283 @@
...
+      SpecHandler.handle(~s({"jsonrpc": "2.0", "method": "subtract", "params": [42, 23], "id": 1}))
+      #=> ~s({"jsonrpc": "2.0", "result": 19, "id": 1})
+  """
+
+  require Logger
+
+  @doc """
+  Respond to a request for `method` with `params`.
+
+  You can return any serializable result (which will be ignored for notifications), or you can throw
```
hayesgm: Throw? Really? That's so un-erlang.

apps/jsonrpc2/lib/jsonrpc2/server/handler.ex
```diff
@@ -0,0 +1,283 @@
...
+      with or without extra data.
+  """
+  @callback handle_request(method :: JSONRPC2.method(), params :: JSONRPC2.params()) ::
+              JSONRPC2.json() | no_return
+
+  defmacro __using__(_) do
+    quote do
+      @spec handle(String.t()) :: {:reply, String.t()} | :noreply
+      def handle(json) do
+        serializer = Jason
```
hayesgm: Oh hm, we hard-code this instead of `serializer = Application.get_env(:jsonrpc2, :serializer)`?

apps/jsonrpc2/lib/jsonrpc2/server/handler.ex
```diff
@@ -0,0 +1,283 @@
...
+        {:error, :invalid, _number} ->
+          standard_error_response(:parse_error, nil)
+      end
+
+    data
+    |> encode_response(module, serializer, json)
+  end
+
+  def handle(module, serializer, json) do
+    data = parse(json)
```
hayesgm: this can be piped below since it's really `json |> parse |> collate |> encode`

apps/jsonrpc2/lib/jsonrpc2/server/handler.ex
```diff
@@ -0,0 +1,283 @@
...
+  defp collate_for_dispatch(batch_rpc, module)
+       when is_list(batch_rpc) and length(batch_rpc) > 0 do
+    merge_responses(Enum.map(batch_rpc, &dispatch(module, &1)))
+  end
+
+  defp collate_for_dispatch(rpc, module) do
+    dispatch(module, rpc)
+  end
+
+  defp parse(requests) when is_list(requests) do
```
hayesgm: What's the typespec on parse? `parse(list() | map() | any()) :: list({_, _, _}) | {_, _, _} | :invalid_request`?

apps/jsonrpc2/lib/jsonrpc2/server/handler.ex
```diff
@@ -0,0 +1,283 @@
...
+    case is_list(params) or is_map(params) do
+      true -> id in [:undefined, nil] or is_binary(id) or is_number(id)
+      false -> false
+    end
+  end
+
+  defp valid_request?(_version, _method, _params, _id), do: false
+
+  defp merge_responses(responses) do
+    case for({:reply, reply} <- responses, do: reply) do
```
hayesgm: This code, while kind of cool, is a little tricky for its own good.

apps/jsonrpc2/lib/jsonrpc2/servers/tcp/protocol.ex
```diff
@@ -0,0 +1,59 @@
+defmodule JSONRPC2.Servers.TCP.Protocol do
```
hayesgm: What is the non-websockets non-HTTP TPC protocol used for?

apps/jsonrpc2/lib/jsonrpc2/servers/http/plug.ex
```diff
@@ -0,0 +1,109 @@
...
+    do_handle_jsonrpc2(conn, body, opts)
+  end
+
+  defp handle_jsonrpc2(conn, %{"_json" => body_params}, opts),
+    do: do_handle_jsonrpc2(conn, body_params, opts)
+
+  defp handle_jsonrpc2(conn, body_params, opts), do: do_handle_jsonrpc2(conn, body_params, opts)
+
+  defp do_handle_jsonrpc2(conn, body_params, %{handler: handler}) do
+    resp_body =
```
hayesgm: This is really the crux of the library for HTTP(s) requests. Take a request, pass it off to handlers, and then put the response following standard plug rules.

apps/jsonrpc2/lib/jsonrpc2/servers/tcp/protocol.ex
InoMurko: unix domain sockets

apps/jsonrpc2/LICENSE
```diff
@@ -0,0 +1,13 @@
+Copyright 2016 FanDuel Inc.
```
ayrat555: is this licence required? we already have the Apache Licence in mana

apps/jsonrpc2/README.md
```diff
@@ -0,0 +1,8 @@
+# JSONRPC2
+
+Protocol independent JSONRPC2 server.
```
ayrat555: independent -> implements?

apps/jsonrpc2/test/jsonrpc2/handler_test.exs
```diff
@@ -0,0 +1,240 @@
...
+            ~s([
+                {"jsonrpc": "2.0", "id": "1", "error": {"message": "Internal error", "code": -32603}},
+                {"jsonrpc": "2.0", "id": "2", "error": {"message": "Internal error", "code": -32603}},
+                {"jsonrpc": "2.0", "id": "3", "error": {"message": "Internal error", "code": -32603}}
+              ])
+          )
+        end)
+
+      assert capture =~
+               "[error] Error in handler JSONRPC2.ErrorHandlerTest for method exit with params: []:"
```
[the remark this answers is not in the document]
InoMurko: Yes, the test would fail if you change the format and I think it's desired.

apps/jsonrpc2/lib/jsonrpc2/servers/tcp/protocol.ex
InoMurko: unix domain sockets

InoMurko:
> I just have lots of comments around the JSONRPC2 library. Basically, it's adding a lot of complexity that I don't exactly understand the need for (we're e.g. adding support for TCP line-based communcation, and we're adding JSON-RPC client libraries). I guess I'm torn between: "This will work and supports a variety of backends for faster WebSocket support" and "We could probably cut this down from 2000 lines of code to 500 lines of code using Plug (for just HTTP(s)) that would be significantly easier to grasp.

Happy to hear thoughts here.

Just addressing the comment, Geth and Parity support transport protocols via HTTP, WebSockets and Unix domain sockets. 
HTTP and Websockets are from Cowboy + Plug -> they give us child specs (the only thing we do custom is Cowboy dispatcher pattern, because Websockets aren't supported by plug).
Unix domain sockets are based on ranch (which is how Cowboy does TCP).

All servers (and clients which are implemented only for testing) share the Request and Response modules that do serialization and deserializations. All servers also point to the same handler (which is a custom module that gets installed on the start of the supervision tree) that implements the JSONRPC2 ethereum methods, like web3_clientVersion etc.

Hope this clears a couple of things, otherwise lets continue the debate 👍

apps/jsonrpc2/test/jsonrpc2/handler_test.exs
```diff
@@ -0,0 +1,240 @@
...
+      )
+
+      assert_rpc_noreply(SpecHandlerTest, ~s({"jsonrpc": "2.0", "method": "foobar"}))
+    end
+
+    test "rpc call of non-existent method" do
+      assert_rpc_reply(
+        SpecHandlerTest,
+        ~s({"jsonrpc": "2.0", "method": "foobar", "id": "1"}),
+        ~s({"jsonrpc": "2.0", "error": {"code": -32601, "message": "Method not found", "data": {"method": "foobar", "params": []}}, "id": "1"})
```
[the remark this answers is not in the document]
willmeister: It doesn't matter much, but was thinking more like 
```
  @parser_error_code -32_700
  @invalid_request_code -32_600
  @method_not_found_code -32_601
...
```
The idea is that the test should verify that the error code is received but shouldn't care what the value of the error code is. Since they're not likely to change much, it probably doesn't matter.

apps/jsonrpc2/test/jsonrpc2/handler_test.exs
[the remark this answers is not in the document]
willmeister: Yeah, it might be a bit too in the weeds. The purist in me thinks that if, in the text of the error message, `params` is changed to `parameters`, the functionality still works, so the test should still pass, but it will fail. In the unlikely event that this happens, we can worry about it then.

apps/jsonrpc2/test/jsonrpc2/handler_test.exs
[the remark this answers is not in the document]
InoMurko: https://github.com/mana-ethereum/mana/pull/600/files#diff-c81349a8dc0508cbfcd6768895910d7dR37

apps/jsonrpc2/test/jsonrpc2/handler_test.exs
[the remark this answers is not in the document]
InoMurko: hm... yeah, I know what you mean. I always stick to the convention that says: unit tests should check the output if X is the input. Strict input-output validation.
Acceptance tests on the other hand - I would use your approach. Is the server returning 503 - yes/no thats OK, I'm not interested in the message.

apps/jsonrpc2/lib/jsonrpc2/server/handler.ex
InoMurko: Thats just saying you can implement your handler this way. It doesn’t enforce it.

InoMurko: On the other hand ... I don't think it's un-erlang like, lots of OTP throws exceptions for the simple purpose of control flow.

hayesgm: really? Do you have examples? I just haven't run into that before.

InoMurko: https://github.com/erlang/otp/blob/master/lib/ssl/src/ssl.erl#L1331

InoMurko: Also an interesting comment (it's said it's quoting Virding but who knows), explaining diff between throws and errors: https://stackoverflow.com/a/13618490/10516731

hayesgm: cool, thanks for the example! Still would prefer to avoid this practice, but not enough to say we should change the way this library has been conceived. Carry on. 👍

PR #602, opened 2018-11-13
Queue Extra Data from EIP-8 Frame

hayesgm:
When we receive an EIP-8 frame, the packet specifies a size for the EIP-8 header data, but may contain more data that is unrelated to the EIP-8 data. We previously included the extra packet data as the pre-warm for the ingress header, that caused packets like this to break our handshake mechanism (or more correctly, break the next inbound packet). We now correctly set the ingress mac and queue the rest of the data as we await the next packet. This fixes an issue where Parity connections would often fail.


germsvel:
Great stuff! 🎉

PR #606, opened 2018-11-13
Release versioning

InoMurko:
Needs to be rebased after #600.


MANA_VERSION
```diff
@@ -0,0 +1 @@
+mana/0.0.1
```
ayrat555: can this PR be separated from json rpc?

InoMurko: Sure. It will be after rebase. Hence the tags in progress.

PR #608, opened 2018-11-14
Connect Discovery to Peer-to-Peer

hayesgm:
This patch begins to connect peer discovery to peer to peer connections. This is pretty early going, but I believe it's an important step in the right direction.

A few comments on this PR.

1. If discovery is disabled, we simply connect to the bootnodes.
2. Otherwise, we don't connect to the bootnodes, and instead we ask them for peers. We try to connect to each of those peers.
  a. We should probably keep track of those peers and then choose to connect to individual peers based maybe on distance?
3. When we connect to peers, we try to keep 10 active peers connected (though peers often drop us before we reach this limit).
  a. We currently maintain the count by just connecting to new nodes. We should change this to keep track of a set of available nodes and use more intelligent logic when to connect to those nodes.
4. When we send messages as part of sync, we now send the messages each to a random node to achieve higher parallelism. This is (by no means) a perfect strategy, but it should be a good start. (For example, if one node said it has some block that the other nodes don't know about, we might ask the wrong peer for the block body.)
5. We continue to add config settings to `ExWire.Config` that probably should be re-factored. We opt to wait until we connect CLI before making these updates.

### DIFF
--- a/apps/ex_wire/README.md
+++ b/apps/ex_wire/README.md
@@ -6,7 +6,7 @@ Elixir Client for RLPx, DevP2P and Eth Wire Protocol.
 
 For the time being, this can be used to sync the block chain from a given network (currently defaulted to Ropsten).
 
-You can run `iex -S mix` and you should see:
+You can run `EXT_IP_ADDRESS=$(curl ifconfig.co) iex -S mix` and you should see:
 
 ```
 12:59:23.818 [debug] [Network] [6ce059...1acd9d] Established outbound connection with 13.84.180.240, sending auth.
@@ -33,6 +33,15 @@ You can run `iex -S mix` and you should see:
 
 12:59:24.595 [debug] [Block Queue] Verified block and added to new block tree
 ```
+
+To run against just a local node, run:
+
+Run geth, e.g.:
+`geth --testnet --nat none`
+
+Get the enode address and then run ExWire:
+
+`BOOTNODES=enode://... DISCOVERY=false mix run --no-halt`
 
 In the future, we will continue to grow and built out a proper syncing ability.
 It's likely the proper interface (with CLI tools) will not be this module directly, but instead a general CLI which calls into this module.
--- a/apps/ex_wire/lib/ex_wire.ex
+++ b/apps/ex_wire/lib/ex_wire.ex
@@ -26,11 +26,17 @@ defmodule ExWire do
   defp get_children(_params) do
     chain = ExWire.Config.chain()
 
+    perform_discovery = Config.perform_discovery?()
+
     sync_children =
       if Config.perform_sync?() do
+        # If we're not performing discovery, let's immediately connect
+        # to the bootnodes given. Otherwise, we'll connect to discovered nodes.
+        start_nodes = if perform_discovery, do: [], else: Config.bootnodes()
+
         [
           # Peer supervisor maintains a pool of outbound peers
-          child_spec({PeerSupervisor, Config.bootnodes()}, []),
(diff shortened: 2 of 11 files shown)


apps/ex_wire/lib/ex_wire/p2p/server.ex
```diff
@@ -103,24 +105,22 @@ defmodule ExWire.P2P.Server do
...
-  @spec init(map()) :: {:ok, Connection.t()}
+  @spec init(map()) :: {:ok, state()}
   def init(opts = %{is_outbound: true, peer: peer}) do
-    {:ok, socket} = TCP.connect(peer.host, peer.port)
-
-    :ok =
-      Logger.debug(fn ->
-        "[Network] [#{peer}] Established outbound connection with #{peer.host}."
-      end)
+    Process.send_after(self(), {:connect, opts}, 0)
```
masonforest: curious: why move this into it's own function?

apps/ex_wire/README.md
```diff
@@ -6,7 +6,7 @@ Elixir Client for RLPx, DevP2P and Eth Wire Protocol.
 
 For the time being, this can be used to sync the block chain from a given network (currently defaulted to Ropsten).
 
-You can run `iex -S mix` and you should see:
+You can run `EXT_IP_ADDRESS=$(curl ifconfig.co) iex -S mix` and you should see:
```
MattMSumner: Nice curl 👍

apps/ex_wire/lib/ex_wire.ex
```diff
@@ -26,11 +26,17 @@ defmodule ExWire do
   defp get_children(_params) do
     chain = ExWire.Config.chain()
 
+    perform_discovery = Config.perform_discovery?()
+
     sync_children =
       if Config.perform_sync?() do
+        # If we're not performing discovery, let's immediately connect
```
MattMSumner: Seems like this is clear for the code below. How would you feel about removing this comment?

apps/ex_wire/lib/ex_wire/kademlia/discovery.ex
```diff
@@ -15,6 +16,15 @@ defmodule ExWire.Kademlia.Discovery do
 
     this_round_nodes = RoutingTable.discovery_nodes(table)
 
+    _ =
```
MattMSumner: Since we're not using this variable, how would you feel about removing the assignment? i.e.

```ex
if Config.perform_sync?() do
  for node <- this_round_nodes do
    node
    |> Peer.from_node()
    |> PeerSupervisor.new_peer()
  end
end
```

apps/ex_wire/lib/ex_wire/peer_supervisor.ex
```diff
@@ -56,12 +109,16 @@ defmodule ExWire.PeerSupervisor do
   end
 
   @impl true
-  def init(nodes) do
-    {:ok, _} =
+  def init(peer_enode_urls) do
+    _ =
```
MattMSumner: Same question here about assignment. We could remove this line entirely.

apps/ex_wire/lib/ex_wire/kademlia/discovery.ex
hayesgm: I believe I only added this due to a credo and/or dialyzer check.

apps/ex_wire/lib/ex_wire.ex
hayesgm: I mostly added this because geth/parity, I believe, do this, but never really make it clear. I figured it'd be nice to spell this out.

PR #612, opened 2018-11-15
release versioning proposal

InoMurko:
This is partly addressing the release process by setting the current Mana version in a file for consistency across different applications and the release process (distillery).


apps/ex_wire/config/config.exs
```diff
@@ -20,7 +20,20 @@ config :ex_wire,
     kademlia_process_name: KademliaState,
     supervisor_name: ExWire.NodeDiscoverySupervisor,
     port: 30_304
-  ]
+  ],
+  mana_version:
+    :erlang.apply(
+      fn ->
+        case String.contains?(System.cwd!(), "apps") do
```
hayesgm: could just be `if`?

apps/jsonrpc2/mix.exs
```diff
@@ -45,7 +43,8 @@ defmodule JSONRPC2.Mixfile do
       {:credo, "~> 1.0.0-rc1", only: [:dev, :test], runtime: false},
       {:dialyxir, "~> 1.0.0-rc.4", only: [:dev, :test], runtime: false},
       {:hackney, "~> 1.6", only: [:test]},
-      {:websockex, git: "https://github.com/InoMurko/websockex.git", branch: "master", only: [:test]},
+      {:websockex,
+       git: "https://github.com/InoMurko/websockex.git", branch: "master", only: [:test]},
```
ayrat555: that's bad. why are we using a fork?

InoMurko: because the original has dialyzier issues and referencing old dependencies (cowboy 1, plug 1).
I fixed that and send them a PR https://github.com/Azolo/websockex/pull/63/files

InoMurko: what if we fork mine into mana repo?

apps/ex_wire/config/config.exs
```diff
@@ -20,7 +20,19 @@ config :ex_wire,
     kademlia_process_name: KademliaState,
     supervisor_name: ExWire.NodeDiscoverySupervisor,
     port: 30_304
-  ]
+  ],
+  mana_version:
+    :erlang.apply(
+      fn ->
+        if String.contains?(System.cwd!(), "apps") do
+          true
```
hayesgm: dangling true statement

InoMurko: ouch!

apps/jsonrpc2/mix.exs
ayrat555: @InoMurko it would be nice

InoMurko:
> I dig it. Do want to consider having an umbrella app to deal with a lot of this common configuration stuff? Then it could be mana_version: ManaConfig.version() or something. Thoughts?

That could work. Would we use it for something else too? The other approach would be if we could identify a current application that's at the top of the dependency graph?

ayrat555:
@InoMurko can we set and read mana version in `mix.exs` file

apps/ex_wire/config/config.exs
```diff
@@ -20,7 +20,18 @@ config :ex_wire,
     kademlia_process_name: KademliaState,
     supervisor_name: ExWire.NodeDiscoverySupervisor,
     port: 30_304
-  ]
+  ],
+  mana_version:
+    :erlang.apply(
```
ayrat555: can this be moved to helper function?

apps/jsonrpc2/config/config.exs
```diff
@@ -3,6 +3,17 @@ use Mix.Config
 config :jsonrpc2,
   ipc: [enabled: true, path: Enum.join([System.user_home!(), "/.ethereum", "/mana.ipc"])],
   http: [enabled: true, port: 4000],
-  ws: [enabled: true, port: 4000]
+  ws: [enabled: true, port: 4000],
+  mana_version:
+    :erlang.apply(
```
ayrat555: duplication with above

rel/config.exs
```diff
@@ -46,16 +45,32 @@ end
...
-    evm: :permanent,
-    ex_wire: :permanent,
-    exth_crypto: :permanent,
-    merkle_patricia_tree: :permanent,
-    jsonrpc2: :permanent
-  ]
-end
+  set(
+    version:
+      :erlang.apply(
```
ayrat555: again

PR #613, opened 2018-11-15
Fix Peer Sync

hayesgm:
This patch fixes peer sync (syncing against a single node). Previously, once we pulled 100 blocks, we would just stop and do nothing. Now, we correctly continue syncing whenever we have no blocks queued up. While we should be smarter about which blocks to sync and when to ask for more and from which peers, we'll probably end up switching to fast or warp sync before long, which might render this moot. With this patch, syncing against a local parity node works very well (and so does geth before it drops us for not having fast sync enabled).

### DIFF
--- a/apps/ex_wire/lib/ex_wire/sync.ex
+++ b/apps/ex_wire/lib/ex_wire/sync.ex
@@ -203,18 +203,13 @@ defmodule ExWire.Sync do
         :random
       )
 
-    next_trie =
-      maybe_request_new_block_or_save(
-        block_tree,
-        next_block_tree,
-        next_trie,
-        next_block_queue
-      )
+    next_maybe_saved_trie = maybe_save(block_tree, next_block_tree, next_trie)
+    :ok = maybe_request_next_block(next_block_queue)
 
     state
     |> Map.put(:block_queue, next_block_queue)
     |> Map.put(:block_tree, next_block_tree)
-    |> Map.put(:trie, next_trie)
+    |> Map.put(:trie, next_maybe_saved_trie)
   end
 
   @doc """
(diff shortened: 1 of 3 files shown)


apps/ex_wire/lib/ex_wire/sync.ex
```diff
@@ -237,29 +242,27 @@ defmodule ExWire.Sync do
...
 
   @doc """
   Determines the next block we don't yet have in our blocktree and
   dispatches a request to all connected peers for that block and the
   next `n` blocks after it.
   """
-  @spec request_next_block(BlockQueue.t(), Blocktree.t()) :: integer()
-  def request_next_block(block_queue, block_tree) do
+  @spec get_next_block_to_request(BlockQueue.t(), Blocktree.t()) :: integer()
+  def get_next_block_to_request(block_queue, block_tree) do
```
InoMurko: is this needed public?

apps/ex_wire/lib/ex_wire/sync.ex
```diff
@@ -281,23 +284,26 @@ defmodule ExWire.Sync do
...
-        ) :: {integer(), Trie.t()}
+          BlockQueue.t()
+        ) :: Trie.t()
   defp maybe_request_new_block_or_save(
          block_tree,
          next_block_tree,
          trie,
-         block_queue,
-         last_requested_block
+         block_queue
```
InoMurko: I try to avoid nested statements. What if you pattern match on empty queue or matching block tree?

InoMurko:
two suggestions, but looks good

apps/ex_wire/lib/ex_wire/sync.ex
```diff
@@ -140,7 +149,7 @@ defmodule ExWire.Sync do
         Otherwise, we request the block bodies for the blocks we don't
         know about.
 
-  Note: we process blocks in memroy and save our state tree every so often.
+  Note: we process blocks in memory and save our state tree every so often.
```
MattMSumner: Nice catch.

apps/ex_wire/lib/ex_wire/sync.ex
```diff
@@ -281,23 +284,26 @@ defmodule ExWire.Sync do
...
          block_tree,
          next_block_tree,
          trie,
-         block_queue,
-         last_requested_block
+         block_queue
        ) do
-    # We can make this better, but it's basically "if we change, request another block"
     if block_tree != next_block_tree do
+      # Let's consider saving the trie if it's changed
```
MattMSumner: How would you feel about removing this comment since the function name seems sufficient?

apps/ex_wire/lib/ex_wire/sync.ex
```diff
@@ -281,23 +284,26 @@ defmodule ExWire.Sync do
           Blocktree.t(),
           Blocktree.t(),
           Trie.t(),
-          BlockQueue.t(),
-          integer()
-        ) :: {integer(), Trie.t()}
+          BlockQueue.t()
+        ) :: Trie.t()
   defp maybe_request_new_block_or_save(
```
germsvel: I know this is not part of the changes, but would it make sense to break this into two functions? Maybe that would help avoid the nested `if` statements?

hayesgm: Yup, moved to two functions.

PR #618, opened 2018-11-15
cache initial account storage value

ayrat555:
We were always reading initial storage from the permanent Merkle Trie storage at least three times for `sstore` operations:
- Gas calculation
- Refund calculation
- Operation itself

fixes https://github.com/mana-ethereum/mana/issues/615


apps/blockchain/lib/blockchain/account/repo.ex
```diff
@@ -377,27 +377,24 @@ defmodule Blockchain.Account.Repo do
...
-          case stored_value do
-            :account_not_found ->
-              if account, do: :key_not_found, else: :account_not_found
-
-            stored_value ->
-              stored_value
-          end
-
-        {updated_repo, found_value}
+        cache_and_get_initial_storage_value(account_repo, address, key)
```
InoMurko: you could move the case out to a separate function

PR #620, opened 2018-11-15
CLI with release

InoMurko:
change cookie every release, make cli work with release, debug flags that enable erlang distribution


apps/cli/lib/mix/tasks/sync.ex
```diff
@@ -23,30 +23,77 @@ defmodule Mix.Tasks.Sync do
       mix sync --chain ropsten --provider-url https://ropsten.infura.io
 
       mix sync --chain ropsten --provider-url ipc:///path/to/file
+
+      or release
+
+      mana sync --chain ropsten --provider-url https://ropsten.infura.io --
```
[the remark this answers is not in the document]
InoMurko: The first one has mix starting the code. The other is release mana.

apps/cli/mix.exs
```diff
@@ -31,9 +31,9 @@ defmodule Cli.MixProject do
   defp deps do
     [
       # Umbrella deps
-      {:blockchain, in_umbrella: true},
-      {:ex_wire, in_umbrella: true, runtime: false},
-      {:merkle_patricia_tree, in_umbrella: true},
+      # {:blockchain, in_umbrella: true},
```
InoMurko: @hayesgm ^^

apps/ex_wire/lib/ex_wire/config.ex
```diff
@@ -25,6 +25,23 @@ defmodule ExWire.Config do
...
           | :public_ip
           | :sync
+          | :db_root
+  @doc """
+  Allows application to configure ExWire before it starts.
+  """
+  @spec configure!(keyword(env_keys)) :: :ok | no_return()
+  def configure!(kw_args) do
+    if Enum.member?(Application.started_applications(), :ex_wire) do
+      ##you could do something like:
```
InoMurko: @hayesgm

apps/ex_wire/mix.exs
```diff
@@ -34,6 +34,7 @@ defmodule ExWire.Mixfile do
     [
       {:ex_doc, "~> 0.16", only: :dev, runtime: false},
       {:ex_rlp, "~> 0.3.1"},
+      {:cli, in_umbrella: true},
```
InoMurko: though CLI is technically not an application. it doesn't have an application module.

rel/config.exs
```diff
@@ -22,15 +22,39 @@ use Mix.Releases.Config,
...
-  # If you are running Phoenix, you should make sure that
-  # server: true is set and the code reloader is disabled,
-  # even in dev mode.
-  # It is recommended that you build with MIX_ENV=prod and pass
-  # the --env flag to Distillery explicitly if you want to use
-  # dev mode.
+  set(
+    commands: [
+      # we can replace sync with whatever, 'run' too for example
+      sync: "rel/commands/sync"
```
InoMurko: @hayesgm ^^

rel/config.exs
```diff
@@ -22,15 +22,39 @@ use Mix.Releases.Config,
...
+    ]
+  )
+
   set(dev_mode: true)
   set(include_erts: true)
-  set(cookie: :"bAFS)[qmAg0XFS5`7Ncbz,$J:K3Tv)t@{OIPWJb`;N!k%z^H,^(L}:qI_$Tp8>EX")
+
+  set(
+    cookie:
+      :erlang.apply(
```
InoMurko: setting a cookie like this allows us to attach to a running node

apps/cli/config/config.exs
```diff
@@ -2,29 +2,18 @@
...
-# file won't be loaded nor affect the parent project. For this reason,
-# if you want to provide default values for your application for
-# 3rd-party users, it should be done in your "mix.exs" file.
+config :cli,
+  cookie:
+    :erlang.apply(
+      fn ->
+        cookie =
+          if String.contains?(System.cwd!(), "apps") do
+            String.trim(File.read!(Enum.join(["../../", "COOKIE"])))
```
germsvel: I think you could use `Path.join` instead of `Enum.join` here. Not sure if it would make a big difference, but it might make more sense since we're dealing with paths?

apps/cli/config/config.exs
```diff
@@ -2,29 +2,18 @@
...
 use Mix.Config
 
-# This configuration is loaded before any dependency and is restricted
-# to this project. If another project depends on this project, this
-# file won't be loaded nor affect the parent project. For this reason,
-# if you want to provide default values for your application for
-# 3rd-party users, it should be done in your "mix.exs" file.
+config :cli,
+  cookie:
+    :erlang.apply(
```
germsvel: Just curious, is there a reason why we can't use elixir's [`apply/2`](https://hexdocs.pm/elixir/Kernel.html#apply/2) function here instead of `:erlang.apply/2`?

apps/cli/lib/cli/config.ex
```diff
@@ -7,14 +7,17 @@ defmodule CLI.Config do
...
       'db/mana-ropsten'
   """
   @spec db_name(atom()) :: nonempty_charlist()
   def db_name(chain_id) do
-    'db/mana-' ++ Atom.to_charlist(chain_id)
+    path = "#{System.cwd!()}/db/mana-#{Atom.to_string(chain_id)}"
+    dirname = Path.dirname(path)
+    :ok = File.mkdir_p(dirname)
+
+    "#{System.cwd!()}/db/mana-#{Atom.to_string(chain_id)}"
```
germsvel: Can we reuse the `path` variable from line 16 here?

apps/cli/lib/cli/state_supervisor.ex
```diff
@@ -0,0 +1,18 @@
...
+  @moduledoc """
+  Supervising CLI State and creating the ETS table that would hold chain changes for JSONRPC.
+  """
+  def start_link(chain) do
+    Supervisor.start_link(__MODULE__, chain)
+  end
+
+  def init(chain) do
+    :cli = :ets.new(:cli, [:set, :public, :named_table])
+    children = [worker(State, [chain])]
```
germsvel: Should we use the new format for specifications? Since `State` is a gen server, I think this would work ->`{State, [chain]}`

apps/cli/config/config.exs
InoMurko: technically it's erlang's apply either way, but I get your point

apps/cli/lib/cli/state_supervisor.ex
InoMurko: tuple childs are new?

apps/cli/lib/cli/config.ex
InoMurko: good point

apps/cli/lib/cli/state_supervisor.ex
germsvel: Not extremely new, but they came in [v1.5](https://elixir-lang.org/blog/2017/07/25/elixir-v1-5-0-released/#streamlined-child-specs)

InoMurko: aha! was not aware of these new api changes, good point! fixed.

apps/ex_wire/lib/ex_wire/config.ex
```diff
@@ -25,6 +25,31 @@ defmodule ExWire.Config do
...
+
+  @type configure :: [{env_keys, any()}]
+
+  @doc """
+  Allows application to configure ExWire before it starts.
+  """
+  @spec configure!(configure) :: [any()] | no_return()
+  def configure!(kw_args) do
+    if Enum.member?(Application.started_applications(), :ex_wire) do
+      ## you could do something like:
```
InoMurko: @hayesgm

PR #626, opened 2018-11-15
Generate dataset for ethash algorithm

germsvel:
This is part of #541

Description
=====

We implement the dataset generation for the Ethash algorithm outlined in Appendix J, section J.3.3, of the Yellow Paper. The algorithm combines data from 256 (J_parents) pseudorandomly selected cache nodes for each.

The algorithm implemented more closely resembles that of the [wiki]. Instead of having a `mix` and `parents` functions, it initializes the mix for each dataset item and then iterates over it 256 times computing the FNV algorithm on it.

It's worth noting that the FNV algorithm implemented here expects unsigned 32 bit integers from little-endian binaries. It is for that reason that our implementation frequently decodes and encodes unsigned integers in little-endian format, and it is also why we transform some binaries into lists of unsigned 32 bit integers.

[wiki]: https://github.com/ethereum/wiki/wiki/Ethash#full-dataset-calculation


MattMSumner:
@ayrat555 this isn't used yet. Not until we implement the Proof-of-work function defined in appendix J.4 of the yellow paper. Once we have that defined we can start using that to verify that miners have produces a valid nonce.

PR #627, opened 2018-11-15
[WIP] Hook up Mana CLI to Release

hayesgm:
This patch adds a Mana CLI and make it available from a release script. The goal of this PR to be able to run:

* `./mana`
* `./mana --no-discovery --bootnodes=enode://...`
* `./mana --no-sync`

And one day:

* `./mana --json-rpc`

Currently, there's an issue parsing args, that everything works as expected, except when `Mix.Tasks.Mana.run/1` gets called, it is missing the `--` params. So if I call the script with `rel/commands/mana.sh a b c` I see `run(["a", "b", "c"])`, but if I call it with `rel/commands/mana.sh --a --b --c`, I get `run(["--a"])`. I assume something that's parsing is deciding the `--<arg>`s are meant for it, not for our script, but I can't figure that out. It might be related to running `elixir -e '...' --argsv`, which seems to be the command called by the release script.


apps/cli/lib/mix/tasks/mana.ex
```diff
@@ -0,0 +1,51 @@
...
+          chain: chain_name,
+          sync: sync,
+          discovery: discovery,
+          bootnodes: bootnodes
+        )
+
+        Application.ensure_all_started(:ex_wire)
+
+        # No Halt
+        Process.sleep(:infinity)
```
ayrat555: For what purpose `Process.sleep` is used here?

from its docs:
```
Use this function with extreme care. 
For almost all situations where you would use sleep/1 in Elixir, there is likely a more correct, faster and precise way of achieving the same with message passing.
```

apps/ex_wire/lib/ex_wire/config.ex
```diff
@@ -128,8 +143,7 @@ defmodule ExWire.Config do
 
   @spec client_id() :: String.t()
   def client_id() do
-    version = Mix.Project.config()[:version]
-    "mana/#{version}"
+    "mana/0.1"
```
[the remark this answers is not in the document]
hayesgm: Yes, I'm waiting on #620 for that

README.md
```diff
@@ -58,6 +58,16 @@ Once syncing begins you will see a timestamp and a running list of verified bloc
 [command-line interface]: https://github.com/mana-ethereum/mana/tree/master/apps/cli
 [Infura API key here]: https://infura.io/register
 
+### Releases
+
+To build a release, run: `mix release`, which will build a release in `_build/dev/rel/mana/bin/mana`.
+
+Then you can run:
+
+`_build/dev/rel/mana/bin/mana run --no-discovery --bootnodes enode://...`
```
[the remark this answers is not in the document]
hayesgm: Yup, `mix mana --no-discovery --bootnodes enode://...` works identically.

apps/cli/lib/mix/tasks/mana.ex
hayesgm: It's just so the process doesn't exit since it otherwise finishes the only function it runs (but it started other supervised applications, so it still is being productive). I took this line from here: https://github.com/elixir-lang/elixir/blob/master/lib/mix/lib/mix/tasks/run.ex#L87

hayesgm:
@InoMurko Try it for `Mix.Tasks.Mana.run/1` maybe? Based on https://github.com/bitwalker/artificery/issues/3, I'm pretty certain there's a bug, since the code seems wrong, or more specifically, to ignore this case.

apps/cli/lib/parser/mana_parser.ex
```diff
@@ -0,0 +1,133 @@
...
+           }}
+          | {:error, String.t()}
+  def mana_args(args) do
+    {kw_args, _extra} =
+      OptionParser.parse!(args,
+        switches: [chain: :string, no_discovery: :boolean, no_sync: :boolean, bootnodes: :string]
+      )
+
+    with {:ok, chain_name} <- get_chain_name(kw_args) do
+      with {:ok, discovery} <- get_discovery(kw_args) do
```
InoMurko: why so many withs?

MattMSumner: I think this could be written as:

```ex
with {:ok, chain_name} <- get_chain_name(kw_args),
     {:ok, discovery} <- get_discovery(kw_args),
     {:ok, sync} <- get_sync(kw_args),
     {:ok, bootnodes} <- get_bootnodes(kw_args) do
     {
       :ok,
       %{
         chain_name: chain_name,
         discovery: discovery,
         sync: sync,
         bootnodes: bootnodes
       }
    }
end
```

If any of the first 4 lines return a non `{:ok, _}` result then it will short circuit and return the "errored" value. If all successfully return an `{:ok, _)` value then the final result will be the `do`.

apps/cli/test/parser_test.exs
```diff
@@ -1,4 +0,0 @@
-defmodule CLI.ParserTest do
```
MattMSumner: Should this be moved instead of deleted so we run the SyncParser doctests?

rel/commands/mana.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+release_ctl eval --mfa "Mix.Tasks.Mana.run/1" --argv -- --extra "$@"
```
MattMSumner: Looks like this file is missing a trailing newline. Here's a blog post on why GitHub things this is relevant https://robots.thoughtbot.com/no-newline-at-end-of-file

apps/cli/lib/mix/tasks/mana.ex
```diff
@@ -0,0 +1,51 @@
+defmodule Mix.Tasks.Mana do
+  @moduledoc """
+  Main entry-point for the mana CLI.
+
+  Command Line Options:
+
+      `--chain` - Chain to load data from (default: ropsten)
```
germsvel: Is the default `ropsten` for the time being, or is this the idea as we go forward? Would mainnet be a better default?

apps/cli/lib/mix/tasks/mana.ex
germsvel: Not sure if we could do something like Phoenix does when it starts a [web server], defaulting the "--no-halt" flag. Not sure what modifications we'd have to do in order to do that, but just throwing another option out there (although it seems from what you linked to, that's what --no-halt does underneath the hood).

rel/config.exs
```diff
@@ -52,10 +64,24 @@ release :mana do
     blockchain: :permanent,
     cli: :permanent,
     evm: :permanent,
-    ex_wire: :permanent,
+    ex_wire: :load,
```
InoMurko: I don't think load is needed.

apps/ex_wire/lib/ex_wire/config.ex
```diff
@@ -128,8 +146,7 @@ defmodule ExWire.Config do
 
   @spec client_id() :: String.t()
   def client_id() do
-    version = Mix.Project.config()[:version]
-    "mana/#{version}"
+    "mana/#{get_env([], :mana_version)}"
```
InoMurko: is having a slash in the name ok? I wonder if it would interfere with directory path

apps/cli/lib/parser/mana_parser.ex
```diff
@@ -0,0 +1,155 @@
...
+          debug: :boolean
+        ]
+      )
+
+    with {:ok, chain_name} <- get_chain_name(kw_args) do
+      with {:ok, discovery} <- get_discovery(kw_args) do
+        with {:ok, sync} <- get_sync(kw_args) do
+          with {:ok, bootnodes} <- get_bootnodes(kw_args) do
+            with {:ok, debug} <- get_debug(kw_args) do
+              {:ok,
```
InoMurko: there's no need to nest with's this way, one is enough if statements are separated with commas

rel/commands/run
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+release_ctl eval --mfa "Mix.Tasks.Mana.run/1" --argv -- --extra "$@"
```
InoMurko: I think we really need to start testing the release so that we can avoid implementation discrepancies...

apps/ex_wire/lib/ex_wire/config.ex
hayesgm: Believe so, Parity is: `client_id: "Parity-Ethereum/v2.1.3-beta-18ddd7c-20181015/x86_64-macos/rustc1.29.0"`

apps/cli/lib/parser/mana_parser.ex
hayesgm: yeah, that's way better, thanks.

apps/cli/lib/mix/tasks/mana.ex
hayesgm: Yes, I didn't actually know what chain to use for mainnet, so that's actually why. But I think as we get closer to launch, we can start to clean up these types of defaults to more sensible values.

rel/commands/run
hayesgm: Yeah. We could even do a very simple test after build like `mana self-diagnose` that runs some general tests and makes sure everything looks in line. That would be useful for users of mana, as well.

rel/config.exs
hayesgm: When I had load permanent before, I think it started the application before I had a chance to configure it.

hayesgm: looks like you're right, doesn't start up without starting it ourselves even if it's permanent. fixed.

apps/cli/config/config.exs
```diff
@@ -2,6 +2,39 @@
...
+      |> File.read()
+
+    case cookie_resp do
+      {:ok, cookie} ->
+        cookie
+        |> String.trim()
+        |> :erlang.binary_to_atom(:utf8)
+
+      _ ->
+        nil
```
InoMurko: should we throw a fit here? actually, I would like to upgrade cookie handling after this gets in. One needs to be able to replace the cookie,...

apps/cli/lib/mix/tasks/mana.ex
InoMurko: We would probably need to add something like a deamon flag as well...

hayesgm:
This covers the case were the CLI is run as non-distributed Erlang.

apps/cli/lib/parser/mana_parser.ex
```diff
@@ -0,0 +1,151 @@
...
+  @default_no_sync false
+  @default_bootnodes "from_chain"
+  @default_debug false
+
+  @doc """
+  Parsers args for syncing
+
+  ## Options:
+    * `--chain` - Chain to load data from (default: ropsten)
+    * `--no-discovery` - Perform discovery (default: false)
```
InoMurko: I think this isn't valid anymore

hayesgm: hm, I think those are accurate, the other comments below in `tasks/mana.ex` were incorrect. I've updated the docs there.

apps/cli/lib/parser/mana_parser.ex
```diff
@@ -0,0 +1,151 @@
...
+      }}
+
+      iex> CLI.Parser.ManaParser.mana_args(["--chain", "pony"])
+      {:error, "Invalid chain: pony"}
+  """
+  @spec mana_args([String.t()]) ::
+          {:ok,
+           %{
+             chain_name: atom(),
+             discovery: boolean(),
```
InoMurko: I guess there'a mixup now because I removed the negation flags.

PR #628, opened 2018-11-16
cache not found accounts

ayrat555:
fixes https://github.com/mana-ethereum/mana/issues/616

### DIFF
--- a/apps/blockchain/test/blockchain/account/repo_test.exs
+++ b/apps/blockchain/test/blockchain/account/repo_test.exs
@@ -350,5 +350,17 @@ defmodule Blockchain.Account.RepoTest do
                storage_root: Account.empty_trie()
              }
     end
+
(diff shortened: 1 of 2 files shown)


apps/blockchain/test/blockchain/account/repo_test.exs
```diff
@@ -350,5 +350,17 @@ defmodule Blockchain.Account.RepoTest do
                storage_root: Account.empty_trie()
              }
     end
+
+    test "caches not found account in the storage", %{state: state} do
+      address = <<999>>
```
germsvel: I've seen us do this in tests before, and I wonder if it would be good to have the address be at least the same size as a real address (I think it's something like `<<_::256>>`, right?). I believe it has led to dialyzer issues before. What do you think about changing that?

PR #629, opened 2018-11-16
cache account code

ayrat555:


apps/blockchain/test/blockchain/account/repo_test.exs
```diff
@@ -184,6 +184,32 @@ defmodule Blockchain.Account.RepoTest do
 
       assert found_code == code
     end
+
+    test "caches machine code and then always reads it from cache", %{state: state} do
+      address = <<1>>
```
germsvel: What do you think about using an address that is the same size as the ones on real accounts? I think it's something like `<<1::256>>`, right?

apps/blockchain/test/blockchain/account/repo_test.exs
```diff
@@ -184,6 +184,32 @@ defmodule Blockchain.Account.RepoTest do
...
+      code = <<5>>
+
+      {updated_repo, {:ok, found_code}} =
+        state
+        |> Account.reset_account(address)
```
germsvel: Just for my own understanding, why do we have to reset the account here? Does that do something with the cache?

ayrat555: @germsvel it creates blank account in the database

PR #632, opened 2018-11-19
Add Warp Sync

hayesgm:
This patch adds Warp Sync to Mana. Note: this is not fully stable yet, as there are a number of issues which can cause the sync to fail, but generally, the entire sync process is fully implemented in this patch. Sync works by returning the information about each block for the last several blocks, building a state tree from that, and then getting every single account state in chunks for the most recent block. After all of those are synced, you have the complete state going forward and can, in the background, sync the earlier parts of the chain.

    We generally code up the following modules:

    1. `ExWire.Sync` - GenServer responsible for warp and non-warp sync. This module contains the logic of how to communciate with remote peers and what to do in response to packets they send.
    2. `ExWire.WarpQueue` - Generally contains the current state of a warp (and we persist it during a warp so we can pick up where we left off). Keeps track of which block and state chunks we've processed so far and which we need to process going forward.
    3. `ExWire.WarpProcessor` - Keeps track of state and block chunks that haven't been processed and kicks off processing parallel. The parallel processing of state and block chunks is paramount to getting warp to run quickly.
    4. `ExWire.WarpProcessor.PowProcess` - The actual logic of how to process the state and block chunks for a warp. We process transaction receipts, store things to our trie, etc.

The good news is: our processing is relatively fast. We process (after a lot of optimization) around 3,000 accounts per second, which means that Ropsten with 20MM accounts takes a little over an hour to warp (I believe it's 30-40 minutes on Parity for Ropsten). These are early promising results.

The bad news is: we don't always match on the state or block trie after completing a warp, so there's still work to do. I generally sync Ellaism since it's small and easy to test with.

Also bad news: PoA chains in Parity have a different variety of warp messages, so we'll need to process them separately to support Kovan, etc.

Also bad news: Geth doesn't support warp, so we'll need to build fast sync, as well.

Finally, we didn't implement snapshotting ourselves, which we'll need to do to be a good warp peer.

References: 
  * Parity's [Warp Sync Docs](https://wiki.parity.io/Warp-Sync))
  * Parity's [Warp Sync Format](https://wiki.parity.io/Warp-Sync-Snapshot-Format.html)
  * Parity's [source for PoA sync format](https://github.com/paritytech/parity-ethereum/blob/master/ethcore/src/snapshot/consensus/authority.rs#L43)

FYI, to run the sync (you'll need to merge the other related PRs), but I copy my peer id for mana into `~/peer` and then run from mana:

`mix compile && TRACE=true iex -S mix mana --chain ropsten --no-discovery --bootnodes enode://2e0b03e3633c0a6b84da62f998c6ed9d3c05ab253a745abf0dd3866d3d8d870f3ba03f416cbe2ce28335dd7dff32c6733b86f714128cadafd479aabd95ccbe9b@127.0.0.1:30303 --warp`

For parity, I then run:

`parity --chain ropsten --no-discovery --reserved-peers ~/peer --reserved-only --logging sync,warp,network=trace`

This gives you great logs with very little noise.

### DIFF
--- a/apps/blockchain/lib/blockchain/block_setter.ex
+++ b/apps/blockchain/lib/blockchain/block_setter.ex
@@ -18,12 +18,18 @@ defmodule Blockchain.BlockSetter do
 
       iex> Blockchain.Block.set_block_number(%Blockchain.Block{header: %Block.Header{extra_data: "hello"}}, %Blockchain.Block{header: %Block.Header{number: 32}})
       %Blockchain.Block{header: %Block.Header{number: 33, extra_data: "hello"}}
+
+      iex> Blockchain.Block.set_block_number(%Blockchain.Block{header: %Block.Header{extra_data: "hello"}}, %Blockchain.Block{header: %Block.Header{number: nil}})
+      %Blockchain.Block{header: %Block.Header{number: nil, extra_data: "hello"}}
   """
   @spec set_block_number(Block.t(), Block.t()) :: Block.t()
-  def set_block_number(block, parent_block) do
-    number = parent_block.header.number + 1
-    header = %{block.header | number: number}
-    %{block | header: header}
+  def set_block_number(block, %Block{header: %Header{number: nil}}) do
+    %{block | header: %{block.header | number: nil}}
+  end
+
+  def set_block_number(block, %Block{header: %Header{number: parent_block_number}})
+      when is_integer(parent_block_number) do
+    %{block | header: %{block.header | number: parent_block_number + 1}}
   end
 
   @doc """
--- a/apps/blockchain/lib/blockchain/blocktree.ex
+++ b/apps/blockchain/lib/blockchain/blocktree.ex
@@ -67,7 +67,7 @@ defmodule Blockchain.Blocktree do
   end
 
   @spec update_best_block(t, Block.t()) :: t
-  defp update_best_block(blocktree, block) do
+  def update_best_block(blocktree, block) do
     best_block = blocktree.best_block
 
     new_best_block =
@@ -77,7 +77,15 @@ defmodule Blockchain.Blocktree do
          do: block,
          else: best_block
 
-    %{blocktree | best_block: new_best_block}
+    # Make sure block is stored with a block hash
+    new_best_block_with_hash =
+      if new_best_block.block_hash == nil do
+        %{new_best_block | block_hash: Block.hash(new_best_block)}
+      else
+        new_best_block
+      end
+
+    %{blocktree | best_block: new_best_block_with_hash}
   end
 
   @doc """
--- a/apps/cli/lib/parser/mana_parser.ex
+++ b/apps/cli/lib/parser/mana_parser.ex
@@ -7,6 +7,7 @@ defmodule CLI.Parser.ManaParser do
   @default_no_discovery false
   @default_no_sync false
   @default_bootnodes "from_chain"
+  @default_warp false
   @default_debug false
 
   @doc """
@@ -17,6 +18,7 @@ defmodule CLI.Parser.ManaParser do
     * `--no-discovery` - Perform discovery (default: false)
     * `--no-sync` - Perform syncing (default: false)
     * `--bootnodes` - Comma separated list of bootnodes (default: from_chain)
+    * `--warp` - Perform warp sync (default: false)
     * `--debug` - Add remote debugging (default: false)
 
   ## Examples
@@ -27,15 +29,17 @@ defmodule CLI.Parser.ManaParser do
         discovery: false,
         sync: false,
         bootnodes: :from_chain,
+        warp: false,
         debug: false
       }}
 
-      iex> CLI.Parser.ManaParser.mana_args(["--chain", "foundation", "--bootnodes", "enode://google.com,enode://apple.com", "--debug"])
+      iex> CLI.Parser.ManaParser.mana_args(["--chain", "foundation", "--bootnodes", "enode://google.com,enode://apple.com", "--warp", "--debug"])
       {:ok, %{
         chain_name: :foundation,
         discovery: true,
         sync: true,
         bootnodes: ["enode://google.com", "enode://apple.com"],
+        warp: true,
         debug: true
       }}
 
@@ -45,6 +49,7 @@ defmodule CLI.Parser.ManaParser do
         discovery: true,
         sync: true,
         bootnodes: :from_chain,
+        warp: false,
         debug: false
       }}
 
@@ -58,6 +63,7 @@ defmodule CLI.Parser.ManaParser do
              discovery: boolean(),
              sync: boolean(),
              bootnodes: :from_chain | list(String.t()),
+             warp: boolean(),
              debug: boolean()
            }}
           | {:error, String.t()}
@@ -69,6 +75,7 @@ defmodule CLI.Parser.ManaParser do
           no_discovery: :boolean,
           no_sync: :boolean,
           bootnodes: :string,
+          warp: :boolean,
           debug: :boolean
         ]
       )
@@ -77,6 +84,7 @@ defmodule CLI.Parser.ManaParser do
          {:ok, discovery} <- get_discovery(kw_args),
          {:ok, sync} <- get_sync(kw_args),
          {:ok, bootnodes} <- get_bootnodes(kw_args),
+         {:ok, warp} <- get_warp(kw_args),
          {:ok, debug} <- get_debug(kw_args) do
       {:ok,
        %{
@@ -84,6 +92,7 @@ defmodule CLI.Parser.ManaParser do
          discovery: discovery,
          sync: sync,
          bootnodes: bootnodes,
+         warp: warp,
          debug: debug
        }}
     end
@@ -138,6 +147,15 @@ defmodule CLI.Parser.ManaParser do
       _ ->
         {:ok, String.split(given_bootnodes, ",")}
     end
+  end
+
+  @spec get_warp(warp: boolean()) :: {:ok, boolean()} | {:error, String.t()}
+  defp get_warp(kw_args) do
+    given_warp =
+      kw_args
+      |> Keyword.get(:warp, @default_warp)
+
+    {:ok, given_warp}
   end
 
   @spec get_debug(debug: boolean()) :: {:ok, boolean()} | {:error, String.t()}
--- a/apps/ex_wire/config/config.exs
+++ b/apps/ex_wire/config/config.exs
@@ -27,10 +27,7 @@ mana_version =
 config :ex_wire,
   p2p_version: 0x04,
   protocol_version: 63,
-  # ropsten
-  network_id: 3,
-  caps: [{"eth", 62}, {"eth", 63}],
-  chain: :ropsten,
+  caps: [{"eth", 62}, {"eth", 63}, {"par", 1}],
   # TODO: This should be set and stored in a file
   private_key: :random,
   bootnodes: :from_chain,
@@ -43,7 +40,8 @@ config :ex_wire,
     port: 30_304
   ],
   db_root: db_root,
-  mana_version: mana_version
+  mana_version: mana_version,
+  warp: false
 
 config :ex_wire, :environment, Mix.env()
 import_config "#{Mix.env()}.exs"
--- a/apps/ex_wire/config/dev.exs
+++ b/apps/ex_wire/config/dev.exs
@@ -2,8 +2,8 @@ use Mix.Config
 
 config :ex_wire,
   network_adapter: {ExWire.Adapter.UDP, NetworkClient},
-  sync: true,
-  discovery: true,
+  sync: false,
+  discovery: false,
   private_key:
     <<10, 122, 189, 137, 166, 190, 127, 238, 229, 16, 211, 182, 104, 78, 138, 37, 146, 116, 90,
       68, 76, 86, 168, 24, 200, 155, 0, 99, 58, 226, 211, 30>>,
--- a/apps/ex_wire/lib/ex_wire/p2p/server.ex
+++ b/apps/ex_wire/lib/ex_wire/p2p/server.ex
@@ -110,7 +110,7 @@ defmodule ExWire.P2P.Server do
   """
   @spec get_peer(pid()) :: Peer.t()
   def get_peer(pid) do
-    GenServer.call(pid, :get_peer)
+    GenServer.call(pid, :get_peer, :infinity)
   end
 
   @doc """
--- a/apps/ex_wire/lib/ex_wire/struct/warp_queue.ex
+++ b/apps/ex_wire/lib/ex_wire/struct/warp_queue.ex
@@ -1,0 +1,302 @@
+defmodule ExWire.Struct.WarpQueue do
+  @moduledoc """
+  `WarpQueue` maintains the current state of an active warp, this mean we will 
+  track the `block_chunk` hashes and `state_chunk` hashes given to us, so we
+  can request each from our connected peers. This structure is also persisted
+  during a warp sync, so that if interrupted, we can resume a warp where we
+  left off.
+
+  TODO: This will likely need to be updated to handle warping from more than
+        one direct peer.
+  """
+  require Logger
+
+  alias Blockchain.{Block, Blocktree}
+  alias Exth.Time
+  alias ExWire.Packet.Capability.Par.SnapshotManifest
+  alias MerklePatriciaTree.Trie
+
+  @type t :: %__MODULE__{
+          manifest: SnapshotManifest.manifest() | nil,
+          manifest_hashes: MapSet.t(EVM.hash()),
+          manifest_block_hashes: MapSet.t(EVM.hash()),
+          manifest_state_hashes: MapSet.t(EVM.hash()),
+          chunk_requests: MapSet.t(EVM.hash()),
+          retrieved_chunks: MapSet.t(EVM.hash()),
+          processed_chunks: MapSet.t(EVM.hash()),
+          processed_blocks: MapSet.t(integer()),
+          processed_accounts: integer(),
+          warp_start: Time.time(),
+          block_tree: Blocktree.t(),
+          state_root: EVM.hash()
+        }
+
+  defstruct [
+    :manifest,
+    :manifest_hashes,
+    :manifest_block_hashes,
+    :manifest_state_hashes,
+    :chunk_requests,
+    :retrieved_chunks,
+    :processed_chunks,
+    :processed_blocks,
+    :processed_accounts,
+    :warp_start,
+    :block_tree,
+    :state_root
+  ]
+
+  @empty_trie Trie.empty_trie_root_hash()
+
+  @doc """
+  Creates a new `WarpQueue`.
+  """
+  def new() do
+    %__MODULE__{
+      manifest: nil,
+      manifest_hashes: MapSet.new(),
+      manifest_block_hashes: MapSet.new(),
+      manifest_state_hashes: MapSet.new(),
+      chunk_requests: MapSet.new(),
+      retrieved_chunks: MapSet.new(),
+      processed_chunks: MapSet.new(),
+      processed_blocks: MapSet.new(),
+      processed_accounts: 0,
+      warp_start: Time.time_start(),
+      block_tree: Blocktree.new_tree(),
+      state_root: @empty_trie
+    }
+  end
+
+  @doc """
+  Handle receiving a new manifest from a peer. The current behaviour is to
+  ignore all but the first received manifest, but later on, we may add matching
+  manifests to track similar peers.
+  """
+  @spec new_manifest(t(), SnapshotManifest.manifest()) :: t()
+  def new_manifest(warp_queue, manifest) do
+    if warp_queue.manifest do
+      # Right now, ignore new manifests
+      warp_queue
+    else
+      manifest_block_hashes = MapSet.new(manifest.block_hashes)
+      manifest_state_hashes = MapSet.new(manifest.state_hashes)
+      manifest_hashes = MapSet.union(manifest_block_hashes, manifest_state_hashes)
+
+      %{
+        warp_queue
+        | manifest: manifest,
+          manifest_hashes: manifest_hashes,
+          manifest_block_hashes: manifest_block_hashes,
+          manifest_state_hashes: manifest_state_hashes
+      }
+    end
+  end
+
+  @doc """
+  When we receive a new block chunk, we want to remove it from requests
+  and add it to our processing queue.
+  """
+  @spec new_block_chunk(t(), EVM.hash()) :: t()
+  def new_block_chunk(warp_queue, chunk_hash) do
+    updated_chunk_requests = MapSet.delete(warp_queue.chunk_requests, chunk_hash)
+    updated_retrieved_chunks = MapSet.put(warp_queue.retrieved_chunks, chunk_hash)
+
+    %{
+      warp_queue
+      | chunk_requests: updated_chunk_requests,
+        retrieved_chunks: updated_retrieved_chunks
+    }
+  end
+
+  @doc """
+  When we receive a new state chunk, we simply add it to our queue, which we'll
+  later process.
+  """
+  @spec new_state_chunk(t(), EVM.hash()) :: t()
+  def new_state_chunk(warp_queue, chunk_hash) do
+    updated_chunk_requests = MapSet.delete(warp_queue.chunk_requests, chunk_hash)
+    updated_retrieved_chunks = MapSet.put(warp_queue.retrieved_chunks, chunk_hash)
+
+    %{
+      warp_queue
+      | chunk_requests: updated_chunk_requests,
+        retrieved_chunks: updated_retrieved_chunks
+    }
+  end
+
+  @spec get_hashes_to_request(t(), number(), number()) :: {t(), list(EVM.hash())}
+  def get_hashes_to_request(
+        warp_queue = %__MODULE__{
+          chunk_requests: chunk_requests,
+          retrieved_chunks: retrieved_chunks,
+          processed_chunks: processed_chunks
+        },
+        request_limit,
+        queue_limit
+      ) do
+    queued_count =
+      Enum.count(
+        MapSet.difference(
+          retrieved_chunks,
+          processed_chunks
+        )
+      )
+
+    allowed_by_parallelism = request_limit - MapSet.size(chunk_requests)
+    allowed_by_queue = queue_limit - queued_count
+
+    desired_requests = min(allowed_by_parallelism, allowed_by_queue)
+
+    if desired_requests > 0 do
+      unfetched_block_hashes =
+        warp_queue.manifest_block_hashes
+        |> MapSet.difference(chunk_requests)
+        |> MapSet.difference(retrieved_chunks)
+        |> MapSet.difference(processed_chunks)
+        |> MapSet.to_list()
+
+      unfetched_state_hashes =
+        warp_queue.manifest_state_hashes
+        |> MapSet.difference(chunk_requests)
+        |> MapSet.difference(retrieved_chunks)
+        |> MapSet.difference(processed_chunks)
+        |> MapSet.to_list()
+
+      total_unfetched_hashes = unfetched_block_hashes ++ unfetched_state_hashes
+      total_unfetched_count = Enum.count(total_unfetched_hashes)
+
+      if min(total_unfetched_count, desired_requests) > 0 do
+        hashes_to_request = Enum.take(total_unfetched_hashes, desired_requests)
+
+        :ok =
+          Logger.debug(fn ->
+            "[Warp] Retreiving #{Enum.count(hashes_to_request)} of #{total_unfetched_count} hash(es) needed."
+          end)
+
+        new_chunk_requests =
+          MapSet.union(
+            chunk_requests,
+            MapSet.new(hashes_to_request)
+          )
+
+        {
+          %{warp_queue | chunk_requests: new_chunk_requests},
+          hashes_to_request
+        }
+      else
+        {warp_queue, []}
+      end
+    else
+      {warp_queue, []}
+    end
+  end
+
+  @spec status(t()) :: {:pending, atom()} | :success | {:failure, atom()}
+  def status(warp_queue) do
+    cond do
+      is_nil(warp_queue.manifest) ->
+        {:pending, :no_manifest}
+
+      MapSet.size(warp_queue.chunk_requests) > 0 ->
+        {:pending, :awaiting_requests}
+
+      !MapSet.equal?(warp_queue.manifest_hashes, warp_queue.processed_chunks) ->
+        {:pending, :awaiting_processing}
+
+      is_nil(warp_queue.block_tree.best_block) ->
+        {:failure, :missing_best_block}
+
+      warp_queue.block_tree.best_block.header.number != warp_queue.manifest.block_number ->
+        :pending
+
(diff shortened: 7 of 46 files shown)


apps/ex_wire/lib/ex_wire/packet/warp_status.ex
```diff
@@ -0,0 +1,180 @@
...
+
+  @doc """
+  Build a WarpStatus packet
+
+  Note: we are currently reflecting values based on the packet received, but
+  that should not be the case. We should provide the total difficulty of the
+  best chain found in the block header, the best hash, and the genesis hash of
+  our blockchain.
+
+  TODO: Don't parrot the same data back to sender
```
MattMSumner: This `TODO` also appears [here](https://github.com/mana-ethereum/mana/pull/632/files#diff-53e9c539a86c0c1e93acf699b41c7fb4R59). Does it make sense to have it only in one place?

apps/ex_wire/lib/ex_wire/packet/warp_status.ex
```diff
@@ -0,0 +1,180 @@
...
+        network_id: 3,
+        total_difficulty: 10,
+        best_hash: <<5>>,
+        genesis_hash: <<6::256>>,
+        snapshot_hash: <<7::256>>,
+        snapshot_number: 8,
+      }
+  """
+  @spec deserialize(ExRLP.t()) :: t()
+  def deserialize(rlp) do
```
MattMSumner: This could be:
```ex
def deserialize([
  protocol_version,
  network_id,
  total_difficulty,
  best_hash,
  genesis_hash,
  snapshot_hash,
  snapshot_number
]) do
  %__MODULE__{
    protocol_version: :binary.decode_unsigned(protocol_version),
    network_id: :binary.decode_unsigned(network_id),
    total_difficulty: :binary.decode_unsigned(total_difficulty),
    best_hash: best_hash,
    genesis_hash: genesis_hash,
    snapshot_hash: snapshot_hash,
    snapshot_number: snapshot_number
  }
end
```

apps/ex_wire/lib/ex_wire/struct/warp_queue.ex
```diff
@@ -0,0 +1,275 @@
...
+    :current_block_number,
+    :current_block_hash,
+    :known_block_hashes,
+    :known_state_hashes,
+    :block_chunk_requests,
+    :unprocessed_block_data
+  ]
+
+  @doc """
+  Creates a new `WarpQueue` that
```
MattMSumner: Does this need to be expanded? Otherwise we could lose the `that`:
```suggestion
  Creates a new `WarpQueue`.
```

apps/ex_wire/lib/ex_wire/struct/warp_queue.ex
```diff
@@ -0,0 +1,275 @@
...
+  end
+
+  @doc """
+  When we receive a new block chunk, we want to remove it from requests
+  and add it to our processing queue.
+  """
+  @spec new_block_chunk(WarpQueue.t(), BlockChunk.t()) :: WarpQueue.t()
+  def new_block_chunk(warp_queue, block_chunk) do
+    next_warp_queue =
+      if !is_nil(warp_queue.first_block_hash) &&
```
MattMSumner: Is it worth extracting some functions to make this check clearer? Something like:

```ex
def new_block_chunk(warp_queue, block_chunk) do
  should_update_queue = first_block_present?(warp_queue) &&
    block_chunk_hash(block_chunk) == warp_queue.first_block_hash

  next_warp_queue = if should_update_queue do
    %{
      warp_queue |
        first_block_hash: nil,
        current_block_number: block_chunk.number,
        current_block_hash: block_chunk.hash
    }
  else
    warp_queue
  end
...

defp first_block_present?(warp_queue)
  !is_nil(warp_queue.first_block_hash)
end

defp block_chunk_hash(block_chunk)
  block_chunk
  |> BlockChunk.serialize()
  |> ExRLP.encode()
  |> ExthCrypto.Hash.Keccak.kec()
end
```

apps/ex_wire/lib/ex_wire/struct/warp_queue.ex
```diff
@@ -0,0 +1,275 @@
...
+    }
+  end
+
+  # TODO: We need to handle requesting state hashes, as well.
+  @spec get_block_hashes_to_request(t(), integer()) :: {t(), list(EVM.hash())}
+  def get_block_hashes_to_request(warp_queue, request_limit) do
+    blocks_to_request = request_limit - Enum.count(warp_queue.block_chunk_requests)
+
+    if blocks_to_request > 0 do
+      unrequested_blocks =
```
MattMSumner: I think you could pipe this into `blocks_hashes_to_request` i.e.

```ex
blocks_hashes_to_request =
  warp_queue.known_block_hashes
  |> MapSet.difference(warp_queue.block_chunk_requests)
  |> Enum.take(blocks_to_request)
```

apps/ex_wire/lib/ex_wire/struct/warp_queue.ex
```diff
@@ -0,0 +1,275 @@
...
+      receipts: block_data.receipts,
+      ommers: block_data.header.ommers
+    }
+
+    Exth.inspect(block, "Processing Block!")
+
+    # Store the block to our db
+    {:ok, {_, next_trie}} = Block.put_block(block, trie)
+
+    # TODO: Store each transaction to our db
```
MattMSumner: Checking for my own understanding, but these TODOs are blockers for actually using warp_sync, correct?

apps/ex_wire/lib/ex_wire/sync.ex
```diff
@@ -18,9 +18,21 @@ defmodule ExWire.Sync do
   alias Block.Header
   alias Blockchain.Block
   alias ExWire.Config
-  alias ExWire.Packet.{BlockBodies, BlockHeaders, GetBlockBodies, GetBlockHeaders}
+
+  alias ExWire.Packet.{
```
MattMSumner: In elixir can you alias all of a module? i.e. `alias ExWire.Packet._`?

apps/ex_wire/lib/ex_wire/sync.ex
```diff
@@ -97,47 +130,165 @@ defmodule ExWire.Sync do
...
     :ok = Exth.trace(fn -> "[Sync] Ignoring packet #{packet.__struct__} from #{peer}" end)
 
     {:noreply, state}
   end
 
   @doc """
-  Dispatches a packet of `GetBlockHeaders` to all peers for the next block
+  Dispatches a packet of `GetSnapshotManifest` to all capable peers.
+
+  # TODO: That "capable peers" part.
```
MattMSumner: Again this is mostly so I understand the meaning here; "capable peers" are peers that tell us they are able to receive the snapshot_manifest, correct? Will peers explicitly ask for snapshot_manifests or is the expectation that once we've setup a connection with peers that we'll keep sending them snapshot_manifests if they are capable?

apps/ex_wire/lib/ex_wire/sync.ex
```diff
@@ -97,47 +130,165 @@ defmodule ExWire.Sync do
...
+  @doc """
+  Dispatches a packet of `GetSnapshotData` to a random capable peer.
+
+  # TODO: That "capable peer" part.
+  """
+  @spec handle_request_block_chunk(EVM.hash(), boolean(), state()) :: state()
+  def handle_request_block_chunk(chunk_hash, is_block_chunk, state) do
+    if send_with_retry(
+         %GetSnapshotData{chunk_hash: chunk_hash},
+         :random,
```
MattMSumner: Do most implementations send these randomly or is the expectation that we'll send them to all peers over time?

apps/ex_wire/lib/ex_wire/sync.ex
```diff
@@ -97,47 +130,165 @@ defmodule ExWire.Sync do
...
-          max_headers: @blocks_per_request,
-          skip: 0,
-          reverse: false
-        },
-        :random
-      )
+  @doc """
+  When we receive a new snapshot manifest, we add it to our warp queue. We may
+  have new blocks to fetch, so we ask the warp queue for more blocks to
+  request. We may already, however, we waiting on blocks, in which case we
```
MattMSumner:
```suggestion
  request. We may already, however, be waiting on blocks, in which case we
```

apps/ex_wire/lib/ex_wire/sync.ex
```diff
@@ -97,47 +130,165 @@ defmodule ExWire.Sync do
...
+  end
+
+  def handle_snapshot_data(
+        %SnapshotData{chunk: state_chunk = %StateChunk{}},
+        _peer,
+        state = %{warp_queue: _warp_queue}
+      ) do
+    Exth.inspect(state_chunk, "Got state_chunk, not doing what I should")
+
+    # WarpQueue.new_block_chunk(warp_queue, block_chunk)
```
MattMSumner: Should this comment be removed? Or uncommented?

apps/exth/test/exth_test.exs
```diff
@@ -0,0 +1,8 @@
+defmodule ExthTest do
+  use ExUnit.Case
+  doctest Exth
+
+  test "greets the world" do
```
MattMSumner: How would you feel about removing this? Unless this is the implementation for `Packet.hello`?

apps/ex_wire/lib/ex_wire/packet/status.ex
```diff
@@ -89,43 +96,36 @@ defmodule ExWire.Packet.Status do
...
+      total_difficulty,
+      best_hash,
+      genesis_hash
+    ] = rlp
 
     %__MODULE__{
-      protocol_version: protocol_version |> :binary.decode_unsigned(),
-      network_id: network_id |> :binary.decode_unsigned(),
-      total_difficulty: total_difficulty |> :binary.decode_unsigned(),
+      protocol_version: :binary.decode_unsigned(protocol_version),
```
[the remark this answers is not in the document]
hayesgm: So RLP encoding encodes `<<5>>` and `5` the same way, so there's no way to tell if we wanted a binary or an integer. In ExRLP, we decode to return a binary, but then it bites us because a lot of our code expects `val == ExRLP.decode(ExRLP.encode(val))` to be true, but for the value `[5]`, the result is `[<<5>>]`. Then, this decode function tries to get deserialize that list value into a struct, but it's stuck not knowing if it's going to get `5` or `<<5>>`, and it's just been too much effort to control for the input when we can just accept either. This `Exth. maybe_decode_unsigned` takes either and returns `5`.

apps/ex_wire/config/dev.exs
```diff
@@ -2,8 +2,8 @@ use Mix.Config
 
 config :ex_wire,
   network_adapter: {ExWire.Adapter.UDP, NetworkClient},
-  sync: true,
-  discovery: true,
+  sync: false,
```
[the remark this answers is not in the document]
hayesgm: Now that we've hooked up the CLI, it doesn't make as much sense to connect when this application starts. To start ExWire, use the mix task `mix mana` which will handle this correctly.

apps/ex_wire/lib/ex_wire/packet.ex
```diff
@@ -19,6 +19,13 @@ defmodule ExWire.Packet do
           :ok | :activate | :peer_disconnect | {:disconnect, atom()} | {:send, struct()}
   @callback handle(packet) :: handle_response
 
+  # TODO: The packet numbers here are... wrong. It's not explained in the docs
```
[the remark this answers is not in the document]
hayesgm: The packet numbers are supposed to be figured out based on the handshake, but 1. I didn't know this until I just read that in parity, I previously thought they were hardcoded (this is undocumented), so 2. these are hard-coded based on me talking to a local Parity node. Instead, this list needs to be the raw offsets and we need to determine the actual `packet_id`s based on the handshake.

apps/ex_wire/lib/ex_wire/packet.ex
```diff
@@ -27,22 +34,25 @@ defmodule ExWire.Packet do
...
+    0x27 => Packet.BlockBodies,
     # 0x17 => Packet.NewBlock,
     ### Fast synchronization (PV63)
     # 0x1d => Packet.GetNodeData,
     # 0x1e => Packet.NodeData,
     # 0x1f => Packet.GetReceipts,
-    # 0x20 => Packet.Receipts
+    # 0x20 => Packet.Receipts,
+    ### Warp Sync (par)
+    # TODO: Fix the hack out of this
```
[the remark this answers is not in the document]
hayesgm: See above, these should no longer we hard-coded. It should really be `0x00 => Packet.WarpStatus, 0x11 => Packet.GetSnapshotManifest`, etc, and the `0x21` "padding" should be negotiated.

apps/ex_wire/lib/ex_wire/packet/get_snapshot_manifest.ex
```diff
@@ -0,0 +1,60 @@
...
+  Given a GetSnapshotManifest packet, serializes for transport over Eth Wire Protocol.
+
+  ## Examples
+
+      iex> %ExWire.Packet.GetSnapshotManifest{}
+      ...> |> ExWire.Packet.GetSnapshotManifest.serialize()
+      []
+  """
+  @spec serialize(t) :: ExRLP.t()
+  def serialize(_packet = %__MODULE__{}) do
```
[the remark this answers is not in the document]
hayesgm: It's just pattern matching against the packet being passed in is a struct of this module. Thus if I called: `ExWire.Packet.GetSnapshotManifest.serialize(%{})` it would fail with a type match error. Acts as a guard.

apps/ex_wire/lib/ex_wire/packet/snapshot_data/block_chunk.ex
```diff
@@ -0,0 +1,292 @@
...
+            timestamp: integer(),
+            extra_data: integer(),
+
+            # Ommers and transactions inline
+            transactions: list(Blockchain.Transaction.t()),
+            transactions_rlp: list(ExRLP.t()),
+            ommers: list(Block.Header.t()),
+            ommers_rlp: list(ExRLP.t()),
+
+            # Seal fields
```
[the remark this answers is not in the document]
hayesgm: This comes from [Parity's documentation](https://wiki.parity.io/Warp-Sync-Snapshot-Format.html#abridged-block-rlp) but it means the `mix_hash` and `nonce`. The proof of work that seals the block or whatever

apps/ex_wire/lib/ex_wire/struct/warp_queue.ex
```diff
@@ -0,0 +1,275 @@
...
+      block,
+      next_trie
+    }
+  end
+
+  # Tries to get the transaction or receipts root by encoding the transaction trie
+  @spec get_transactions_or_receipts_root([ExRLP.t()]) :: MerklePatriciaTree.Trie.root_hash()
+  def get_transactions_or_receipts_root(transactions_or_receipts_rlp) do
+    # this is a throw-away
+    db = MerklePatriciaTree.Test.random_ets_db()
```
ayrat555: why test in-memory database is used? Also, I think we should use `TrieStorage` behaviour to working with database

hayesgm: If we can sure, the idea is simply turning a list of transactions into the state root of a trie with all of the transactions included. We **never** store this value to disk (hence why we use ets and this discard it).

apps/ex_wire/lib/ex_wire/packet.ex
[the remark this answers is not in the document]
MattMSumner: How would you feel about something like:
```suggestion
  # TODO: The packet numbers here are hardcoded and shouldn't be. It's not documented
```

apps/ex_wire/lib/ex_wire/packet/snapshot_data/state_chunk.ex
```diff
@@ -0,0 +1,194 @@
+defmodule ExWire.Packet.SnapshotData.StateChunk do
+  @moduledoc """
+  State chunks store the entire state of a given block. A "rich" account
+  structure is used to save space. Each state chunk consists of a list of
+  lists, each with two items: an address' sha3 hash and a rich account
```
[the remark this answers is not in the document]
hayesgm: I think it is mean to be a possessive (comes right from Parity's docs)

apps/ex_wire/lib/ex_wire/packet/snapshot_data/state_chunk.ex
```diff
@@ -0,0 +1,194 @@
...
+  def serialize(state_chunk = %__MODULE__{}) do
+    for {hash, rich_account} <- state_chunk.account_entries do
+      [
+        hash,
+        [
+          rich_account.nonce,
+          rich_account.balance,
+          RichAccount.encode_code_flag(rich_account.code_flag),
+          rich_account.code,
+          for {key, val} <- rich_account.storage do
```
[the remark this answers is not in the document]
hayesgm: cool! I've give that a whirl.

apps/ex_wire/lib/ex_wire/packet/snapshot_data/state_chunk.ex
```diff
@@ -0,0 +1,194 @@
...
+      for [hash, rich_account_rlp] <- account_entries_rlp do
+        [
+          nonce,
+          balance,
+          encoded_code_flag,
+          code,
+          storage_rlp
+        ] = rich_account_rlp
+
+        storage =
```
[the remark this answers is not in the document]
hayesgm: And this, looks awesome.

apps/ex_wire/lib/ex_wire/packet/snapshot_manifest.ex
```diff
@@ -0,0 +1,151 @@
...
+        }
+      }
+
+      iex> ExWire.Packet.SnapshotManifest.deserialize([[3, [<<1::256>>, <<2::256>>], [<<3::256>>, <<4::256>>], <<5::256>>, 6, <<7::256>>]])
+      ** (MatchError) no match of right hand side value: 3
+  """
+  @spec deserialize(ExRLP.t()) :: t
+  def deserialize([]), do: %__MODULE__{manifest: nil}
+
+  def deserialize([rlp]) do
```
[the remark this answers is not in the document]
hayesgm: Awesome catch, yeah, we're probably serializing wrong. This deserialization appears to be right.

apps/ex_wire/lib/ex_wire/packet/warp_status.ex
hayesgm: As above, I tend to like calling rlp `rlp`, and matching it in the body. For me, it's just slightly more intuitive, if slightly less clean.

apps/ex_wire/lib/ex_wire/struct/warp_queue.ex
hayesgm: Yup, I'm still.... working this one out. But once I do, for sure!

apps/ex_wire/lib/ex_wire/struct/warp_queue.ex
hayesgm: You're right, will do.

apps/ex_wire/lib/ex_wire/struct/warp_queue.ex
hayesgm: Yes, it shouldn't be too difficult, it's mostly: "commit KV values to database," but waiting for advice from @ayrat555 on how best to do that storage.

apps/ex_wire/lib/ex_wire/sync.ex
hayesgm: Not sure, but I like keeping the imports kind of explicit here. It basically is saying `sync.ex` has gotten too complicated :)

apps/ex_wire/lib/ex_wire/sync.ex
hayesgm: Peers advertise capabilities during the handshake process. In this case, we should send a request to all peers with `{'par', 1}` capabilities. Finding and sending to the right peers is kind of the easy part, dealing with multiple manifests and coming to a sane way of processing them will be the hard part (maybe we take the best and still farm out requests to multiple peers? what happens when we're started a sync and then get a better manifest? etc)

apps/ex_wire/lib/ex_wire/sync.ex
hayesgm: If I remember how Geth did it for Fast sync, it broke up the space into chunks and requested some chunks from some peers and other chunks from other peers. Random should effectively achieve the same result.

apps/ex_wire/lib/ex_wire/sync.ex
hayesgm: Yeah, this is straight-WIP

apps/exth/test/exth_test.exs
hayesgm: Yes, will be removed. `Exth` should be tested. I know most people don't love catch-alls, but we're getting close to a lot of repeated code.

apps/ex_wire/lib/ex_wire/packet/status.ex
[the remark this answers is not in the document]
willmeister: I like the flexibility, but the fact that @MattMSumner needed to ask (and I was going to as well), suggests that the name could be improved. How do you feel about something like `decode_to_unsigned_scalar`?

apps/ex_wire/lib/ex_wire/packet/snapshot_data.ex
```diff
@@ -0,0 +1,112 @@
...
+      ...> |> ExWire.Packet.SnapshotData.deserialize(false)
+      %ExWire.Packet.SnapshotData{
+        chunk: %ExWire.Packet.SnapshotData.StateChunk{
+          account_entries: []
+        }
+      }
+  """
+  @spec deserialize(ExRLP.t(), boolean()) :: t()
+  def deserialize(rlp, is_block_chunk \\ true) do
+    [chunk_data] = rlp
```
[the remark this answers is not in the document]
willmeister: Same thing can be done [here](https://github.com/mana-ethereum/mana/pull/632/files#diff-37ff14e7147c96498071221c4e9398cbR46)

apps/ex_wire/lib/ex_wire/packet/snapshot_data/block_chunk.ex
```diff
@@ -0,0 +1,292 @@
...
+            receipts_rlp: list(ExRLP.t())
+          }
+    defstruct [
+      :header,
+      :receipts,
+      :receipts_rlp
+    ]
+  end
+
+  @type t :: %__MODULE__{
```
willmeister: Would prefer if these were up by the `defmodule` so it's easy to see what `%__MODULE__` is referring to. Unless this is pretty standard in Elixir, in which case I'll just have to get used to it =D

apps/ex_wire/lib/ex_wire/struct/warp_queue.ex
```diff
@@ -0,0 +1,293 @@
...
+      transactions: block_data.header.transactions,
+      receipts: block_data.receipts,
+      ommers: block_data.header.ommers
+    }
+  end
+
+  # Note: we haven't connected state hashes, so we're not there yet
+  @spec process_block(Block.t(), BlockData.t(), Trie.t()) :: Trie.t()
+  defp process_block(block, block_data, trie) do
+    # First, we need to validate some aspect of the block, that's currently
```
willmeister: Add `TODO:`?

hayesgm:
This is getting pretty close to being done, the remaining open topics are:

1. I don't believe we're properly calculating the final state root, which is necessary to verify the sync was successful.
2. We should store the block-tree in the warp-queue, so we reload it after a failed sync if we load from warp state.
3. We don't currently support PoA chains. This is "documented" here: https://github.com/paritytech/parity-ethereum/blob/master/ethcore/src/snapshot/consensus/authority.rs#L43

That said, we've done a lot of work on getting this to work efficiently. Previously, loading account data was enough to crash erlang for a variety of reasons. We know can handle parallel processing of blocks and accounts with relative ease.

hayesgm:
Question: would others prefer that I break up a variety of the speed improvements to other PRs? Those will block this PR, but probably would be much clearer than this now-very-long PR. Thoughts?

ayrat555:
+1 for breaking up this PR

apps/blockchain/lib/blockchain/block_setter.ex
```diff
@@ -18,12 +18,18 @@ defmodule Blockchain.BlockSetter do
...
-  def set_block_number(block, parent_block) do
-    number = parent_block.header.number + 1
-    header = %{block.header | number: number}
-    %{block | header: header}
+  def set_block_number(block, %Block{header: %Header{number: nil}}) do
+    %{block | header: %{block.header | number: nil}}
+  end
+
+  def set_block_number(block, %Block{header: %Header{number: parent_block_number}})
+      when is_integer(parent_block_number) do
```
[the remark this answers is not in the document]
hayesgm: I was just trying to help dialyzer - I think your PR addresses that better.

apps/blockchain/lib/blockchain/blocktree.ex
```diff
@@ -77,7 +77,15 @@ defmodule Blockchain.Blocktree do
          do: block,
          else: best_block
 
-    %{blocktree | best_block: new_best_block}
+    # Make sure block is stored with a block hash
+    new_best_block_with_hash =
+      if new_best_block.block_hash do
```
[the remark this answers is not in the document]
hayesgm: yeah, I wish elixir has a built in `is_not_nil` operator since I don't love doing `if !is_nil(...)`. Maybe `!= nil`?

apps/ex_wire/lib/ex_wire/p2p/server.ex
```diff
@@ -110,7 +110,7 @@ defmodule ExWire.P2P.Server do
   """
   @spec get_peer(pid()) :: Peer.t()
   def get_peer(pid) do
-    GenServer.call(pid, :get_peer)
+    GenServer.call(pid, :get_peer, :infinity)
```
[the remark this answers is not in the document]
hayesgm: Yes, right now, there's really no reason to timeout here. Though, this might be mitigated by better arrangement of servers. But again, the timeout wasn't really helping us.

apps/ex_wire/config/config.exs
```diff
@@ -27,10 +27,7 @@ mana_version =
 config :ex_wire,
   p2p_version: 0x04,
   protocol_version: 63,
-  # ropsten
-  network_id: 3,
-  caps: [{"eth", 62}, {"eth", 63}],
-  chain: :ropsten,
+  caps: [{"eth", 62}, {"eth", 63}, {"par", 1}],
```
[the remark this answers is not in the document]
hayesgm: Yup, "parity's first capability"

apps/ex_wire/lib/ex_wire/sync/warp_processor.ex
```diff
@@ -0,0 +1,570 @@
...
+          pid(),
+          EVM.hash(),
+          StateChunk.t(),
+          Trie.t(),
+          processor_mod(),
+          pid()
+        ) :: Task.t()
+  defp new_state_chunk_task(processor_pid, sup, chunk_hash, state_chunk, trie, processor_mod, pid) do
+    Task.Supervisor.async(sup, fn ->
+      start = Time.time_start()
```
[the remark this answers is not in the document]
hayesgm: Yup, I played around with this a few ways, but I wanted something to help us report times for sync since the process takes between a few minutes and a few hours and I really want to see `blocks/s`, etc, so I could figure out the rate. Over time, I plan to evolve this, but it seemed easy  enough to use our own little lib that had nicer reporting facilities.

apps/blockchain/lib/blockchain/blocktree.ex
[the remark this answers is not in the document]
InoMurko: maybe `case`?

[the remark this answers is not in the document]
hayesgm: reversed the cases and checked for nil

apps/ex_wire/lib/ex_wire/sync/warp_state.ex
```diff
@@ -0,0 +1,57 @@
+defmodule ExWire.Sync.WarpState do
+  @moduledoc """
+  Warp may take some time, and thus we want to occausionally snapshot
```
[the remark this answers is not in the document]
hayesgm: fixed

apps/blockchain/lib/blockchain/blocktree.ex
```diff
@@ -77,7 +77,15 @@ defmodule Blockchain.Blocktree do
          do: block,
          else: best_block
 
-    %{blocktree | best_block: new_best_block}
+    # Make sure block is stored with a block hash
+    new_best_block_with_hash =
+      if new_best_block.block_hash == nil do
+        %{new_best_block | block_hash: Block.hash(new_best_block)}
```
ayrat555: I think this may cause multiple calculations of hash for the same block. Because it's calculated when the block is stored

apps/cli/lib/parser/mana_parser.ex
```diff
@@ -27,15 +29,17 @@ defmodule CLI.Parser.ManaParser do
         discovery: false,
         sync: false,
         bootnodes: :from_chain,
+        warp: false,
         debug: false
       }}
 
-      iex> CLI.Parser.ManaParser.mana_args(["--chain", "foundation", "--bootnodes", "enode://google.com,enode://apple.com", "--debug"])
+      iex> CLI.Parser.ManaParser.mana_args(["--chain", "foundation", "--bootnodes", "enode://google.com,enode://apple.com", "--warp", "--debug"])
```
ayrat555: :apple:

apps/ex_wire/lib/ex_wire/packet.ex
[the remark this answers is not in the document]
ayrat555: @hayesgm are you going to fix it?

apps/ex_wire/lib/ex_wire/packet/snapshot_data.ex
```diff
@@ -0,0 +1,154 @@
...
+      ...> |> ExWire.Packet.SnapshotData.serialize()
+      [<<145, 1, 20, 248, 143, 248, 141, 160, 0, 118, 1, 0, 20, 1, 248, 106, 2, 3, 1, 126, 38, 0, 16, 5, 248, 68, 248, 66, 126, 37, 0, 130, 70, 0, 0, 2>>]
+  """
+  @spec serialize(t) :: ExRLP.t()
+  def serialize(%__MODULE__{chunk: chunk = %{__struct__: mod}}) do
+    {:ok, res} =
+      chunk
+      |> mod.serialize()
+      |> ExRLP.encode()
+      |> :snappyer.compress()
```
ayrat555: what does ` :snappyer.compress` do?

apps/blockchain/lib/blockchain/blocktree.ex
hayesgm: It's a bit weird, it's really meant to be a write-through cache, as in, don't calculate it until you need it, but then you should store it in the block to not have to calculate it again. If, downstream, the code reads block hash, it should really check to see if it was already calculated here.

hayesgm: Also, we should probably just add a function in `Block` that is `Block.with_hash` that just handles this logic.

PR #633, opened 2018-11-19
methods implementation

InoMurko:
Added support for starting_block_number and highest_block_number as per
https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_syncing and https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_blocknumber

added all methods for RPC (but they currently return a custom not supported error).


apps/ex_wire/lib/ex_wire/packet/block_headers.ex
```diff
@@ -13,11 +13,11 @@ defmodule ExWire.Packet.BlockHeaders do
   """
 
   require Logger
-
+  alias Block.Header
```
MattMSumner: Nice 👍

apps/ex_wire/lib/ex_wire/sync.ex
```diff
@@ -17,13 +17,18 @@ defmodule ExWire.Sync do
...
+  alias Blockchain.Blocktree.State
+  alias Blockchain.Chain
   alias ExWire.Config
   alias ExWire.Packet.{BlockBodies, BlockHeaders, GetBlockBodies, GetBlockHeaders}
   alias ExWire.PeerSupervisor
-  alias ExWire.Struct.{BlockQueue, Peer}
-  alias Blockchain.{Blocktree, Chain}
-  alias Blockchain.Blocktree.State
-  alias MerklePatriciaTree.{CachingTrie, DB.RocksDB, Trie, TrieStorage}
+  alias ExWire.Struct.BlockQueue
```
MattMSumner: I see this has moved from the `{}` alias syntax to an alias per line. Is there a reason this is desirable or is this mostly a stylistic choice?

apps/jsonrpc2/lib/jsonrpc2/spec_handler.ex
```diff
@@ -1,5 +1,114 @@
...
+    current_block = get_last_sync_block(sync_state)
+
+    %{
+      currentBlock: current_block,
+      startingBlock: sync_state.starting_block_number,
+      highestBlock: sync_state.highest_block_number
+    }
+  end
+
+  def handle_request("eth_coinbase", _), do: {:error, :not_supported}
```
MattMSumner: This is great. I like that we can add these slowly one by one while also clearly seeing what is missing 👍

apps/jsonrpc2/lib/jsonrpc2/spec_handler.ex
```diff
@@ -1,5 +1,114 @@
...
+  def handle_request("eth_newPendingTransactionFilter", _), do: {:error, :not_supported}
+  def handle_request("eth_uninstallFilter", _), do: {:error, :not_supported}
+  def handle_request("eth_getFilterChanges", _), do: {:error, :not_supported}
+  def handle_request("eth_getFilterLogs", _), do: {:error, :not_supported}
+  def handle_request("eth_getLogs", _), do: {:error, :not_supported}
+  def handle_request("eth_getWork", _), do: {:error, :not_supported}
+  def handle_request("eth_submitWork", _), do: {:error, :not_supported}
+  def handle_request("eth_submitHashrate", _), do: {:error, :not_supported}
+  def handle_request("eth_getProof", _), do: {:error, :not_supported}
+  # db_putString is deprecated
```
MattMSumner: Does this mean that we'll keep `db_putString` returning `not_supported` forever?

InoMurko: Depends on when we release the 1.0 version :) and if they remove it by then...

rel/config.exs
```diff
@@ -60,7 +60,6 @@ environment :dev do
   # dev mode.
   set(dev_mode: true)
   set(include_erts: true)
-  set(cookie: :"bAFS)[qmAg0XFS5`7Ncbz,$J:K3Tv)t@{OIPWJb`;N!k%z^H,^(L}:qI_$Tp8>EX")
```
atoulme: is this change related?

InoMurko: No. But it was breaking the CLI remote console functionality.

apps/jsonrpc2/lib/jsonrpc2/spec_handler.ex
```diff
@@ -1,5 +1,127 @@
...
   use JSONRPC2.Server.Handler
 
-  def handle_request("web3_clientVersion", _), do: Application.get_env(:jsonrpc2, :mana_version)
+  alias Blockchain.Block
+  alias Blockchain.Blocktree
+  alias ExthCrypto.Hash.Keccak
+  alias ExthCrypto.Math
+  alias ExWire.PeerSupervisor
+  alias ExWire.Sync
+  # TODO split methods into separate modules per topic
```
ayrat555: maybe it's better to create a github issue instead of leaving TODO comments?

apps/jsonrpc2/lib/jsonrpc2/spec_handler.ex
```diff
@@ -1,5 +1,127 @@
...
+  # web3 Methods
+
+  def handle_request("web3_clientVersion", _),
+    do: "#{Application.get_env(:jsonrpc2, :mana_version)}"
```
ayrat555: I think string interpolation is not needed here

apps/jsonrpc2/lib/jsonrpc2/spec_handler.ex
```diff
@@ -1,5 +1,127 @@
...
+
+  def handle_request("web3_sha3", [param = "0x" <> _]) do
+    "0x" <>
+      (param
+       |> Math.hex_to_bin()
+       |> Keccak.kec()
+       |> Math.bin_to_hex())
+  rescue
+    _ ->
+      {:error, :invalid_params}
```
ayrat555: maybe we can move this to somewhere (module attribute or separate module) to reuse in other methods?

InoMurko: Which part exactly?

apps/jsonrpc2/lib/jsonrpc2/spec_handler.ex
InoMurko: I think you’re right!

apps/jsonrpc2/lib/jsonrpc2/spec_handler.ex
InoMurko: Sure!

apps/jsonrpc2/lib/jsonrpc2/spec_handler.ex
ayrat555: `:invalid_params` atom or `{:error, :invalid_params}` tuple

InoMurko: Ok, yeah, I know what you mean now. I'll do that once I get a couple of more examples cleared out so that I understand the problem more.

apps/ex_wire/lib/ex_wire/sync.ex
```diff
@@ -321,4 +347,13 @@ defmodule ExWire.Sync do
       trie
     end
   end
+
+  @spec get_highest_block_number(non_neg_integer(), non_neg_integer()) :: non_neg_integer()
+  defp get_highest_block_number(current, header_block_number) do
```
willmeister: Can just be:
```
defp get_highest_block_number(current, header_block_number), do: Integer.max(current, header_block_number)
```
Right?

InoMurko: Good catch. Yes!

apps/ex_wire/lib/ex_wire/sync.ex
```diff
@@ -161,39 +181,43 @@ defmodule ExWire.Sync do
...
+          next_header_hashes =
+            if should_request_block do
+              :ok = Logger.debug(fn -> "[Sync] Requesting block body #{header.number}" end)
+
+              [header_hash | header_hashes]
+            else
+              header_hashes
+            end
+
+          next_highest_block_number = Integer.max(highest_block_number, header.number)
```
InoMurko: @ayrat555 would love your opinion on this. is this the correct way of estimating the highest block number?

willmeister: Looks like I lead you astray. It's `Kernel.max(...)`

apps/ex_wire/lib/ex_wire/sync.ex
```diff
@@ -161,39 +181,43 @@ defmodule ExWire.Sync do
...
+          next_header_hashes =
+            if should_request_block do
+              :ok = Logger.debug(fn -> "[Sync] Requesting block body #{header.number}" end)
+
+              [header_hash | header_hashes]
+            else
+              header_hashes
+            end
+
+          next_highest_block_number = Kernel.max(highest_block_number, header.number)
```
InoMurko: @ayrat555 would this be the correct way to estimate the highest block?

PR #641, opened 2018-11-20
add Byzantium to Constantinople transition test

ayrat555:
https://github.com/ethereum/tests/pull/547


ayrat555:
@MattMSumner Only one test.

PR #646, opened 2018-11-20
cast messages quicker in ExWire.Sync

InoMurko:
GenServer.cast always returns :ok. The additional checks were unnecessary and can also be made in a separate process which will release the parent proc.


InoMurko:
complicated approach for what it does.

PR #647, opened 2018-11-20
[WIP] Adding ExWire Responses for Status, GetBlockHeaders and GetBlockBodies

willmeister:


apps/blockchain/lib/blockchain/block.ex
```diff
@@ -230,6 +230,54 @@ defmodule Blockchain.Block do
     end
   end
 
+  @doc """
+  Returns the specified block from the database, if the block number is found in the provided map
+  and the resulting block hash is found in the provided TrieStorage.
+
+  ## Examples
```
InoMurko: hm... should we move these into block_test.exs? makes more sense as the module gets bigger I think

apps/ex_wire/lib/ex_wire/packet/get_block_bodies.ex
```diff
@@ -52,17 +59,61 @@ defmodule ExWire.Packet.GetBlockBodies do
...
   ## Examples
-
+      # Bodies not found test
       iex> %ExWire.Packet.GetBlockBodies{hashes: [<<5>>, <<6>>]}
       ...> |> ExWire.Packet.GetBlockBodies.handle()
-      :ok
+      {:send, %ExWire.Packet.BlockBodies{blocks: []}}
+
+      # Body found test
+      iex> ExWire.BridgeSyncMock.start_link(%{})
```
InoMurko: these are not so trivial examples anymore, why not move them into tests?

apps/ex_wire/lib/ex_wire/packet/get_block_bodies.ex
```diff
@@ -52,17 +59,61 @@ defmodule ExWire.Packet.GetBlockBodies do
...
+
+    {:send, BlockBodies.new(bodies)}
+  end
+
+  defp get_block_bodies(trie, hashes) do
+    hashes
+    |> Stream.map(fn hash ->
+      case Block.get_block(hash, trie) do
+        {:ok, block} ->
+          ExWire.Struct.Block.new(block)
```
InoMurko: alias

apps/ex_wire/lib/ex_wire/packet/get_block_headers.ex
```diff
@@ -11,15 +11,22 @@ defmodule ExWire.Packet.GetBlockHeaders do
...
+  alias Blockchain.Block
+  alias ExWire.Bridge.Sync
   alias ExWire.Packet
   alias ExWire.Packet.BlockHeaders
+  require Logger
 
   @behaviour ExWire.Packet
 
+  @sync Application.get_env(:ex_wire, :sync_mock, Sync)
+  # TODO: What's a reasonable number for this?
```
InoMurko: Ayrat once pointed out it might be better to start opening issues then leaving TODO's

apps/ex_wire/lib/ex_wire/packet/get_block_headers.ex
```diff
@@ -98,11 +105,86 @@ defmodule ExWire.Packet.GetBlockHeaders do
...
+        {:error, error} ->
+          _ =
+            Logger.debug(fn ->
+              "Error calling Sync.get_current_trie(): #{error}. Returning empty headers."
+            end)
+
+          []
+      end
+
+    get_headers_response(headers)
```
InoMurko: wouldn't this be set_headers_response?

apps/ex_wire/lib/ex_wire/packet/get_block_headers.ex
```diff
@@ -98,11 +105,86 @@ defmodule ExWire.Packet.GetBlockHeaders do
...
+  defp get_block_headers(_trie, _identifier, 0, _skip, _reverse) do
+    []
+  end
+
+  defp get_block_headers(trie, block_hash, num_headers, skip, reverse)
+       when is_binary(block_hash) do
+    case Block.get_block(block_hash, trie) do
+      {:ok, block} ->
+        next_number = next_block_number(block.header.number, skip, reverse)
+        [block.header | get_block_headers(trie, next_number, num_headers - 1, skip, reverse)]
```
InoMurko: hmmm... I think it might be a good exercise in writing this tail recursive.

apps/ex_wire/lib/ex_wire/packet/get_block_headers.ex
```diff
@@ -98,11 +105,86 @@ defmodule ExWire.Packet.GetBlockHeaders do
...
+  end
+
+  defp get_block_headers(trie, block_number, num_headers, skip, reverse) do
+    case Block.get_block_by_number(block_number, trie) do
+      {:ok, block} ->
+        next_block_number = next_block_number(block.header.number, skip, reverse)
+
+        [
+          block.header
+          | get_block_headers(trie, next_block_number, num_headers - 1, skip, reverse)
```
InoMurko: same as above, tail recursion

apps/ex_wire/lib/ex_wire/packet/status.ex
```diff
@@ -136,12 +138,27 @@ defmodule ExWire.Packet.Status do
   this peer. E.g. do our network and protocol versions match?
 
   ## Examples
+      iex> ExWire.BridgeSyncMock.start_link(%{})
```
InoMurko: yeah, definitely tests that require mocking shouldn't be in doctests. https://github.com/poanetwork/blockscout/issues/290 good read!

apps/ex_wire/lib/ex_wire/struct/block.ex
```diff
@@ -21,6 +21,16 @@ defmodule ExWire.Struct.Block do
           ommers: list(Header.t())
         }
 
+  @spec new(Blockchain.Block.t()) :: t()
```
InoMurko: alias

apps/ex_wire/lib/ex_wire/sync.ex
```diff
@@ -38,6 +38,11 @@ defmodule ExWire.Sync do
           last_requested_block: integer() | nil
         }
 
+  @spec get_state() :: state
```
InoMurko: hm... did you rebase? I think I added that, no? 
https://github.com/mana-ethereum/mana/blob/master/apps/ex_wire/lib/ex_wire/sync.ex#L49

apps/ex_wire/test/ex_wire/packet/get_block_bodies_test.exs
```diff
@@ -1,4 +1,4 @@
 defmodule ExWire.Packet.GetBlockBodiesTest do
-  use ExUnit.Case, async: true
+  use ExUnit.Case, async: false
```
InoMurko: if you move the tests here, where they belong, they can run async.

apps/ex_wire/lib/ex_wire/packet/status.ex
```diff
@@ -63,7 +65,7 @@ defmodule ExWire.Packet.Status do
       protocol_version: ExWire.Config.protocol_version(),
       network_id: ExWire.Config.network_id(),
       total_difficulty: packet.total_difficulty,
-      best_hash: packet.genesis_hash,
```
hayesgm: `genesis_hash` was returned here, previously, I think so say: "Our best hash is the genesis hash, because we know nothing!" We may want to (when there's no sync) just look at the current chain and return its `genesis_hash` as our best hash, etc, to correctly state we know nothing.

apps/ex_wire/test/ex_wire/packet/get_block_bodies_test.exs
hayesgm: Moving tests into the test from doctest wouldn't make them be able to run async. The change would be on not using named processes that conflict with one-another, right?

InoMurko: Ah yes, didn’t notice it’s named!

apps/ex_wire/lib/ex_wire/sync.ex
willmeister: Your branch wasn't merged at the time of writing, so I copied your changes until they were merged. Will rebase and make sure it merges properly.

apps/ex_wire/lib/ex_wire/packet/get_block_headers.ex
```diff
@@ -98,11 +105,86 @@ defmodule ExWire.Packet.GetBlockHeaders do
...
+            trie,
+            packet.block_identifier,
+            packet.max_headers,
+            packet.skip,
+            packet.reverse
+          )
+
+        {:error, error} ->
+          _ =
+            Logger.debug(fn ->
```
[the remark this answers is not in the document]
willmeister: Good call. I think I'll go with `warn` since it still returns a valid response, even though it probably won't accurately reflect our internal state.

apps/ex_wire/lib/ex_wire/packet/get_block_headers.ex
```diff
@@ -98,11 +105,86 @@ defmodule ExWire.Packet.GetBlockHeaders do
...
+  defp get_block_headers(trie, block_hash, num_headers, skip, reverse)
+       when is_binary(block_hash) do
+    case Block.get_block(block_hash, trie) do
+      {:ok, block} ->
+        next_number = next_block_number(block.header.number, skip, reverse)
+        [block.header | get_block_headers(trie, next_number, num_headers - 1, skip, reverse)]
+
+      _ ->
+        _ =
+          Logger.debug(fn -> "Could not find block with hash: #{Base.encode16(block_hash)}." end)
```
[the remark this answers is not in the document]
willmeister: Does `warn` assume that the request is for a valid block hash? I'm ambivalent, as `warn` might allow a caller to blow up our logs. That said happy to switch to `warn` if you think it makes sense.

apps/ex_wire/lib/ex_wire/struct/block.ex
willmeister: Is it weird to alias `Block` to mean a different `Block` than the containing module that is named `Block`? In other languages, I've used the full path for the module that isn't the current one to indicate that it's not what the reader might assume it is.

apps/ex_wire/lib/ex_wire/packet/get_block_bodies.ex
willmeister: `Blockchain.Block` is already aliased as `Block` (see line 108). That said, it may make sense to alias `ExWire.Struct.Block` as `Block`, since it is what ExWire considers a Block.

apps/ex_wire/lib/ex_wire/struct/block.ex
InoMurko: **It really depends on what you agree to.** I think our problem is that we're using aliases somewhere and on others we don't. We've yet to define when to alias and when not to for the sake of consistency. A good thing about Erlang is that it doesn't have aliases :)

So if you decide to alias, there's a thing `alias as`, if not, then don't. We should then disable `AliasUsage` in Credo and come up with a definition on when a developer should alias and when not... even for the sake of public contributions and our code consistency.

apps/ex_wire/lib/ex_wire/packet/get_block_headers.ex
willmeister: Was just for discussion in the PR. I'll remove it before merging. I'll look into what other clients do, but if nothing else, I'll just put in a number, and we'll change it if/when it becomes an issue.

apps/ex_wire/lib/ex_wire/packet/status.ex
willmeister: @hayesgm what is meant by _current chain_? Something like `genesis_hash = Chain.load_chain(@network_id_to_chain_name[packet.network_id]).genesis.parent_hash`?

apps/ex_wire/lib/ex_wire/struct/block.ex
hayesgm: FYI, I only alias because Credo tells us to. I think it's easier to have full names everywhere since it makes it simple to look up a module (versus reading "Sync.. is that CLI.Sync or ExWire.Sync? Let's look at the aliases). However, our current credo settings tells us we must alias **every module**. That said, `Blockchain.Block` versus `Block.Header` is completely terrible and my fault. We originally had `Blockchain.Block.Header`, but I realized that `EVM` needed `Block.Header`, so it was moved out. @germsvel has discussed several times rolling back this change and possible merging the EVM and Blockchain apps.

apps/ex_wire/lib/ex_wire/packet/status.ex
hayesgm: Hm, should it just be `genesis.hash` instead of `genesis.parent_hash`? I honestly don't know which is correct.

apps/ex_wire/lib/ex_wire/packet/get_block_headers.ex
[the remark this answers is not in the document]
hayesgm: Sure, that makes sense.

apps/ex_wire/lib/ex_wire/packet/status.ex
willmeister: I'm not sure either. If `Genesis` had a `block_hash` ([I don't think it does](https://github.com/mana-ethereum/mana/blob/master/apps/blockchain/lib/blockchain/genesis.ex#L17)), I think it'd be correct. Could use `Genesis.create_block(Chain.t(), TrieStorage.t())` with our trie or an empty one. I can update to that if you think it makes sense.

PR #648, opened 2018-11-20
cast messages quicker

InoMurko:


apps/ex_wire/lib/ex_wire/peer_supervisor.ex
```diff
@@ -67,8 +74,9 @@ defmodule ExWire.PeerSupervisor do
 
         Server.send_packet(child, packet)
       end
+    end)
 
-    if Enum.member?(results, :ok), do: :ok, else: :unsent
```
germsvel: It seems we're losing the feedback of whether or not any children received the message. Do we want to use something like `Task.async/await` or `Task.async_stream/3` to handle that?

InoMurko: Cast is by default sending async messages and it always returns :ok, no matter whether the process exists or if the message was not delivered at all. The code is written in a way that doesn't imply that, hence my fix. Whether we need ack for the message and making the calls synchronous should not be part of this PR. cc @willmeister

germsvel: Cool. That sounds good to me.

PR #659, opened 2018-11-22
get supervision tree based on http and ws configuration

InoMurko:
This is matching Parity and Geth functionality where they allow passing binding interface from the CLI.

```
    --jsonrpc-port=[PORT]
        Specify the port portion of the HTTP JSON-RPC API server. (default: 8545)

    --jsonrpc-interface=[IP]
        Specify the hostname portion of the HTTP JSON-RPC API server, IP should be an interface's IP address, or all
        (all interfaces) or local. (default: local)

--ws-port=[PORT]
        Specify the port portion of the WebSockets JSON-RPC server. (default: 8546)

    --ws-interface=[IP]
        Specify the hostname portion of the WebSockets JSON-RPC server, IP should be an interface's IP address, or all
        (all interfaces) or local. (default: local)
```

### DIFF
--- a/apps/jsonrpc2/config/config.exs
+++ b/apps/jsonrpc2/config/config.exs
@@ -22,8 +22,8 @@ mana_version =
 
 config :jsonrpc2,
   ipc: [enabled: true, path: Path.join([System.user_home!(), ".ethereum", "mana.ipc"])],
-  http: [enabled: true, port: 3999, interface: :local],
-  ws: [enabled: true, port: 4000, interface: :all],
+  http: [enabled: true, port: 3999, interface: :local, max_connections: 10],
+  ws: [enabled: true, port: 4000, interface: :all, max_connections: 10],
   mana_version: mana_version
 
 import_config "#{Mix.env()}.exs"
--- a/apps/jsonrpc2/config/config.exs
+++ b/apps/jsonrpc2/config/config.exs
@@ -22,8 +22,8 @@ mana_version =
 
 config :jsonrpc2,
   ipc: [enabled: true, path: Path.join([System.user_home!(), ".ethereum", "mana.ipc"])],
-  http: [enabled: true, port: 3999, interface: :local, max_connections: 10],
-  ws: [enabled: true, port: 4000, interface: :all, max_connections: 10],
+  http: [enabled: true, port: 8545, interface: :local, max_connections: 10],
+  ws: [enabled: true, port: 8545, interface: :local, max_connections: 10],
   mana_version: mana_version
 
 import_config "#{Mix.env()}.exs"
--- a/apps/jsonrpc2/config/config.exs
+++ b/apps/jsonrpc2/config/config.exs
@@ -22,8 +22,8 @@ mana_version =
 
 config :jsonrpc2,
   ipc: [enabled: true, path: Path.join([System.user_home!(), ".ethereum", "mana.ipc"])],
-  http: [enabled: true, port: 4000],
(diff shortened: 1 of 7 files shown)


InoMurko:
The last commit adds Parity options for setting max concurrency on specific protocol (HTTP,WS).
These are just placeholders for CLI API parameters that @cleishm is working on.
```
--ws-max-connections=[CONN]
        Maximum number of allowed concurrent WebSockets JSON-RPC connections. (default: 100)
--jsonrpc-server-threads=[NUM]
        Enables multiple threads handling incoming connections for HTTP JSON-RPC server.
```

apps/jsonrpc2/lib/jsonrpc2/servers/websocket_http.ex
```diff
@@ -20,26 +228,43 @@ defmodule JSONRPC2.Servers.WebSocketHTTP do
...
+
+  @spec child_spec(:web_ws | :web | :ws, module, configuration) :: Supervisor.child_spec()
+  defp child_spec(type, handler, configuration) do
+    port = Map.get(configuration, :port)
+
+    ip =
+      case Map.get(configuration, :interface) do
+        :local -> {0, 0, 0, 0}
+        :all -> :any
+        ip -> ip
```
ayrat555: Are you validating provided ip anywhere?

apps/jsonrpc2/mix.exs
```diff
@@ -18,7 +18,7 @@ defmodule JSONRPC2.Mixfile do
       description: description(),
       name: "JSONRPC2",
       elixirc_paths: elixirc_paths(Mix.env()),
-      elixirc_options: [warnings_as_errors: true],
+      # elixirc_options: [warnings_as_errors: true],
```
ayrat555: I think this line should be uncommented, right?

apps/jsonrpc2/lib/jsonrpc2/servers/websocket_http.ex
InoMurko: No. I decided that input validations should be validated beforehand on the CLI argument parsing.

apps/jsonrpc2/config/config.exs
```diff
@@ -22,8 +22,8 @@ mana_version =
 
 config :jsonrpc2,
   ipc: [enabled: true, path: Path.join([System.user_home!(), ".ethereum", "mana.ipc"])],
-  http: [enabled: true, port: 4000],
-  ws: [enabled: true, port: 4000],
+  http: [enabled: true, port: 3999, interface: :local, max_connections: 10],
```
ayrat555: maybe we should use port `8545` like parity and geth?

apps/jsonrpc2/lib/jsonrpc2.ex
```diff
@@ -38,60 +38,27 @@ defmodule JSONRPC2 do
...
       strategy: :one_for_one
     )
   end
 
-  @spec get_ipc_child(Keyword.t()) :: Supervisor.child_spec() | nil
+  @spec get_ipc_child(Keyword.t()) :: Supervisor.child_spec() | []
   defp get_ipc_child(enabled: true, path: path) do
     dirname = Path.dirname(path)
     :ok = File.mkdir_p(dirname)
     _ = File.rm(path)
```
ayrat555: why are we creating and then deleting this?

InoMurko: this way we ensure that the socket isn't used and that connections get dropped

PR #663, opened 2018-11-26
Add proof-of-work algorithm

MattMSumner:
Implements part of #541 

Why:

We would like to be able to verify the results given by other nodes when
adding a new block that the required work had been done to do so.

This PR:

Implements the Proof-of-work algorithm described in Appendix J of the
yellow paper (Byzantium edition).

Note: We've only implemented the full version and have added a pending
test for the light version. This will be implemented in a follow up PR.

### DIFF
--- a/apps/blockchain/lib/blockchain/ethash.ex
+++ b/apps/blockchain/lib/blockchain/ethash.ex
@@ -19,7 +19,10 @@ defmodule Blockchain.Ethash do
   @j_cacherounds 3
   @j_parents 256
   @j_wordbytes 4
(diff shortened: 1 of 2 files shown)


InoMurko:
I wonder how much will Ethhash be difficult to use since there's lots of computation and this isn't Elixir's strong side. WDYT @MattMSumner ?

ayrat555:
Can we use it somewhere in the project?

germsvel:
I think we can use it to validate block headers, but we should implement the light version first.

PR #664, opened 2018-11-26
Adds pow_light function to Ethash

germsvel:
This relates to #541

Description
===========

Adds the light version of the Proof-of-work Ethash function.

This version uses an in-memory cache version for data aggregation instead of the full dataset.

Note: We extract the common functionality of `Ethash.pow_full` and `Ethash.pow_light` into a private `pow` function.

Wiki description
----------------

We closely follow the implementation outlined in the [wiki](https://github.com/ethereum/wiki/wiki/Ethash#main-loop), so I include a copy of it here for convenience. 

screen shot 2018-11-26 at 3 42 42 pm


ayrat555:
@germsvel I think we can you this version for validation?

germsvel:
@ayrat555 yes, we should be able to use this one for validation. 🎉

PR #665, opened 2018-11-27
Optimize Ethash cache generation in two places

germsvel:
This is part of #541 

Why?
===

When attempting to run the ethereum common tests for the Ethash PoW, it became apparent that the cache generation is slow.

Note: the tests are not included in this commit because they are not yet passing since they take too long.

What?
=====

This commit includes two optimizations:

1. The initial cache generation (based on a seed hash) recurses N times, and each element is the Keccak 512 of the previous element. Previous to this commit, we were starting with the first element and going to element N, without keeping an accumulated list of elements. So element N would generate all previous elements only for its own calculation. Now we simply keep the values so that we only run through the initial cache generation step N times. (where N is cache_size/@j_hasbytes).

2. When performing the RandMemoHash algorithm, we were using an elixir List since that is the format of the cache. But we must constantly update that list as part of the RandMemoHash algorithm. Since getting and updating elements of a linked list is slow, we now transform the cache into a Map with the indexes as keys. This allows for faster accessing and updating of the cache.

### DIFF
--- a/apps/blockchain/lib/blockchain/ethash.ex
+++ b/apps/blockchain/lib/blockchain/ethash.ex
@@ -301,14 +301,22 @@ defmodule Blockchain.Ethash do
   defp initial_cache(seed, cache_size) do
     adjusted_cache_size = div(cache_size, @j_hashbytes)
 
-    for i <- 0..(adjusted_cache_size - 1) do
-      cache_element(i, seed)
-    end
+    do_initial_cache(0, adjusted_cache_size - 1, seed, [])
   end
 
-  defp cache_element(0, seed), do: Keccak.kec512(seed)
+  defp do_initial_cache(limit, limit, _seed, acc = [previous | _rest]) do
+    result = Keccak.kec512(previous)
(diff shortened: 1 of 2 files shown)


MattMSumner:
Before:
```
iex(1)> :fprof.apply(Blockchain.Ethash, :generate_cache, [<<0::256>>, 64*1000])
iex(2)> :fprof.profile
iex(3)> :fprof.analyse([callers: true, sort: :own, totals: true, details: true])
%                                               CNT       ACC       OWN
[{ totals,                                     6477384,12689.872,12628.531}].
%%%
```
After:

```
iex(1)> :fprof.apply(Blockchain.Ethash, :generate_cache, [<<0::256>>, 64*1000])
iex(2)> :fprof.profile
iex(3)> :fprof.analyse([callers: true, sort: :own, totals: true, details: true])
%                                               CNT       ACC       OWN
[{ totals,                                     81147,  212.267,  211.467}].
%%%
```

PR #667, opened 2018-11-28
Adjust CircleCI caching

cleishm:
* Removes the env cache (it was not being saved anywhere, so would either always be missing or
always out of date).
* Only caches `deps` based on the `mix.lock` checksum.
* Stops caching `_build` between builds, just pass it between jobs using `persist_to_workspace`.


cleishm:
I'll be interested to see what effect this has on build times. Hopefully little!

PR #670, opened 2018-11-28
Optimize Ethash proof-of-work

germsvel:
This is part of #541

Description
===========

Using `Enum.at/2` is expensive with lists. We have been using lists throughout the Ethash module because the Yellow Paper and other implementations tend to use arrays. But since we have to access and update those lists constantly, we update the Cache to be a map with indexes for keys. This allows for faster reads and inserts.

When profiling, we found that one of the hotspots was `FNV.hash_lists/2`. We update that function to no longer use `Enum.at` but to recursively walk through each pair of elements instead.

Profiling
======

CNT: the total number of function calls that was found in the trace
ACC: time of the trace from first timestamp to last
OWN: the sum of the execution time in functions found in the trace, not including called functions

Before

screen shot 2018-11-28 at 2 32 45 pm

After

screen shot 2018-11-28 at 2 32 34 pm


apps/blockchain/lib/blockchain/ethash.ex
```diff
@@ -233,7 +232,10 @@ defmodule Blockchain.Ethash do
       |> initialize_mix(index, cache_size)
       |> binary_into_uint32_list()
 
-    uint32_cache = Enum.map(cache, &binary_into_uint32_list/1)
+    uint32_cache =
```
hayesgm: This might also be a good place for `Enum.into/2` which does similar reduction logic

apps/blockchain/lib/blockchain/ethash.ex
```diff
@@ -251,7 +253,16 @@ defmodule Blockchain.Ethash do
       |> FNV.hash(Enum.at(mix, mix_index))
       |> Integer.mod(cache_size)
 
-    Enum.at(modified_cache, cache_index)
+    %{^cache_index => element} = uint32_cache
```
hayesgm: Cool pattern match! Is this different than just `Map.fetch!()`?

apps/blockchain/lib/blockchain/ethash.ex
```diff
@@ -251,7 +253,16 @@ defmodule Blockchain.Ethash do
...
+    element
+  end
+
+  defp cache_into_indexed_map(original_cache) do
+    original_cache
+    |> Enum.with_index()
+    |> Enum.reduce(%{}, fn {v, k}, acc ->
```
hayesgm: `Enum.into(%{})`

apps/blockchain/lib/blockchain/ethash/fnv.ex
```diff
@@ -4,10 +4,14 @@ defmodule Blockchain.Ethash.FNV do
...
-      hash(Enum.at(x, i), Enum.at(y, i))
-    end
+  def hash_lists(x_list, y_list) do
+    do_hash_lists(x_list, y_list, [])
+  end
+
+  defp do_hash_lists([], _, acc), do: Enum.reverse(acc)
+
+  defp do_hash_lists([x | x_rest], [y | y_rest], acc) do
+    do_hash_lists(x_rest, y_rest, [hash(x, y) | acc])
```
hayesgm: good tail recursion!

apps/blockchain/lib/blockchain/ethash.ex
germsvel: Yeah, we had it as, 

```elixir
Enum.map(cache, fn {i, element} -> {i, binary_into_unint32_list(element)} end)
|> Enum.into(%{})
```

but we ended up going with this to reduce iterations through the cache. That being said, with your comment, I just took a look at `Enum.into/3` which is exactly what I think we need. So I'll update that.

apps/blockchain/lib/blockchain/ethash.ex
germsvel: It's a good point. I prefer `Map.fetch!` because I think it's more legible in this case, but I was going with the pattern matching because I had read that [pattern matching was faster than fetch](https://medium.com/learn-elixir/speed-up-data-access-in-elixir-842617030514) (though I don't know how reliable those estimates are). But now that I'm looking back at that blog post, the difference was with `Map.fetch` not `Map.fetch!` because the former creates an extra tuple `{:ok, result}`. I'll dig a little bit more into it and see if there's a difference.

apps/blockchain/lib/blockchain/ethash.ex
germsvel: Updated in 2fb5acfcb8b0e96d75662e12e880719bd1a69fd8. It's not only more legible but seems to be faster too! Great call 👍

apps/blockchain/lib/blockchain/ethash.ex
germsvel: Updated in https://github.com/mana-ethereum/mana/pull/670/commits/2fb5acfcb8b0e96d75662e12e880719bd1a69fd8. More legible and seems to be faster.

germsvel:
The latest profiling info 

screen shot 2018-11-29 at 9 02 31 am

PR #674, opened 2018-11-29
Add Exth Helper App

hayesgm:
Currently, we have a variety of functions (like Ethereum specific helpers for encoding/decoding hex, etc.) that do not have a home, or live in the project where they happened to be first used. This creates a lot of repeat code that ends up being potentially bugging. Here, we create a new app, Exth, which can be used for a variety of shared utilities. Currently, we simply have some debug functions and hex encoders, but we could move in functions from other areas that do not necessarily belong anywhere else. Again, these functions are specific to Ethereum/mana, but not to any other sub-project in particular.


hayesgm:
Yeah, the hope is that this does not become a dumping ground. The functions in here should be 1) specific to mana, 2) not belonging to any other sub-project, and 3) used in more than one sub-project. So the biggest qualification is that we have a function that we've built for one project and need in another and the cross-dependency does not apply. For instance, hex encoding/decoding falls well into this.

Reasons to not include: 
1. not "specific to mana" - if it's not specific to mana, it should be its own app, or maybe just a module in a sub-app that could later become a full app.
2. "belonging to another sub-project" - If we're talking about, say, whether or not a function reverts, that should be in the EVM app. If we add another app called `eWASM` and it needs shared functionality, they need to figure out their own dependency tree instead of putting it `Exth` since it's still super specific.
3. not "used in more than one sub-project" - just make it a module in that sub-project and move on until it needs to be used elsewhere.

apps/evm/lib/block/header.ex
```diff
@@ -183,11 +183,11 @@ defmodule Block.Header do
       transactions_root: transactions_root,
       receipts_root: receipts_root,
       logs_bloom: logs_bloom,
-      difficulty: :binary.decode_unsigned(difficulty),
```
InoMurko: since this is getting called a lot, are you sure it makes sense to wrap it in a if statement wrapper?

hayesgm: If statement? Do you mean the guard-based  check to verify its type?

The problem is this:

```elixir
[5, <<5>>] |> ExRLP.encode |> ExRLP.decode
[<<5>>, <<5>>]
```

ExRLP can't determine if a value was a binary or a decimal. So, we could always decode, but then it breaks code that does this:

```
MyPacket.serialize() # serialize returns an RLP-encodable structure
[5, 5, 6]
MyPacket.deserialize([5, 5, 6])
** (ArgumentError) argument error
    (stdlib) :binary.decode_unsigned(5)
```

We'd expect serialize and deserialize to work.

So maybe we could enforce that serialize always returns binaries, but that's not technically a requirement of the RLP specification... and would tend to make our lives harder.

So instead.... we just add this check and decode only if it's not already an integer.

InoMurko: Ahaaa. I didn't know it was breaking! Looking at the old code I thought we *knew* it's a binary and that we're now just double checking it is :)

PR #680, opened 2018-11-29
Peer mngmt

InoMurko:
- Removing configurable process names (?).
- Spreading process state over modules.
- Removed keyword type server name passing


apps/ex_wire/lib/ex_wire/kademlia.ex
```diff
@@ -1,85 +1,199 @@
 defmodule ExWire.Kademlia do
   @moduledoc """
-  Handles Kademlia algorithm state.
+  Handles Kademlia algorithm state. also a GenServer that manages Kademlia state.
```
ayrat555: I think the separation of GenServer and domain logic is cleaner so I'm against these changes

ayrat555: See https://pragdave.me/blog/2017/07/13/decoupling-interface-and-implementation-in-elixir.html

InoMurko: I do not see any domain logic here, it's just spreading the process state over several modules. which is  what you would want to avoid (in my experience).

The other part that was also addressed was awkward server name passing in a Keyword type. Not used for anything else...

InoMurko: and if you read closely to the link @ayrat555 this is not how they propose separating domain logic and gen servers. In their case, the beginning of the server would have had to been here: https://github.com/mana-ethereum/mana/blob/master/apps/ex_wire/lib/ex_wire/kademlia.ex
but it's not.

InoMurko:
#681 is another approach

apps/ex_wire/lib/ex_wire/kademlia.ex
hayesgm: Can't we split the difference and keep the server and logic separate, but now use a globally registered name (e.g. `@server_name __MODULE__`) to prevent having to thread the options through?

InoMurko: something like #681?

hayesgm: Yes, I think I like that a lot more :)

InoMurko:
in fav of #681

PR #681, opened 2018-11-29
Peer mngmt - 2

InoMurko:
- Removing configurable process names (?).
- Spreading process state over modules.
- Removed keyword type server name passing

An alternative to #680


apps/ex_wire/config/config.exs
```diff
@@ -39,7 +39,7 @@ config :ex_wire,
   discovery: true,
   node_discovery: [
     network_adapter: {ExWire.Adapter.UDP, NetworkClient},
-    kademlia_process_name: KademliaState,
+    # kademlia_process_name: KademliaState,
```
hayesgm: TODO: Remove

ayrat555:
this looks great!

PR #683, opened 2018-11-29
Wire Sub-Protocol Message ID negotiation 638

### LINKED ISSUE #638: Correct Packet Identifiers
From https://github.com/ethereum/devp2p/blob/master/devp2p.md#message-contents

> devp2p is designed to support arbitrary sub-protocols (aka capabilities) over the basic wire protocol. Each sub-protocol is given as much of the message-ID space as it needs. All such protocols must statically specify how many message IDs they require. On connection and reception of the Hello message, both peers have equivalent information about what subprotocols they share (including versions) and are able to form consensus over the composition of message ID space.

> Message IDs are assumed to be compact from ID 0x10 onwards (0x00-0x10 is reserved for devp2p messages) and given to each shared (equal-version, equal name) sub-protocol in alphabetic order. Sub-protocols that are not shared are ignored. If multiple versions are shared of the same (equal name) sub-protocol, the numerically highest wins, others are ignored.

We, previously, incorrectly assumed that packet ids had a fixed id, as defined in `ExWire.Packet` under `@packet_types`. It packet ids are based on the capabilities of the peer, and the algorithm in other clients goes like this:

```
 Let index=0, offset=0x10
   for (cap, index) in sorted(caps):
   let cap.offset = offset
   offset += cap.packet_count
```

We need to change `@packet_types` to link to offers (possibility associated with our capabilities) and send the correct packet to peers with that capability. We should also, then, know when we cannot send a given packet to a peer if that peer is missing the available capabilities. For reference, capability negotiation happens in the `Hello` message, defined [here](https://github.com/ethereum/devp2p/blob/master/devp2p.md#p2p-sub-protocol-messages), which is `ExWire.Packet.Hello` in mana. Also for reference, here are other sub protocols: https://github.com/ethereum/wiki/wiki/Ethereum-Wire-Protocol#ethereum-sub-protocol and Warp sub-protocols: https://wiki.parity.io/Warp-Sync

willmeister:
Fixes #638

# Description
Right now DEVp2p Sub-Protocol Message IDs are hardcoded in the `Packet` behaviour. Per the [specification](https://github.com/ethereum/devp2p/blob/master/devp2p.md#message-contents), Message IDs 0x00 - 0x09 (inclusive) are reserved for p2p messages, but shared Sub-Protocol Capabilities are negotiated by peers with the exchange of `Hello` messages. Message IDs for mutually-supported Sub-Protocols are assigned sequentially starting at 0x10 and sorted by Sub-Protocol string (3-chars) ASC and then by offset within the Sub-Protocol ASC.

This PR creates the concept of Capabilities, containing their specified Messages by version and a negotiated `PacketIdMap` per Peer Session of the shared Capability messages.


apps/ex_wire/lib/ex_wire/packet/capability.ex
```diff
@@ -0,0 +1,71 @@
...
+    Map.has_key?(mana_capabilities_map, name) &&
+      apply(Map.get(mana_capabilities_map, name), :get_packet_types, [version]) !=
+        :unsupported_version
+  end
+
+  defp sort_asc_name_desc_version(%__MODULE__{name: name1, version: version1}, %__MODULE__{
+         name: name2,
+         version: version2
+       }) do
+    name1 < name2 || (name1 == name2 && version1 > version2)
```
hayesgm: Can we relay exclusively on erlang's sorting and maybe just do `{name1, version1} < {name2, version2}`?

apps/ex_wire/lib/ex_wire/config.ex
```diff
@@ -139,9 +140,15 @@ defmodule ExWire.Config do
     get_env(given_params, :p2p_version)
   end
 
-  @spec caps(Keyword.t()) :: [{String.t(), integer()}]
+  @spec caps(Keyword.t()) :: [Capability.t()]
   def caps(given_params \\ []) do
-    get_env(given_params, :caps)
+    caps = get_env(given_params, :caps)
```
hayesgm: This (by its spec) shouldn't return nil. I'd just add a default value of the empty list `[]` here and then your map function below will work fine on that.

Issue #639, opened 2018-11-20
Peer Management


hayesgm:
Currently when we run our sync code, we start two supervisors: `ExWire.PeerSupervisor` and `ExWire.TCP.InboundConnectionsSupervisor`. The latter is used for inbound connections, which isn't heavily supported right now. For `ExWire.PeerSupervisor`, when we're running discovery mode, we connect to peers as they are discovered until we reach a certain count of max peers (currently hard-coded). We don't try to connect to other peers when our connected peers drop, and again, we ignore any inbound connections. This is pretty poor behavior.

This issue is to improve peer connection management. It should do a variety of tasks:

1. Keep track of all peers (discovered by the discovery protocol, inbound or from bootnodes / reserved peers), even if we decide not to connect. This acts as a pool of nodes we might connect to.
2. Connect to good peers. This might be from the Kademlia distance formula, or maybe nodes we've connected to in the past, or maybe based on capabilities (e.g. connect, check the caps, and drop the node if it's not useful).
3. Keep track of dropped connections (maybe a peer disconnected from us) and connect to new nodes when this happens. We should (probably) not reconnect to a peer after it asks to disconnect, but we should keep that peer if we just had a transport error.

This will require a good amount of work, but it's vital to getting a healthy node up and running.

PR #688, opened 2018-11-30
Peer mngmt #639

InoMurko:
This is the first step to establish exact requirements of #639.
What this does at the moment is that it passes a _observer_ module down the supervision tree to workers. The workers are inbound and outbound connections. These worker processes link themselves to the _observer_ module, so that in case of disconnect, the observer can catch the exit and:
1. store the peer we were connected to, to one of the mapsets
2. check the reason for the disconnect or process crash (while it also has access to the process state)

There's no logic yet when a connection gets dropped  (either from ur or from them) - because I'm still figuring out what the logic should be. Like ... based on what should we pick a new peer?

The other part was that I removed the link between the PeerSupervisor and Kademlia. Kademlia now notifies the _observer_ module of rounds and the flow of attaching dynamic children to outbound connections tree.


.gitignore
```diff
@@ -22,3 +22,4 @@ apps/blockchain/config/dev.secret.exs
 .erlang_cookie
 
 !/**/.gitkeep
+/.vscode
```
ayrat555: I think adding editor specific files to gitignore is bad practice

apps/ex_wire/config/test.exs
```diff
@@ -10,6 +10,7 @@ config :ex_wire,
     port: 30_399
   ],
   sync_mock: ExWire.BridgeSyncMock,
+  kademlia_mock: ExWire.FakeKademlia,
```
ayrat555: I think this parameter shouldn't be called `kademlia_mock`. Maybe just `kademlia`? It will be mocked in test env

InoMurko: The reason I named it this way was to keep consistency with the above.

.gitignore
willmeister: Is there a downside here? If you use the editor, you won't accidentally add editor-specific files, and if you don't, nothing will happen. Maybe we just group editor-specific ignores and put them at the bottom?

apps/ex_wire/lib/ex_wire/p2p/server.ex
```diff
@@ -244,4 +250,7 @@ defmodule ExWire.P2P.Server do
 
     {:ok, new_state}
   end
+
+  @spec link(module()) :: true
+  defp link(connection_observer), do: Process.link(Process.whereis(connection_observer))
```
ayrat555: can we change it to:
```
connection_observer
|> Process.whereis()
|> Process.link()
```

.gitignore
InoMurko: @ayrat555 , @willmeister I removed it (though I don't see much problems with having it) and added it in my local _.git/info/exclude_ file.

apps/ex_wire/config/test.exs
willmeister: I think the idea is that in the non-test environments, the application will handle it properly -- it shouldn't be up to config. With config, you can choose to mock it, but otherwise, config shouldn't be specifying an implementation:
```
@kademlia Application.get_env(:ex_wire, :kademlia_mock, Kademlia)
```
Let me know if we agree to change this, though, because I use this pattern elsewhere and will update that as well.

InoMurko: I do think this isn't the most elegant way to mock, because you can specify only one function behaviour per environment. One of the approaches that gives you full control of the behaviour is to always pass the dependencies between processes down the supervision tree, but it gives much overhead on implementation...
If it was me... I would start using Mock libraries. Like :meck for example. WDYT @ayrat555 @willmeister @hayesgm ?

hayesgm: I *really* dislike Meck, and significantly prefer this variety (or argument-pass) behaviours. Meck changes the system under test too much that it makes it difficult to understand behavior. I generally like doing:

```
@my_lib Application.get_env(:ex_wire, :some_lib, DefaultLib)

def my_function(arg1, arg2, my_lib \\ @my_lib) do
   my_lib.some_function()
end
```

This makes it very easy to inject mock versions of a library in a way that is very clear and predictable. With a behavior on the library, you can also have dialyzer verify that the "mock" version matches the signature of the true library. Additionally, it enforces that you should only mock when you *truly* have a library that is only useful when it is side-effecting. The reduction in side-effects and push toward pure functions is a bonus.

PR #689, opened 2018-11-30
Wire Sub-Protocol Capability Configuration

willmeister:
Fixes an issue where we blindly pull our capabilities list from config, instead starting with the capabilities that we _can_ support and paring them down based on what is configured.

Also adds logic to `ExWire.Packet.Capability.Eth.Hello` to only recommend activating peers if they have shared capabilities.

### DIFF
--- a/apps/ex_wire/lib/ex_wire/packet/capability/eth.ex
+++ b/apps/ex_wire/lib/ex_wire/packet/capability/eth.ex
@@ -1,8 +1,11 @@
 defmodule ExWire.Packet.Capability.Eth do
+  alias ExWire.Config
   alias ExWire.Packet.Capability
   alias ExWire.Packet.Capability.Eth
 
   @behaviour Capability
+
+  @name "eth"
 
   @version_to_packet_types %{
     62 => [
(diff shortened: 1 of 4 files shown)


apps/ex_wire/lib/ex_wire/packet/capability/eth.ex
```diff
@@ -17,11 +20,20 @@ defmodule ExWire.Packet.Capability.Eth do
     ]
   }
 
-  @supported_versions Map.keys(@version_to_packet_types)
+  @supported_versions MapSet.to_list(
```
InoMurko: what exactly is happening here? 
could this be a function that you're able to test instead of a module attribute?

willmeister: Just getting the intersection of the Capabilities that we support and the Capabilities that are configured. I could make it a function, but I wanted it to be calculated once at compile time instead of on each call.

InoMurko: seems somewhat stretching a module attribute. IDK, I have never used them for something else then simple values. if you're willing to bake them in the module... you could also write them by hand?

InoMurko:
```
a = Map.keys(version_to_packet_types)
b = Config.caps()|> |> Enum.filter(fn cap -> cap.name == name end) |> Enum.map(fn cap -> cap.version end)
'>' = [62] = a -- (a -- b)
```

InoMurko: after we discussed offline, I think it also makes sense to leave them as attributes.

apps/ex_wire/lib/ex_wire/packet/capability/mana.ex
```diff
@@ -1,11 +1,26 @@
...
-  @our_capabilities %{
+  @our_capabilities_map %{
     Eth.get_name() => Eth
   }
 
+  @our_capabilities @our_capabilities_map
+                    |> Enum.map(fn {name, mod} ->
+                      versions = apply(mod, :get_supported_versions, [])
+
+                      versions
```
hayesgm: This should probably be not pipelined, or:

```elixir
for version <- apply(mod, :get_supported_versions, []) do
  Capability.new({name, version})
end
```

PR #690, opened 2018-12-01
Fix Frame Size Issue

hayesgm:
When packets come in, sometimes they are incomplete. We previously handled packets that had a valid header but insufficient body size. This patch adds support for packets that don't even have complete headers. We queue the packet into queued data and wait for more data before processing.


apps/ex_wire/lib/ex_wire/framing/frame.ex
```diff
@@ -107,80 +112,84 @@ defmodule ExWire.Framing.Frame do
...
-
-          {
-            :ok,
-            packet_type_rlp |> ExRLP.decode() |> :binary.decode_unsigned(),
-            packet_data_rlp |> ExRLP.decode(),
-            frame_rest,
-            %{frame_secrets | ingress_mac: ingress_mac, decoder_stream: decoder_stream}
-          }
+            frame_enc::binary-size(frame_size),
+            frame_padding::binary-size(frame_padding_bytes),
```
willmeister: Should this just be `frame_enc_with_padding::binary-size(frame_size + frame_padding_bytes)`? Looks like they're pattern matched separately only to be concatenated below and never used individually.

InoMurko:
I've noticed that binary isn't a dependency of ex_wire, could you include it? `{:binary, "~> 0.0.5"}`

hayesgm:
I addressed the PR feedback, ready for merge.

PR #692, opened 2018-12-01
Connect Warp Sync to Proper Packets

hayesgm:
This patch connects Warp Sync to the correct packets that come back from Parity when warp syncing. There were a few issues with the algorithm for determining packet ids. It appears that each capability has a size (to allow for protocol upgrades) and we skip that many before the next protocol. These aren't documented from come by Parity's source code. Additionally, it appears that once you negotiate a protocol, you still use the offests from Pv62 but from the higher negotiated ids. It's all a bit weird, but this seems to get everything working. Finally, we had an issue with default map getting called *a lot*, so we instead opt to hard-code the value and try to pass nil around instead of default map.

### DIFF
--- a/apps/ex_wire/lib/ex_wire/dev_p2p/session.ex
+++ b/apps/ex_wire/lib/ex_wire/dev_p2p/session.ex
@@ -65,7 +65,7 @@ defmodule ExWire.DEVp2p.Session do
   """
   @spec disconnect(t) :: t
   def disconnect(session = %__MODULE__{}) do
-    %{session | hello_sent: nil, hello_received: nil, packet_id_map: PacketIdMap.default_map()}
+    %{session | hello_sent: nil, hello_received: nil, packet_id_map: nil}
   end
 
   @doc """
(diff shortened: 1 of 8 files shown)


apps/ex_wire/lib/ex_wire/dev_p2p/session.ex
```diff
@@ -100,6 +100,6 @@ defmodule ExWire.DEVp2p.Session do
   """
   @spec compatible_capabilities?(t) :: boolean()
   def compatible_capabilities?(%__MODULE__{packet_id_map: packet_id_map}) do
-    packet_id_map != PacketIdMap.default_map()
+    Map.has_key?(packet_id_map.ids_to_modules, 0x10)
```
willmeister: Hardcoding 0x10 here presumes to know how `PacketIdMap` works. Pretty safe assumption, but might work better as a function of `PacketIdMap`.

hayesgm: Yeah, and I don't think this PR has tests passing yet, since I wanted to get some design review. I wasn't sure what's the best way to say "We have compatibility" but having a 0x10 status packet seems to be pretty common on protocols. Also, it's faster to verify than the previous solution, even if possibly less accurate.

hayesgm:
@willmeister Was this otherwise addressed in a different PR that you made?

PR #695, opened 2018-12-03
log to disk

InoMurko:
CLI should offer configuration of logger level (TBD in a later PR).

FYI, this logger backend does not support log rotation which is kinda disappointing. If it's a requirement, we should consider Lager.

### DIFF
--- a/.formatter.exs
+++ b/.formatter.exs
@@ -3,6 +3,7 @@
     "apps/*/{lib,config,test}/**/*.{ex,exs}",
     "apps/*/mix.exs",
     "rel/config.exs",
-    "config/*"
+    "config/*",
+    "mix.exs"
   ]
 ]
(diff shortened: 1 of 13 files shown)


ayrat555:
why do you think we need to write logs to disk now? there is no monitoring, no log rotation etc

InoMurko:
It's meant more for us to think what the approach is, start thinking about it. I've personally used it to have more control of crash reports when connections drop... so I thought others would find it useful too. _I don't necessarily need it in master._

As you said and I agree, log rotation is needed. I don't know of other (good enough) approaches then to leverage Lager.
Monitoring has it's own separate ticket. We could decide to drop this and include it in the monitoring ticket and do it properly with log rotation...

hayesgm:
I guess we all have different opinions, I tend to think it's better to have your system manage log rotation instead of built into the binary. Either way, looks good.

PR #700, opened 2018-12-04
fix block hash key

ayrat555:
while working on https://github.com/mana-ethereum/mana/pull/677 I found that block hash keys are generated with `}` at the end

Example
`"hash_for_1}"`


willmeister:
Good catch!

PR #701, opened 2018-12-04
getting p2p version and caps into the peer

InoMurko:
This expand a Peer struct with p2p_version and caps. 1. Setup for rating and 2. Both Parity and Geth support JSONRPC admin methods where caps is returned for connected peers. 
My next step in the works is to create a separate struct that will have Peer + other related fields that will help us maintain Peers rating, but related just to ratings.


hayesgm:
Looks awesome

PR #704, opened 2018-12-05
set time on struct init, server state bugs

InoMurko:


apps/ex_wire/lib/ex_wire/p2p/manager.ex
```diff
@@ -21,28 +21,28 @@ defmodule ExWire.P2P.Manager do
   and a `peer` to be provided. This function starts the encrypted handshake with
   the `peer`.
   """
-  @spec new_outbound_connection(ExWire.TCP.socket(), Peer.t()) :: Connection.t()
-  def new_outbound_connection(socket, peer) do
+  @spec new_outbound_connection(Connection.t()) :: Connection.t()
+  def new_outbound_connection(state) do
```
willmeister: `conn` or something related to Connection would be a better name than `state` here and below.

apps/ex_wire/lib/ex_wire/p2p/connection.ex
```diff
@@ -22,7 +22,8 @@ defmodule ExWire.P2P.Connection do
           sent_message_count: integer(),
           datas: [binary()],
           last_error: any() | nil,
-          is_outbound: boolean()
+          is_outbound: boolean(),
+          timer: Time.t() | nil
```
willmeister: This is not really a timer, instead a time. Might be better to name it `created_time`, or something else based on how it's used.

apps/ex_wire/lib/ex_wire/p2p/server.ex
```diff
@@ -147,27 +158,24 @@ defmodule ExWire.P2P.Server do
...
     :ok =
       Logger.debug(fn ->
         "[Network] [#{peer}] Established outbound connection with #{peer.host_name}."
       end)
 
-    new_conn =
-      socket
-      |> Manager.new_outbound_connection(peer)
-      |> Map.put(:subscribers, Map.get(opts, :subscribers, []))
+    state0 = Manager.new_outbound_connection(%{state | socket: socket0})
```
willmeister: Would prefer not numbering variables or at least making them sequential. `state0` seems like it would be the state before `state`.

InoMurko: Indeed, this is an error.

PR #709, opened 2018-12-06
add json rpc fields

ayrat555:
resolves https://github.com/mana-ethereum/mana/issues/707


apps/blockchain/lib/blockchain/block.ex
```diff
@@ -23,18 +23,21 @@ defmodule Blockchain.Block do
...
   # transactions: B_T,
   # ommers: B_U
+  # additonal_info: precomputated data required by JSON RPC spec
   defstruct block_hash: nil,
             header: %Header{},
             transactions: [],
             receipts: [],
-            ommers: []
+            ommers: [],
+            additional_info: nil
```
willmeister: would `metadata` be a better name for this? If not, is there anything more specific than `additional_info` that would fit here?

apps/blockchain/lib/blockchain/block.ex
```diff
@@ -279,12 +284,40 @@ defmodule Blockchain.Block do
...
+    end
+  end
+
+  def get_block_with_additional_info(number, trie) when is_integer(number) do
+    with {:ok, hash} <- get_block_hash_by_number(trie, number) do
+      get_block_with_additional_info(hash, trie)
+    end
+  end
+
+  @spec get_additional_info(EVM.hash(), TrieStorage.t()) :: {:ok, t()} | :not_found
```
willmeister: This doesn't return `{:ok, t()}` right?

apps/blockchain/lib/blockchain/block.ex
```diff
@@ -255,12 +258,14 @@ defmodule Blockchain.Block do
...
     hash = if predefined_key, do: predefined_key, else: hash(block)
+
     block_rlp = block |> serialize |> ExRLP.encode()
 
     updated_trie =
       trie
       |> TrieStorage.put_raw_key!(hash, block_rlp)
-      |> TrieStorage.put_raw_key!(get_block_hash_key(block.header.number), hash)
+      |> TrieStorage.put_raw_key!(block_hash_key(block.header.number), hash)
+      |> put_additional_info(hash, block, block_rlp)
```
willmeister: Since this is also indexed by block hash, should we just store blocks as `[block_rlp, additional_info]`? Since I imagine they'll be fetched sequentially frequently, might consider just storing them together to prevent additional calls to disk to store / lookup. 

This would also allow you to add the additional properties directly to `t()` instead of as nullable `additional_info`

ayrat555: I'll address it in the next PR. We won't store RLP anymore

PR #713, opened 2018-12-07
store block as erlang term

ayrat555:
Changes:
- store blocks as erlang terms instead of RLP encoding
- add JSON RPC fields (size, total_difficulty) to header struct instead of metadata field


apps/blockchain/lib/blockchain/block.ex
```diff
@@ -940,4 +885,27 @@ defmodule Blockchain.Block do
...
+  defp add_metadata(block, trie, predefined_hash) do
+    block_rlp_size =
+      block
+      |> serialize
+      |> ExRLP.encode()
+      |> byte_size()
+
+    total_difficulty =
+      case get_block(block.header.parent_hash, trie) do
+        {:ok, block} ->
```
willmeister: would change this to `{:ok, parent_block}` because the `block` in this case is different than the `block` in the other case.

PR #714, opened 2018-12-10
JSON RPC transaction

ayrat555:
Changes:
- Encode Transaction to JSON RPC specification format

Part of https://github.com/mana-ethereum/mana/issues/675


apps/jsonrpc2/lib/jsonrpc2/response/transaction.ex
```diff
@@ -0,0 +1,67 @@
...
+    :nonce,
+    :to,
+    :transactionIndex,
+    :value,
+    :v,
+    :r,
+    :s
+  ]
+
+  def new(internal_transaction, internal_block, network_id \\ nil) do
```
willmeister: `@spec` would be nice here

apps/jsonrpc2/lib/jsonrpc2/response/transaction.ex
```diff
@@ -0,0 +1,67 @@
...
+
+    encode_hex(index)
+  end
+
+  defp hash(internal_transaction) do
+    internal_transaction
+    |> Transaction.serialize()
+    |> ExRLP.encode()
+    |> Keccak.kec()
+    |> encode_hex
```
willmeister: Might suggest adding `()` for consistency with `serialize()`, `encode()`, and `kec()`

apps/jsonrpc2/test/jsonrpc2/response/block_test.exs
```diff
@@ -2,11 +2,12 @@ defmodule JSONRPC2.Response.BlockTest do
   use ExUnit.Case, async: true
 
   alias JSONRPC2.Response.Block
-  alias JSONRPC2.TestFactory
+
+  import JSONRPC2.TestFactory
```
willmeister: Does `import` give us any benefits over `alias`? If not, it's a bit clearer to `alias` because then someone unfamiliar with the code knows where `build` is defined.

PR #716, opened 2018-12-11
add transaction getter methods for JSON RPC API

ayrat555:
Changes:
- Implement `eth_getTransactionByBlockNumberAndIndex` JSON RPC method
- Implement `eth_getTransactionByBlockHashAndIndex` JSON RPC method

part of https://github.com/mana-ethereum/mana/issues/675

### DIFF
--- a/apps/jsonrpc2/lib/bridge/sync.ex
+++ b/apps/jsonrpc2/lib/bridge/sync.ex
@@ -4,6 +4,7 @@ defmodule JSONRPC2.Bridge.Sync do
   alias ExWire.PeerSupervisor
   alias ExWire.Sync
   alias JSONRPC2.Response.Block, as: ResponseBlock
+  alias JSONRPC2.Response.Transaction, as: ResponseTransaction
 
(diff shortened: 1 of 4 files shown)


apps/jsonrpc2/lib/bridge/sync.ex
```diff
@@ -45,4 +46,30 @@ defmodule JSONRPC2.Bridge.Sync do
       _ -> nil
     end
   end
+
+  def get_transaction_by_block_hash_and_index(block_hash, trx_index) do
```
willmeister: Should these return the `Blockchain.Transaction` instead of `ResponseTransaction`? The caller can call `ResponseTransaction.new` if they wish to create one from it. The answer may be "no, they shouldn't. This works fine."

ayrat555: these methods are used in `SpecHandler` so objects should be serializable with Jason. Our internal structs cannot be serialized with Jason

PR #717, opened 2018-12-11
return transactions with block for JSON RPC API

ayrat555:
Changes:
- return transactions with block in JSON RPC API
- return transaction based on boolean parameter:  If true it returns the full transaction objects, if false only the hashes of the transactions.


apps/jsonrpc2/lib/bridge/sync.ex
```diff
@@ -28,20 +28,20 @@ defmodule JSONRPC2.Bridge.Sync do
   @spec get_last_sync_state() :: Sync.state()
   defp get_last_sync_state(), do: Sync.get_state()
 
-  def get_block_by_number(number) do
+  def get_block_by_number(number, full_transactions) do
```
willmeister: `include_full_transactions` might be clearer here and throughout.

apps/jsonrpc2/lib/jsonrpc2/response/block.ex
```diff
@@ -47,13 +73,28 @@ defmodule JSONRPC2.Response.Block do
...
+  defp block_size(block) do
+    block
+    |> Block.serialize()
+    |> ExRLP.encode()
+    |> byte_size()
+  end
+
+  @spec transactions([Transaction.t()], Block.t(), boolean()) ::
+          [ResponseTransaction.t()] | [binary()]
+  def transactions(transactions, block, true) do
```
willmeister: Would rename to `get_transactions` or `format_transactions`

apps/jsonrpc2/lib/jsonrpc2/response/block.ex
```diff
@@ -28,7 +29,32 @@ defmodule JSONRPC2.Response.Block do
...
+          size: binary(),
+          gasLimit: binary(),
+          gasUsed: binary(),
+          timestamp: binary(),
+          transactions: transactions(),
+          uncles: []
+        }
+
+  @spec new(Block.t(), boolean()) :: t()
+  def new(internal_block, full_transactions) do
```
willmeister: Would it be useful for `full_transactions` to have a default? Maybe `false` for backward compatibility?

Issue #706, opened 2018-12-05
Support capabilities that share packets


willmeister:
Right now packet ID map maps IDs -> Packet Modules and Packet Modules -> IDs. If we support Eth 63 and Par 1 capabilities, Par 1 shares a number of the packets from Eth 63, so `Eth.NewBlockHashes` for example would point to both 17 (its ID in Eth), and 34 (its ID in Par), and unfortunately that's not how maps work.

This issue is to either duplicate Eth.NewBlockHashes and the other packets into Par.NewBlockHashes since they only share message format and not handling logic or fix our `PacketIdMap` to support capabilities with shared packets.

PR #720, opened 2018-12-11
Support Capabilities with Overlapping Packets 706

willmeister:
Fixes issue #706 where `eth` and `par` cannot support the same packet types.


ayrat555:
@willmeister can common methods be moved into a separate module?

willmeister:
@ayrat555 yes, we could move common methods, but they're not guaranteed to be common. They just happen to be at the moment. For instance, the `par` capability could choose to re-define parts of the shared `eth` packets and/or how to handle them, at which point we'd have to pull them apart again.

PR #724, opened 2018-12-12
decode block number in json rpc

ayrat555:
Changes:
- block number parameters should be encoded as hex strings in input
- implement `eth_getBlockTransactionCountByHash`
- implement `eth_getBlockTransactionCountByNumber`


apps/jsonrpc2/lib/bridge/sync.ex
```diff
@@ -72,4 +72,34 @@ defmodule JSONRPC2.Bridge.Sync do
...
+        |> Enum.count()
+        |> :binary.encode_unsigned()
+        |> Exth.encode_hex()
+
+      _ ->
+        nil
+    end
+  end
+
+  def get_block_transaction_count_by_number(hash) do
```
willmeister: Should this be `get_block_transaction_count_by_hash`?

ayrat555: @willmeister done

apps/exth/lib/exth.ex
```diff
@@ -91,4 +91,25 @@ defmodule Exth do
...
+      1
+
+      iex> Exth.decode_unsigned_from_hex("9c")
+      156
+
+      iex> Exth.decode_unsigned_from_hex("0x0f4240")
+      1_000_000
+  """
+  @spec decode_unsigned_from_hex(String.t()) :: integer()
+  def decode_unsigned_from_hex(hex_integer) do
```
hayesgm: Great function (since this comes up a lot)!

apps/jsonrpc2/lib/bridge/sync.ex
```diff
@@ -72,4 +72,34 @@ defmodule JSONRPC2.Bridge.Sync do
...
+
+  def get_block_transaction_count_by_number(number) do
+    state_trie = get_last_sync_state().trie
+
+    case Block.get_block_by_number(number, state_trie) do
+      {:ok, block} ->
+        block.transactions
+        |> Enum.count()
+        |> :binary.encode_unsigned()
+        |> Exth.encode_hex()
```
hayesgm: Do we want an `Exth.encode_unsigned_hex` since I'm a little unsure what the returned value should look like. Is this standard Hex or that weird-variety used by Web3 libraries (e.g. with *all* leading zeros removed)?

PR #726, opened 2018-12-13
uncle json rpc methods

ayrat555:
Changes:
- return uncle hashes with block
- implement `eth_getUncleCountByBlockHash`
- implement `eth_getUncleCountByBlockNumber`

### DIFF
--- a/apps/blockchain/lib/blockchain/block.ex
+++ b/apps/blockchain/lib/blockchain/block.ex
@@ -289,6 +289,8 @@ defmodule Blockchain.Block do
     end
   end
 
+  def get_block(nil, _), do: :not_found
+
   @spec block_hash_key(integer()) :: String.t()
   defp block_hash_key(number) do
     "hash_for_#{number}"


apps/jsonrpc2/lib/bridge/sync.ex
```diff
@@ -100,4 +100,32 @@ defmodule JSONRPC2.Bridge.Sync do
...
+        block.ommers
+        |> Enum.count()
+        |> Exth.encode_unsigned_hex()
+
+      _ ->
+        nil
+    end
+  end
+
+  def get_uncle_count_by_block_number(number) do
```
willmeister: Seeing this more, we should replace `Block.get_block_by_number` with 
`Block.get_block(id, trie) when is_integer(id)` and `Block.get_block(id, trie) when is_binary(id)` overloads, so that we don't have to write two functions for all callers of `get_block` / `get_block_by_number`. This is really my fault, as I created `get_block_by_number`.

apps/jsonrpc2/lib/jsonrpc2/response/block.ex
```diff
@@ -108,4 +109,9 @@ defmodule JSONRPC2.Response.Block do
       |> encode_hex()
     end)
   end
+
+  @spec format_uncles([Header.t()]) :: [binary()]
+  def format_uncles(uncles) do
+    Enum.map(uncles, fn uncle -> Header.hash(uncle) end)
```
willmeister: Can be `Enum.map(uncles, &Header.hash/1)`, right?

PR #732, opened 2018-12-17
return transaction by hash

ayrat555:
Changes:
- Implement `eth_getTransactionByHash` JSON RPC method

### DIFF
--- a/apps/blockchain/lib/blockchain/block.ex
+++ b/apps/blockchain/lib/blockchain/block.ex
@@ -401,8 +401,8 @@ defmodule Blockchain.Block do
     end
   end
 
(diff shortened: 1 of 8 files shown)


apps/blockchain/lib/blockchain/block.ex
```diff
@@ -906,4 +932,25 @@ defmodule Blockchain.Block do
...
+      location = :erlang.term_to_binary({block.block_hash, idx})
+
+      TrieStorage.put_raw_key!(trie_acc, transaction_key, location)
+    end)
+  end
+
+  defp transaction_key(hash) do
+    hash_hex = Base.encode16(hash, case: :lower)
+
+    "transaction_#{hash_hex}"
```
willmeister: The hash should be unique, right? Probably don't need "transaction_", but it doesn't hurt much.

PR #734, opened 2018-12-18
Fixing warp processor caching trie thread safety issues

willmeister:


apps/merkle_patricia_tree/lib/merkle_patricia_tree/caching_trie.ex
```diff
@@ -37,6 +37,11 @@ defmodule MerklePatriciaTree.CachingTrie do
     }
   end
 
+  @impl true
+  def with_clean_cache(trie) do
+    new(%{trie.trie | root_hash: trie.in_memory_trie.root_hash})
```
ayrat555: @willmeister I'm not sure that this is right. You're creating a new `CachingTrie` with `root_hash` from in-memory trie. But in-memory trie is not committed to the permanent database

willmeister: @ayrat555 yeah, I think you're right. It works here because that `trie` is always empty. I'll either add a `TrieStorage.commit!(trie)` call before the call to `new(trie)` or change the name of this function to indicate that expectation.

PR #736, opened 2018-12-19
encode hex values according to Ethereum JSON RPC spec

ayrat555:
encode hex values according to [Ehereum JSON RPC spec](https://github.com/ethereum/wiki/wiki/JSON-RPC):

When encoding QUANTITIES (integers, numbers): encode as hex, prefix with "0x", the most compact representation (slight exception: zero should be represented as "0x0"). Examples:

    0x41 (65 in decimal)
    0x400 (1024 in decimal)
    WRONG: 0x (should always have at least one digit - zero is "0x0")
    WRONG: 0x0400 (no leading zeroes allowed)
    WRONG: ff (must be prefixed 0x)

When encoding UNFORMATTED DATA (byte arrays, account addresses, hashes, bytecode arrays): encode as hex, prefix with "0x", two hex digits per byte. Examples:

    0x41 (size 1, "A")
    0x004200 (size 3, "\0B\0")
    0x (size 0, "")
    WRONG: 0xf0f0f (must be even number of digits)
    WRONG: 004200 (must be prefixed 0x)


apps/jsonrpc2/lib/jsonrpc2/response/helpers.ex
```diff
@@ -1,19 +1,40 @@
...
-  def encode_hex(binary) when is_binary(binary) do
-    Exth.encode_hex(binary)
+  @moduledoc """
+  Encodes an usigned integer as a binary as described in Ethereum JSON RPC spec.
+  """
+  @spec encode_quantity(binary() | non_neg_integer() | nil) :: binary() | nil
+  def encode_quantity(binary) when is_binary(binary) do
+    hex_binary = Base.encode16(binary, case: :lower)
+
+    case hex_binary do
```
hayesgm: I don't think you need to check this case as it should be covered by the check on line 16

willmeister: I think removing this case would change the logic. I'm not sure if it _should_ change, but I think it would change. For instance, if this is hit, and `hex_binary` is `""`, `""` will be returned. If removed, `"0x0"` will be returned.

ayrat555: @willmeister yep, good concern. but `hex_binary` shouldn't be `""`, because only encoded integers are passed to `encode_quantity` function

PR #745, opened 2018-12-24
improve json rpc tests

ayrat555:
Changes:
- setup database after each test so test don't depend on each other 
- use string interpolation in requests for json rpc tests


apps/jsonrpc2/test/jsonrpc2/mana_handler_test.exs
```diff
@@ -131,18 +132,16 @@ defmodule JSONRPC2.ManaHandlerTest do
 
   describe "eth_getBlockByNumber" do
     test "fetches block by number" do
-      block =
-        TestFactory.build(:block,
-          block_hash: <<0x2::256>>,
-          header: TestFactory.build(:header, number: 10)
-        )
+      block = TestFactory.build(:block)
```
willmeister: Does it make sense to change the function name to `build_block`?

ayrat555: symbols are used in ex_machina https://github.com/thoughtbot/ex_machina

PR #750, opened 2018-12-26
estimate gas JSON RPC method

ayrat555:


apps/blockchain/lib/blockchain/transaction/validity.ex
```diff
@@ -46,21 +54,31 @@ defmodule Blockchain.Transaction.Validity do
...
   end
 
   @spec check_sender_nonce([atom()], Transaction.t(), Account.t()) :: [atom()]
   defp check_sender_nonce(errors, transaction, account) do
-    if account.nonce != transaction.nonce do
-      [:nonce_mismatch | errors]
-    else
-      errors
+    cond do
+      !is_nil(transaction.from) ->
```
willmeister: Can this case be removed and handled by the `true ->` case?

apps/jsonrpc2/lib/jsonrpc2/spec_handler/gas_estimater.ex
```diff
@@ -0,0 +1,83 @@
+defmodule JSONRPC2.SpecHandler.GasEstimater do
+  alias Blockchain.Block
+  alias Blockchain.Transaction
+  alias MerklePatriciaTree.TrieStorage
+
+  @max_gas_limit 50_000_000
```
willmeister: Could be a call to get config with a default value of `50_000_000`

apps/jsonrpc2/lib/jsonrpc2/spec_handler/gas_estimater.ex
```diff
@@ -0,0 +1,83 @@
...
+      {:ok, block} -> {:ok, block}
+      _ -> %{error: "Block is not found"}
+    end
+  end
+
+  defp estimate_gas(state, transaction, gas, block_header, chain) do
+    transaction = %{transaction | gas_limit: gas}
+
+    {_repo, gas_used, receipt} =
+      Transaction.execute_with_validation(state, transaction, block_header, chain)
```
willmeister: It's not necessary now, but I think we can improve the efficiency of this since this gets called `O(log2(upper_bound - lower_bound))` times. Instead of actually executing the transaction, we could just do initial gas validation followed by summing all of the gas costs for the op codes to be executed.

Unnecessary steps include (but are not limited to):
* Checking account & signature validity
* Creating / updating the account
* Contract creation
* Trie access / manipulation

The spec says that estimates may be significantly higher than actual gas costs, so as long as we take the most expensive route when faced with uncertainty, we should be fine.

Issue #637, opened 2018-11-20
Fast Sync


hayesgm:
Turns out, Warp Sync doesn't strictly replace fast sync. Warp sync gets you a node that's up to date for only recent transactions quickly. You still need to sync in the background the origin of the chain, and for that, we should use fast sync (instead of our current process and requesting blocks one-by-one).

Fast Sync is best described here: https://github.com/ethereum/go-ethereum/pull/1889

It's very similar to warp sync, except it begins from the beginning of the chain, pulls in all known block headers, etc, but instead of processing the blocks, it uses the receipts and decides which account states it needs to request.

This issue to build Fast Sync. The goal would be to allow a user to specify standard, fast or warp sync for the time-being and then we could reconcile the process.

PR #752, opened 2019-01-03
[WIP] Fast Sync

willmeister:
Implements Fast Sync as described by #637 

All that is left for this PR is:
* Add tests
* Add fast sync eligibility / configuration detection and set `BlockQueue.fast_sync_in_progress` accordingly
* Determine the chain's block height, flip `fast_sync_in_progress` to `false` and load necessary tries when we reach `block height - 1024`
* Validation for `Receipts` received against block header hash


InoMurko:
👍

PR #760, opened 2019-03-01
 #639 Peer Management - ranking peers

InoMurko:
This is another fraction of work for grading peers.
We're logging how long a connection was opened. The next step is to count incoming/outgoing messages, persist the ranking, and filter and/or upgrade the peers that we get from Kademlia discovery.


apps/ex_wire/lib/ex_wire/rank/connection_observer.ex
```diff
@@ -0,0 +1,128 @@
+defmodule ExWire.Rank.ConnectionObserver do
```
ayrat555: maybe we can add this module under `Kademlia` namespace?
(`ExWire.Kademlia.Rank.ConnectionObserver`)

InoMurko: Yeah makes sense!

ayrat555: @InoMurko Actually, it was a bad idea. Because this module keeps track of not kademlia nodes but peers

InoMurko: Yeah - I thought it’s ok since the peers come from the same protocol. But 🤘👍🏻

apps/ex_wire/lib/ex_wire/rank/peer_details.ex
```diff
@@ -0,0 +1,17 @@
...
+  @type t :: %__MODULE__{
+          peer: Peer.t() | nil,
+          connection_duration: Time.t() | nil,
+          sent_message_count: non_neg_integer(),
+          received_message_count: non_neg_integer()
+        }
+  defstruct peer: nil,
+            connection_duration: nil,
+            sent_message_count: 0,
+            received_message_count: 0
```
ayrat555: this two fields are currently not used, right? `sent_message_count` and `received_message_count`

InoMurko: No. That’s what I’m aiming next!

